PR #307, opened 2016-06-21
Use GitHub releases instead of branch tarball

nschonni:
Master branch download might include changes that aren't stable


BrentOzar:
Congratulations! You are pull request #1.

Great catch, too. That master branch is a hot mess right now. Once we get the Going Public milestone done, we'll do a release for that, and it'll be more stable.

Thanks!

Issue #278, opened 2016-06-11
sp_BlitzFirst - create schema specified in @OutputSchemaName if it doesn't exist


BrentOzar:
From Andy Schwabe:

sp_AskBrent @OutputSchemaName doesn't create a schema if it doesn't exist. Since AskBrent is more than fine with creating tables if they don't exist, could that be extended to the Schema that gets specified?

BlitzErik:
Creating schema is annoying because it wants to be the only statement ij

BlitzErik:
Overly sensitive phone screen ftl. Creating schema is annoying because it wants to be the only statement for 100 miles. Needs to be a dynamic chunk to avoid weird errors when you check to see if it exists first.

aschwabe10:
Erik,

Yeah it's a pain.... I'm still tweaking it to be more dynamic in my code for my job, but I figured if I ran into this, others might too. 

Is it a fair assumption to say that if someone is using @Output params they're probably putting it in a job? One thing I noticed was that I didn't get a create table failure, I got an error that a table didn't exist for inserts. Would it make sense as a stop gap measure to just validate at the top of the sproc if database and schema exist and then if they don't, raiserror('lorem ipsum', 20,-1) with log? I figure that people don't always check for job failures but they may check error logs... 

This kind of brings up another question. At this point in time if I specify a database that doesn't exist, I don't get any failures at all. I have no way besides actually checking my work to find out that this didn't actually do what I expected. Probably out of the question to create the db, but would the raiserror be a good fit here?

BrentOzar:
Oh yeah, we're totally not against doing it - Erik was just chiming in on why this will be a little tough. (He closed it accidentally when doing a comment on his phone.)

After looking at it for a few minutes, I think we'd just do a raiserror, and include the T-SQL to create the schema as part of the error if the schema doesn't exist. (If the database doesn't exist, I wouldn't wanna try to guess the T-SQL on that one since drives/files/etc., but creating the schema would be easy enough.)

aschwabe10:
Just build the db in C:\Windows\system32. It's where the good databases live!
I do like that the sproc still runs to completion if the output table doesn't exist. Just want it to give a little bit more verbose output. 
If I wanted to work on this myself, just branch, commit, test, pull request when done? Or are you guys not quite ready for that yet?
Part of me feels a bit bad just submitting small feature requests that I could really easily work on myself...

BlitzErik:
[...]

PR #323, opened 2016-06-28
Adding in Output Parameter Validation for sp_BlitzFirst

AndySchwabe:
This is basic validation of only the output parameters. This can use a lot of expansion and work. It also needs to be incorporated into the procedure to be used as flow control to stop other errors, especially with empty parameters. If parameters are empty strings instead of just NULL, a lot of errors are thrown during the attempt to INSERT INTO the output tables. 

I'm submitting this as is to only this procedure to ensure I'm not going down the wrong path with my current checks. It has been tested on SQL 2008 Express CS_AS and SQL 2016 Express CS_AS.

Please rip this to shreds if need be. 

This is for issue #278


AndySchwabe:
Closing to move code to dev branch

PR #349, opened 2016-07-01
Blitzcache erik issue339

BlitzErik:
Working fix for stored proc and statement level costs and row info


BlitzErik:
Nope nope nope nope nope.

PR #357, opened 2016-07-11
Blitzfirst erik issue 356

BlitzErik:
Fixes:
  Filter out perfmon counters of 0
  Include login_name in what's running
  All reserved words are now upper case


nschonni:
Shared this with Brent on Slake, but FYI https://help.github.com/articles/closing-issues-via-commit-messages/

PR #376, opened 2016-07-14
Update what is running queries to reflect new columns

BlitzErik:
This ended up being a bit of an overhaul, but its been tested across all
relevant AWS instances

1. Checks product versions the same way sp_Blitz does
2. Remove support for 2005 -- it won't compile with the new OFFSET data
type -- nothing will run there anyway
3. For 2008 and 2008R2, runs a stripped down version of a 'more
interesting' what is running query that for some reason wasn't used the
first time
4. for versions of 2012 and 2014 that don't have the new columns for:
mem grants etc, a case expression takes them out of the dynamic sql
5. for versions that do support it, they're included

Testing was good on this. Works across all versions, with confidence,
since our 2014 instance in AWS isn't patched to SP2, but my local is,
and the minor version of our 2016 instance is pre-RTM, and that worked.
2012 Worked including the columns, but there's no pre-SP2 handy to test
the without, but I think I got the comparison right on this one ;)

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -221,335 +221,249 @@ BEGIN
 	@ProductVersionMinor = PARSENAME(CONVERT(VARCHAR(32), @ProductVersion), 2)
 
 
-    /* What's running right now? This is the first result set. */
-	IF @ProductVersionMajor > 9 and @ProductVersionMajor < 11
+    /* What's running right now? This is the first and last result set. */
+    IF @SinceStartup = 0 AND @Seconds > 0 AND @ExpertMode = 1 
     BEGIN
-    SET @StringToExecute = N'
-							SELECT  GETDATE() AS [run_date] ,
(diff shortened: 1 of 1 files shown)


BrentOzar:
I checked in some more code on this - putting everything into a single IF statement, tweaking the wording. Can you sync to get my changes in your branch, then test this on 2008-2016 again? If you can, test with both memory grant queries and queries that don't need RAM. I've been using these two:

select * from StackOverflow.dbo.Posts
select * from StackOverflow.dbo.Posts ORDER BY Title DESC

BlitzErik:
Can do.

BlitzErik:
Sure as heck works in 2008! Check out that query cost.

image

BlitzErik:
This is re-tested, and looks good. My one concern is that the query cost column is NULL for the first query, but populated for the second one. We could do something with ISNULL, but it's misleading to replace with a 0 here. Any thoughts?

BrentOzar:
Oh great point. I like it being null since we don't know it yet - we can leave it that way.

PR #386, opened 2016-07-16
Blitzcache erik multiple

BlitzErik:
THIS RULES OK

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -1036,7 +1036,7 @@ IF (@SortOrder IN ('memory grant', 'avg memory grant'))
 AND ((@v < 11)
 OR (@v = 11 AND @build < 6020) 
 OR (@v = 12 AND @build < 5000) 
-OR (@v = 13 AND @build < 1708))
+OR (@v = 13 AND @build < 1601))
 BEGIN
    RAISERROR('Your version of SQL does not support sorting by memory grant or average memory grant. Please use another sort order.', 16, 1);
    RETURN;
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -23,38 +23,38 @@ CREATE TABLE ##bou_BlitzCacheResults (
 
 CREATE TABLE ##bou_BlitzCacheProcs (
     SPID INT ,
-    QueryType nvarchar(256),
+    QueryType NVARCHAR(256),
     DatabaseName sysname,
-    AverageCPU decimal(38,4),
-    AverageCPUPerMinute decimal(38,4),
-    TotalCPU decimal(38,4),
-    PercentCPUByType money,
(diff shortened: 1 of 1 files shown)


BlitzErik:
Pictures to prove it:

image

image

BrentOzar:
DAAAAAAANG SON

Not gonna lie, the memory grant improvements get me more excited than two for one night at Cliff Lede. This is fantastic.

BlitzErik:
Yeah! I need to add something to raise an error if the version is wrong. That's easy though. I'll add it tomorrow morning.

BlitzErik:
How to repro and test this:

If you're using StackOverflow, you can run these queries:

```
SELECT [u].[DisplayName], SUM([u].[DownVotes]), SUM([u].[UpVotes])  
FROM [dbo].[Users] AS [u]  
WHERE [u].[UpVotes] > 1000  
AND [u].[DownVotes] > 1000  
AND [u].[Views] >= 100  
AND [u].[LastAccessDate] >= '2014-01-01'  
GROUP BY [u].[DisplayName]

SELECT [u].[DisplayName], SUM([u].[DownVotes]), SUM([u].[UpVotes])  
FROM [dbo].[Users] AS [u]  
WHERE [u].[Age] < 50  
AND [u].[DownVotes] > 1000  
AND [u].[Views] >= 100  
GROUP BY [u].[DisplayName]

SELECT [u].[DisplayName], SUM([u].[DownVotes]), SUM([u].[UpVotes])  
FROM [dbo].[Users] AS [u]  
WHERE [u].[UpVotes] > 1000  
AND [u].[DownVotes] > 1000  
AND [u].[Views] >= 100  
AND [u].[CreationDate] >= '2014-01-01'  
GROUP BY [u].[DisplayName]
```

They should all result in some large but unused memory grants, like these:
image

BrentOzar:
Can you test this in all the machines in the lab? The 2016 one is showing the memory grant columns (the pictured above ones) as all nulls.

For example, ran this in ConsumeMassQuantities1:

CREATE TABLE dbo.Stuffing (ID INT IDENTITY(1,1) PRIMARY KEY CLUSTERED, Stuff CHAR(6000));
INSERT INTO dbo.Stuffing (Stuff) SELECT 'Stuff' FROM sys.all_columns;
GO
SELECT * FROM dbo.Stuffing ORDER BY Stuff DESC;
GO

While that SELECT was running, I checked this:

SELECT * FROM sys.dm_exec_query_memory_grants

And it showed grants. But when I run this:

sp_BlitzCache  @ExpertMode = 1, @SortOrder  = 'memory grant'

The Serial Desired and Serial Required columns are populated, but the rest aren't:

BlitzErik:
This is fixed in the last commit! Working in AWS 2016 now. Will still fail on 2014 because versions.

PR #387, opened 2016-07-17
Blitzfirst erik multiple

BlitzErik:
Add waiting tasks to wait stats sampling.


BlitzErik:
To test this, here's an example of what you can do:

You'll need three query editor windows open in SSMS. Run this, not including the ROLLBACK in two of them:

```
USE StackOverflow
SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
BEGIN TRAN
UPDATE dbo.Posts
SET ClosedDate = GETDATE()
WHERE ClosedDate IS NULL
ROLLBACK

```

In the third window, run this:

```
USE StackOverflow
SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
SELECT *
FROM [dbo].[Posts] AS [p]
WHERE [p].[AnswerCount] > 0
AND [p].[LastEditorDisplayName] IS NOT NULL
AND [p].[ClosedDate] IS NOT NULL
AND [p].[PostTypeId] = 1
GO
```

EXEC dbo.sp_BlitzFirst @ExpertMode = 1, @Seconds = 10

You should end up with results like this, which accurately tells you how many seconds of the sample window you spent waiting on long running, incomplete transactions
image

BlitzErik:
image

Issue #293, opened 2016-06-13
Use @OutputServerName to write results to a linked server


BrentOzar:
If this parameter is populated, check to see if a linked server exists by this name. If not, print an error.

If the linked server exists, use @OutputDatabaseName, @OutputSchemaName, and @OutputTableName to write the results.

We don't have to do big time error checking here - no need to get fancy and check for the database and the schema for now, can always add that in a later version.

Use case: someone runs the stored procedures regularly as an Agent job, and writes the results to a central server for reporting purposes.

PerfektNerd:
The best way to do this IMO would be to just rewrite the existing @OutputTableName section with some new logic. That way we can avoid unnecessary code bloat. 

Sounds like fun! I'll give it a shot.

PerfektNerd:
Just an update on this guy. I started working on this feature for sp_Blitz.

I know you didn't want fancy error checking, but since I needed to check whether or not there was a linked server on the other side, I added a bit more than just the linked server check. I check if the linked server is valid, then check if the database exists on the linked server. I took the logic for the local location dump out as well, and just put it all in an error checking block that spits out whether or not it's a valid location in bit form.

```
/* Checks if @OutputServerName is populated with a valid linked server, and that the database name specified is valid */

				DECLARE @ValidOutputServer BIT
				DECLARE @ValidOutputLocation BIT
				DECLARE @LinkedServerDBCheck NVARCHAR(2000)
				DECLARE @Count INT
				IF @OutputServerName IS NOT NULL
					BEGIN
						IF EXISTS (SELECT server_id FROM sys.servers WHERE QUOTENAME([name]) = @OutputServerName)
							BEGIN
								SET @LinkedServerDBCheck = 'SELECT * FROM '+@OutputServerName+'.master.sys.databases WHERE QUOTENAME([name]) = '''+@OutputDatabaseName+''''
								EXEC sys.sp_executesql @LinkedServerDBCheck
								SELECT @Count = @@ROWCOUNT
								IF (@Count > 0)
									BEGIN
										SET @ValidOutputServer = 1
										SET @ValidOutputLocation = 1
									END
								ELSE
									RAISERROR('The specified database was not found on the output server', 16, 0)
							END
						ELSE
							BEGIN
								RAISERROR('The specified output server was not found', 16, 0)
							END
					END
				ELSE
					BEGIN
						IF @OutputDatabaseName IS NOT NULL
[...]

PR #391, opened 2016-07-20
#293 Support for @OutputServerName in sp_Blitz

PerfektNerd:
Adding support for outputting results of sp_Blitz to table on Linked server along with some error checking for @OutputDatabaseName.

I have tested on all supported instances of SQL Server except 2014 and 2016. Help with testing would be appreciated.


BrentOzar:
Works beautifully! Nice job.

PR #405, opened 2016-07-29
Make consistent output of Script Version and Date - sp_BlitzIndex

jorsol:
When the Version and Date parameters are updated, there where 2 places
where it has to be mannually updated as well, now when the Version and Date parameters are updated, it use uniform output of Script Version and Date.

It was confusing that the output was 4.0 and the real version is 4.1.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -130,7 +130,7 @@ SOFTWARE.
 '
 
 
-
+DECLARE @ScriptVersionName NVARCHAR(50);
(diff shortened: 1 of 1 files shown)


BrentOzar:
Hi! Brent here. Thanks for the pull request, and great idea. Totally agree with how you did it, nice work.

I'm going to go ahead and merge this one, but for the future, check out the contributing guidelines. We ask that folks create an issue first and discuss it just to make sure changes are done in a way that makes sense for the overall project. (Changes like this one are a total no-brainer though - good work!)

PR #419, opened 2016-08-10
#293 Support for @OutputServerName in sp_BlitzCache

### LINKED ISSUE #293: Use @OutputServerName to write results to a linked server

PerfektNerd:
Fixes #293 for sp_BlitzCache.

- Support for @OutputServerName
- Updated error checking for @OutputDatabaseName variables for both linked and local servers

How to test this code:
- Run procedure output with @expertmode = 1 (for comparison)
- Run procedure with valid @OutputDatabaseName/SchemaName/TableName specified (should succeed in creating local table)
- Run procedure with @OutputServerName specified as valid linked server along with other Output variables (should succeed in creating remote table on linked server)
- Rerun both tests above without deleting testing tables to ensure that procedure doesn't try and create new table (it should dump results to the same table; tests IF EXISTS logic in permanent table creation)
- Run procedure with invalid @OutputServerName (server that is not linked and random name; both should fail)
- Run procedure with invalid local DB name (should fail)
- Run procedure with invalid linked DB name (Specify valid @OutputServerName; should fail)

Use the output from the @expertmode = 1 run to compare the output on both the local and remote tables...they should be the same.

Has been tested on:
- SQL Server 2008
- SQL Server 2008 R2
- SQL Server 2012
- **(UPDATE)** SQL Server 2016

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -22,127 +22,117 @@ CREATE TABLE ##bou_BlitzCacheResults (
 );
 
 CREATE TABLE ##bou_BlitzCacheProcs (
-        SPID INT ,
-        QueryType NVARCHAR(256),
-        DatabaseName sysname,
-        AverageCPU DECIMAL(38,4),
-        AverageCPUPerMinute DECIMAL(38,4),
-        TotalCPU DECIMAL(38,4),
-        PercentCPUByType MONEY,
-        PercentCPU MONEY,
-        AverageDuration DECIMAL(38,4),
-        TotalDuration DECIMAL(38,4),
-        PercentDuration MONEY,
-        PercentDurationByType MONEY,
-        AverageReads BIGINT,
-        TotalReads BIGINT,
-        PercentReads MONEY,
-        PercentReadsByType MONEY,
-        ExecutionCount BIGINT,
-        PercentExecutions MONEY,
-        PercentExecutionsByType MONEY,
-        ExecutionsPerMinute MONEY,
-        TotalWrites BIGINT,
-        AverageWrites MONEY,
-        PercentWrites MONEY,
-        PercentWritesByType MONEY,
-        WritesPerMinute MONEY,
-        PlanCreationTime DATETIME,
-        LastExecutionTime DATETIME,
-        PlanHandle VARBINARY(64),
-		[Remove Plan Handle From Cache] AS 
-			CASE WHEN [PlanHandle] IS NOT NULL 
-			THEN 'DBCC FREEPROCCACHE (' + CONVERT(VARCHAR(128), [PlanHandle], 1) + ');'
-			ELSE 'N/A' END,
-		SqlHandle VARBINARY(64),
-			[Remove SQL Handle From Cache] AS 
-			CASE WHEN [SqlHandle] IS NOT NULL 
-			THEN 'DBCC FREEPROCCACHE (' + CONVERT(VARCHAR(128), [SqlHandle], 1) + ');'
-			ELSE 'N/A' END,
-		[SQL Handle More Info] AS 
-			CASE WHEN [SqlHandle] IS NOT NULL 
-			THEN 'EXEC sp_BlitzCache @OnlySqlHandles = ''' + CONVERT(VARCHAR(128), [SqlHandle], 1) + '''; '
-			ELSE 'N/A' END,
-		QueryHash BINARY(8),
-		[Query Hash More Info] AS 
-			CASE WHEN [QueryHash] IS NOT NULL 
-			THEN 'EXEC sp_BlitzCache @OnlyQueryHashes = ''' + CONVERT(VARCHAR(32), [QueryHash], 1) + '''; '
-			ELSE 'N/A' END,
-        QueryPlanHash BINARY(8),
-        StatementStartOffset INT,
-        StatementEndOffset INT,
-        MinReturnedRows BIGINT,
-        MaxReturnedRows BIGINT,
-        AverageReturnedRows MONEY,
-        TotalReturnedRows BIGINT,
-        LastReturnedRows BIGINT,
-		/*The Memory Grant columns are only supported 
-		  in certain versions, giggle giggle.
-		*/
-		MinGrantKB BIGINT,
-		MaxGrantKB BIGINT,
-		MinUsedGrantKB BIGINT, 
-		MaxUsedGrantKB BIGINT,
-		PercentMemoryGrantUsed MONEY,
-		AvgMaxMemoryGrant MONEY,
-        QueryText NVARCHAR(MAX),
-        QueryPlan XML,
-        /* these next four columns are the total for the type of query.
-            don't actually use them for anything apart from math by type.
-            */
-        TotalWorkerTimeForType BIGINT,
-        TotalElapsedTimeForType BIGINT,
-        TotalReadsForType BIGINT,
-        TotalExecutionCountForType BIGINT,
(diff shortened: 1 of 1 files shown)


PerfektNerd:
Tested this on SQL Server 2016 (Got my instance connected to my domain) :)

sp_BlitzCache.sql
```diff
@@ -2096,9 +2098,59 @@ SET    Warnings = SUBSTRING(
...
+				DECLARE @ValidOutputServer BIT
+				DECLARE @ValidOutputLocation BIT
+				DECLARE @LinkedServerDBCheck NVARCHAR(2000)
+				DECLARE @ValidLinkedServerDB INT
+				DECLARE @tmpdbchk table (cnt int)
+				IF @OutputServerName IS NOT NULL
+					BEGIN
+						IF EXISTS (SELECT server_id FROM sys.servers WHERE QUOTENAME([name]) = @OutputServerName)
+							BEGIN
+								SET @LinkedServerDBCheck = 'SELECT 1 WHERE EXISTS (SELECT * FROM '+@OutputServerName+'.master.sys.databases WHERE QUOTENAME([name]) = '''+@OutputDatabaseName+''')'
```
BrentOzar: Does this work if someone only has permissions to a single database, but not to master.sys.databases? I think generally in linked server situations, the cross-server account users don't have a lot of permissions.

sp_BlitzCache.sql
```diff
@@ -2163,72 +2215,122 @@ BEGIN
...
+          StatementStartOffset INT,
+          StatementEndOffset INT,
+          MinReturnedRows BIGINT,
+          MAXReturnedRows BIGINT,
+          AverageReturnedRows MONEY,
+          TotalReturnedRows BIGINT,
+          QueryText NVARCHAR(MAX),
+          QueryPlan XML,
+          NumberOfPlans INT,
+          NumberOfDistinctPlans INT,
```
BrentOzar: Uh oh - looks like maybe this was started from an older version, and it didn't have all of the current fields. Can you add the ones from lines 2182-2187 in the last version, and test that?

sp_BlitzCache.sql
```diff
@@ -2163,72 +2215,122 @@ BEGIN
...
-          + N' AverageWrites, TotalWrites, PercentWrites, PercentWritesByType, ExecutionCount, PercentExecutions, PercentExecutionsByType, '
-          + N' ExecutionsPerMinute, PlanCreationTime, LastExecutionTime, PlanHandle, SqlHandle, QueryHash, StatementStartOffset, StatementEndOffset, MinReturnedRows, MaxReturnedRows, AverageReturnedRows, TotalReturnedRows, QueryText, QueryPlan, NumberOfPlans, NumberOfDistinctPlans, Warnings, '
-          + N' SerialRequiredMemory, SerialDesiredMemory, MinGrantKB, MaxGrantKB, MinUsedGrantKB, MaxUsedGrantKB, PercentMemoryGrantUsed, AvgMaxMemoryGrant '
-          + N' FROM ##bou_BlitzCacheProcs '
+	IF @ValidOutputServer = 1
+		BEGIN
+			SET @insert_sql = REPLACE(@insert_sql,''''+@OutputSchemaName+'''',''''''+@OutputSchemaName+'''''')
+			SET @insert_sql = REPLACE(@insert_sql,''''+@OutputTableName+'''',''''''+@OutputTableName+'''''')
+			SET @insert_sql = REPLACE(@insert_sql,'''DBCC FREEPROCCACHE ('' + CONVERT(VARCHAR(128), [PlanHandle], 1) + '');''','''''DBCC FREEPROCCACHE ('''' + QUOTENAME(CONVERT(VARCHAR(128), [PlanHandle], 1), CHAR(39)) + '''');''''')
```
BrentOzar: Dumb question - what are we search/replacing this part for? (I'm guessing you found something neat during testing, but if there's an easier way to fix it than doing search/replace, I'm totally down with that - replacing in long strings can be pretty slow.)

sp_BlitzCache.sql
```diff
@@ -2163,72 +2215,122 @@ BEGIN
...
-                                                    WHEN 'avg cpu' THEN N' AverageCPU'
-                                                    WHEN 'avg reads' THEN N' AverageReads'
-                                                    WHEN 'avg writes' THEN N' AverageWrites'
-                                                    WHEN 'avg duration' THEN N' AverageDuration'
-                                                    WHEN 'avg executions' THEN N' ExecutionsPerMinute'
-													WHEN 'avg memory grant' THEN N' AvgMaxMemoryGrant'
-                                                    END + N' DESC '
-
-    SET @insert_sql += N' OPTION (RECOMPILE) ; '    
+			SELECT @insert_sql += N' ORDER BY ' + CASE @SortOrder WHEN 'cpu' THEN ' TotalCPU '
```
BrentOzar: Should we still be doing an order by in the insert, come to think of it?

sp_BlitzCache.sql
PerfektNerd: Sure thing!

sp_BlitzCache.sql
PerfektNerd: Since I have to use sp_executesql on the other box:

`EXEC('EXEC('''+@insert_sql+''') AT ' + @OutputServerName);`

...I end up needing extra single-quotes in the dynamic SQL string to make it function.

For the Plan Handles and the Query Hashes, it's the same problem. Without the correct number of single-quotes I end up inserting either` DBCC FREEPROCCACHE ('' + CONVERT(VARCHAR(128), [SqlHandle], 1) + ''); `or the Plan Handle/Query Hash with a pair of single quotes around it into the output table.

Actually come to think of it, would this even work? Would it end up executing the DBCC on the other server? Maybe I need to dump the results to a temp table which I copy over instead...

PerfektNerd: I should also note that I did it this way for code reuse; and that performance was never an issue here since the insert_sql string is not insanely long or anything. If we were updating a ton of strings it would probably be an issue though.

sp_BlitzCache.sql
PerfektNerd: As long as they are in the Public role it should be fine. If the DBA/Vendor has removed all of the precreated roles for their own, the linked server account needs to have VIEW ANY DATABASE permission.

 From the docs on sys.databases:

> By default, the public role has the VIEW ANY DATABASE permission, allowing all logins to see database information. To block a login from the ability to detect a database, REVOKE the VIEW ANY DATABASE permission from public, or DENY the VIEW ANY DATABASE permission for individual logins.

sp_BlitzCache.sql
BrentOzar: OK, just update the issue when that code has been added. (Just went in and checked and it still wasn't there, so noting in here in case you were missing that.)

sp_BlitzCache.sql
BrentOzar: Nah, you don't want to do that - let's not try to execute SQL on other servers. (I can see that being a security nightmare.)

BrentOzar:
This still hasn't been fixed after a few months, so I'm going to go ahead and close it. Thanks though.

Issue #429, opened 2016-08-25
BlitzIndex allow skip of partitioning checks 


BlitzErik:
Add ability to skip partitioning checks. Despite many efforts, it's really slow sometimes.

BlitzErik:
#430

Issue #347, opened 2016-07-01
sp_BlitzIndex - duplicate index warning doesn't include table name


BrentOzar:
It's in the More Info column in the T-SQL, but needs to also be in the Details column.

BlitzErik:
#430

Issue #317, opened 2016-06-25
sp_BlitzIndex - when >50 databases, return error in result rather than printing it


BrentOzar:
sp_BlitzIndex v4 correctly bails out when we have >50 databases, but that it's printing an error message rather than returning a result set. We should return a result set so folks parsing the output can see it:

Starting run. sp_BlitzIndex(R) v3.0 - March 20, 2016
Create temp tables.
You're trying to run sp_BlitzIndex on a server with 102 databases. 
Running sp_BlitzIndex on a server with 50+ databases may cause temporary insanity for the server and/or user.
If you're sure you want to do this, run again with the parameter @BringThePain = 1.

BlitzErik:
#430

Issue #314, opened 2016-06-22
Possible bug in sp_blitzindex?


thornersetrans:
When running sp_blitzindex with mode=0, I get all heap tables with forwarded fetches. With mode=4, I get a subset of those.  Looking at SP it appears that when using mode=4 heap size has to be greater than ThresholdMB to be reported, yet all are reported if mode=0.  Is this as designed?

BrentOzar:
It isn't designed that way, no.

BlitzErik:
#430

PR #430, opened 2016-08-26
Blitzindex erik multiple

BlitzErik:
#82 
#146 
#253 
#262 
#314 
#317 
#347 
#429


BrentOzar:
Try this in the lab - it doesn't return any results on any of the servers/databases I tried.

PR #431, opened 2016-08-26
Issue 407

### LINKED ISSUE #407: sp_BlitzFirst - Move the "What's running right now?" to its own stored procedure
I'd like to move the "What's running right now?" section out from sp_BlitzFirst into its own stored procedure.

2 reasons for this recommendation come to mind.
1. It would be useful to run on its own.  This can be worked around by running `@ExpertMode=1, @Seconds=1` but it generates a lot of extra data and takes a little bit longer to execute
2. Code reuse.  It's already called twice (4 if you count versions independently), so if a bug is found or a new feature is added it has to be changed in multiple locations.

rwhoward:
Fixes #407

 - Split the what's running now into it's own stored procedure
 - Modified that script into a version runnable against any recent version
 - Altered sp_BlitzFirst to call this code

How to test this code:
 - Make sure sp_BlitzFirst installs correctly
 - Make sure sp_BlitzFirst runs with no errors if sp_BlitzWho is not installed (there should be a note printed, but no errors)
 - Make sure sp_BlitzWho installs correctly
- Make sure sp_BlitzWho runs with no errors and returns an expected dataset

 - Case-sensitive SQL Server instance (2008)
 - SQL Server 2008
 - SQL Server 2014
 - SQL Server 2016


BrentOzar:
Deleting this since it had merge problems. Absolutely no problems with the code - it's just that when you do several commits like this, with changes on each, it can make it tough for Github to merge your code. I just started over with a single commit using your code, made a couple of small tweaks, did one commit, and it worked fine. Good job!

PR #433, opened 2016-08-26
Fix for issue 432

BlitzErik:
Information wasn't aggregating correctly with getalldatabases = 1


BlitzErik:
To repro:
Run with GetAllDatabases = 1 and Mode = 1

It will associate all aggregated data with the last database it ran against.

PR #457, opened 2016-09-26
Replace jsfiddle.net with sqlfiddle.com

rpellicaan:
Fixes a link to a non relevant site .


BrentOzar:
Thanks for the pull request! Good point on the fix.

PR #512, opened 2016-10-14
BlitzIndex now checks for some potential statistics issues

BlitzErik:
Depending on version, there are two possible syntax blocks. They both
populate a temp table called #Statistics, and so far I've added two
checks. One for outdated stats, and one for stats with a low sample
rate.

Tested in the lab on
2008
2008R2
2012
2014

Tested locally
2016

Ready for more testing!


BrentOzar:
Don't kick my butt, but can you use datetime instead of date? sp_BlitzIndex worked on 2005 before this change, and I'd hate to break it just for that one tweak.

Issue #549, opened 2016-10-19
BlitzIndex - Change column references for stats insert


BlitzErik:
**Do you want to request a *feature* or report a *bug*?**
OBJECT_NAME and OBJECT_SCHEMA_NAME are wonky. 

**What is the current behavior?**
NULLs

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via http://sqlfiddle.com

**What is the expected behavior?**
Table and Schema names

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
All

BlitzErik:
#550

Issue #548, opened 2016-10-19
BlitzIndex - Need to outer apply stats function to catch some problems


BlitzErik:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Not all statistics objects are captured when using CROSS APPLY

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via http://sqlfiddle.com

**What is the expected behavior?**
All statistics objects should be examined

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
2008R2, 2012, 2014, 2016

BlitzErik:
#550

Issue #545, opened 2016-10-19
sp_BlitzIndex - check for scalar UDFs in computed fields


BrentOzar:
**Do you want to request a _feature_ or report a _bug_?**
Feature

**What is the current behavior?**
If you have a computed field with a scalar UDF, all access on the table is single-threaded. (Not sure if this is true for selects when the field is persisted, but definitely is for insert/update/deletes.)

**What is the expected behavior?**
Add a new warning about scalar UDFs in computed fields.

BlitzErik:
Yeah, it's true for selects. I got this one.

BlitzErik:
#550

BlitzErik:
Merged, closing.

Issue #471, opened 2016-10-04
sp_BlitzIndex -- Should aggressive index warnings occur when the average duration is 0 seconds?


BlitzErik:
Should there be two warnings for this? Maybe for one where the locks are aggressive but the lock waits are short, vs where the lock waits are >= 5 seconds?

BlitzErik:
#550

BlitzErik:
Merged, closing.

PR #550, opened 2016-10-20
Sp blitzindex 548

BlitzErik:
#548 #549 #254 #471 #545


BlitzErik:
This was tested in the lab on 2005-2012
Tested locally on 2014 and 2016 (patching shouldn't matter here, we only need to check for 2008+ for the sys.computed columns view)

PR #589, opened 2016-11-15
Blitzcache 587

BlitzErik:
Fixes logic bug for deleting data from warnings

Adds more verbose output to let you know what it's up to

Adds proper deletion from ##bou_BlitzCacheProcs back so performance doesn't become expoentially worse!

Adds graceful failure for 2000/2005

Adds top level warning for queries with no execution plan

Adds version number to output

Adds check for ColumnStore indexes operating in row mode (Warnings, too)

Adds a couple more plan cache checks for single use plans and lots of plan stubs.


BlitzErik:
Tested in the lab:
2005
2008
2008R2
2012

BlitzErik:
Added checks for lots of single use plans and lots of plan stubs. They are working, and here are what they look like with the WHERE clause removed

image

I took the suggested levels from the issues. If anyone has better thresholds, I'm totally open to changing them.

BrentOzar:
Awesome! Looks good, sir. This will be in the December release. (Aiming for late this week.)

PR #612, opened 2016-11-30
sp_Blitz Check if Min Server Memory = Max Server Memory

### LINKED ISSUE #603: sp_Blitz - Warn when min server memory = max server memory
Feature; idea from https://github.com/Microsoft/tigertoolbox/blob/master/BPCheck/Check_BP_Servers.sql

We don't warn about this

Seems like a good thing to warn about. What with the non-dynamic memory and such.

The Tiger Toolbox is licensed under MIT's license. The code involved is here:

`WHEN @minservermem = @maxservermem THEN '[WARNING: Min Server Mem setting is equal to Max Server Mem. This will not allow dynamic memory. Please revise memory settings]'`

RichBenner:
Fixes #603 
 - New alert to check if max server memory = min server memory

How to test this code:
 - Change server memory to the same as max memory and run the script.
 - Test with different settings to confirm it doesn't show

 - SQL Server 2014
 - SQL Server 2016

Added two new parameters and then a block of code that will warn if min memory is the same as max.


tonyfountain:
FYI this is an SAP standard. Why would this need to be a warning?

RichBenner:
Original enhancement raised by Erik; https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/603

PR #619, opened 2016-12-05
typo fix for case-sensistive servers

mikejwhitty:
New to gitflow, sorry for any mistakes...

Fixes # .

 - Allow sp_Blitz to be created / modified on case-sensitive server

 - Case-sensitive SQL Server instance - YES
 - SQL Server 2008
 - SQL Server 2008 R2
 - SQL Server 2012 - YES
 - SQL Server 2014 - YES
 - SQL Server 2016
 - Amazon RDS
 - Azure SQL DB

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -1526,7 +1526,7 @@ AS
 										'Performance',
 										'Non-Dynamic Memory',
 										'http://BrentOzar.com/go/memory',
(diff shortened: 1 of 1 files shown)


BrentOzar:
Son of a gun! We're usually good about testing on case sensitive stuff and I must have missed that. Great catch, and thanks!

PR #620, opened 2016-12-05
sp_BlitzIndex - Flag if using columnstore indexes & trace 834

### LINKED ISSUE #615: sp_BlitzIndex - add check for trace flag 834 when using columnstore indexes
Turns out that large pages and columnstore indexes don't work well together:
https://support.microsoft.com/en-us/kb/3210239

For the more-info URL, can link directly to that page.

RichBenner:
Fixes #615 

 - Added a check to see if columnstore indexes are in use alongside trace flag 834

How to test this code:
 - Find a server with columnstore indexes and flag 834 switched on (or make one) and run to ensure it flags.
 - Test on servers with either and both switched off to ensure it doesn't show.

 - SQL Server 2016 Developer (with and without trace flags)
(couldn't test on earlier versions as the servers are currently in use and it requires a reboot for flag 834)

Inserts into mode = 0 if columnstore indexes exist and the trace flag 834 is switched on. Currently set to a 150 priority error under 'Abnormal Psychology'. 

#TraceStatus table and data insert logic taken from sp_Blitz for consistency.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -3095,7 +3095,7 @@ BEGIN;
             AND NOT (@GetAllDatabases = 1 OR @Mode = 0)
 
 			RAISERROR(N'check_id 72: Columnstore indexes with Trace Flag 834', 0,1) WITH NOWAIT;
-                IF EXISTS (SELECT * FROM #IndexSanity WHERE Index_Type IN (5,6))
+                IF EXISTS (SELECT * FROM #IndexSanity WHERE index_type IN (5,6))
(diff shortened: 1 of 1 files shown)


RichBenner:
As an aside, would this be useful in sp_Blitz too? Would be simple enough to add it.

BlitzErik:
There's a section of Blitz that adds some extra explanation to active trace flags (if they're weird). It might be cool to add it there. Look for this section:

`/*Identify globally enabled trace flags*/`

RichBenner:
Great, I'll get that done tomorrow

BrentOzar:
Good work, sir! Made one little tweak for case sensitive compatibility but that was it. Thanks!

PR #623, opened 2016-12-05
sp_BlitzIndex -- stop counting hypothetical indexes in NC index info

### LINKED ISSUE #622: sp_BlitzIndex -- stop counting hypothetical indexes in NC index info
Bug

We count hypothetical indexes created by DTA towards the total NC indexes on a table in some places

image

Only count real live indexes

All

BlitzErik:
Closes #622


BrentOzar:
Amazing that we'd never run into this before! Nice catch, sir.

PR #624, opened 2016-12-06
sp_Blitz - Trace Flag 834 with Columnstore Indexes

RichBenner:
Referenced in #620  relates to #615 .

 - Add check in trace flags to see if trace 834 is being used alongside columnstore indexes

How to test this code:
 - Test on a server that contains columnstore indexes and also has trace flag 834 switched on. The alert will appear at Priority 200 - Informational - Trace Flag On
 - Check on a server with no columnstore indexes to ensure issue doesn't appear
 - Check on a server with trace 834 switched off to ensure it doesn't appear
 - If the trace flag is switched on and no columnstore indexes are in use then it should show a generic text stating that this flag is on ('Trace flag 834 is enabled globally.') rather than the specific error message ('Trace flag  834 is enabled globally. Using this Trace Flag with Columnstore Indexes is not a great idea.').' 

 - SQL Server 2016 Developer
(can't check on another instance locally due to the requirement for a server restart for the trace flag)

Added flag to highlight if trace flag 834 is being used along with columnstore indexes. Originally raised for sp_BlitzIndex on issue #615 and now added to sp_Blitz as per pull request #620.

Have added the link to the microsoft page from issue #615 but I'm aware this is at odds with the other URL's in sp_BlitzIndex. Didn't want to add this to the details column as it would get pretty big. Might be an idea if the microsoft URL could be added to BrentOzar.com/go/traceflags for consistency.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -140,7 +140,8 @@ AS
 			,@Processors int
 			,@NUMANodes int
 			,@MinServerMemory bigint
-			,@MaxServerMemory bigint;
+			,@MaxServerMemory bigint
+			,@ColumnStoreIndexUse bit;
 
 
 		SET @crlf = NCHAR(13) + NCHAR(10);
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks sir! Slight gotcha with this - the below code only checks the current database to see if it has columnstore indexes:

`/* Check if columnstore indexes are in use */
		IF EXISTS (SELECT * FROM sys.indexes WHERE type IN (5,6))
		BEGIN
		SET @ColumnStoreIndexUse = 1
		END`

Instead, we'll need to add a loop in the @CheckUserDatabaseObjects = 1 section looping through all databases looking for those indexes. Accepting the pull request, then adding that part of the code.

PR #630, opened 2016-12-08
sp_BlitzFirst - Should accept code to run for the duration of

### LINKED ISSUE #629: sp_BlitzFirst - Should accept code to run for the duration of
Feature

Right now, if you want to measure a piece of code, you have to run it in one SSMS tab, and sp_BlitzFirst in another. 

Guessing the duration can kind of suck. You should be able to feed a stored proc or ad hoc code to sp_BlitzFirst so that it executes until the code completes.

All

BlitzErik:
Closes #629

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -144,6 +144,18 @@ IF (@UserStoredProc IS NOT NULL AND @UserAdHoc IS NOT NULL)
    RETURN;
    END  
 
+IF (@UserStoredProc IS NOT NULL AND LEN(@UserStoredProc) <= 6)
+   BEGIN
+   RAISERROR('Are you sure you supplied a stored procedure?',0,1) WITH NOWAIT;
+   RETURN;
+   END  
+
+IF (@UserAdHoc IS NOT NULL AND LEN(@UserAdHoc) <= 7)
+   BEGIN
+   RAISERROR('Are you sure you supplied working code?',0,1) WITH NOWAIT;
+   RETURN;
+   END  
+
 IF (@UserStoredProc IS NOT NULL OR @UserAdHoc IS NOT NULL)   
    BEGIN
    SELECT @SinceStartup = 0, @ExpertMode = 1, @Seconds = 1, @StartSampleTime = SYSDATETIMEOFFSET()
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -139,29 +139,47 @@ ELSE IF @Seconds = 0 AND CAST(SERVERPROPERTY('edition') AS VARCHAR(100)) <> 'SQL
 ELSE
     SELECT @StartSampleTime = SYSDATETIMEOFFSET(), @FinishSampleTime = DATEADD(ss, @Seconds, SYSDATETIMEOFFSET());
 
+/*Begin code duration value checks*/
+
+/*Check if configurations are missing.*/
+IF (@UserStoredProc IS NOT NULL OR @UserAdHoc IS NOT NULL)
+		AND (
+				(SELECT CONVERT(BIT, c.value_in_use) FROM sys.configurations AS c WHERE c.name = 'show advanced options') = 0
+			OR 
+				(SELECT CONVERT(BIT, c.value_in_use) FROM sys.configurations AS c WHERE c.name = 'Ad Hoc Distributed Queries') = 0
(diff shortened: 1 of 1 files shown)


BlitzErik:
This is an interesting one!

I added a couple parameters:
@UserStoredProc
@UserAdHoc

These accept either a fully qualified stored procedure name (dbname.schemaname.sp_name) =, or fully qualified ad hoc queries (select * from dbname,schemanme.table_or_view_name). You can't test both, and I added a catch for that.

There's also logic to change some other configuration options if either of the new parameters is not null.

Added logic to skip the WAITFOR if either of the new parameters is not null, and start a new branch of code that creates a random temporary stored procedure based on code [from here](http://stackoverflow.com/questions/653714/insert-results-of-a-stored-procedure-into-a-temporary-table). Changes configuration options to allow for ad hoc distributed queries, dumps code results into a temp table via OPENROWSET.

I encountered one divide by zero error when testing code that runs almost instantly, so there are sure to be more. Adding isnull/nullif to every piece of division might be overkill, so let's leave it for testing. Also hoping no one uses this to get stats on code that finishes instantly. Right?

So, downside:
Terribly sql injection prone if you're using certs to give devs elevated access to execute. Otherwise, you're running this under an account that already has the ability to screw things up immensely.

This is ***READY FOR TESTING*** 

**_This pull request is not ready for merging yet. It was made to give it visibility._** 

Sample code to run:

Test stored proc
```
CREATE   PROCEDURE [dbo].[GetUserCrap] @UserId INT = 0
AS
BEGIN

DECLARE @TodaysDate DATETIME2 = SYSDATETIME()

SELECT [p].[OwnerUserId], [p].[LastActivityDate]
FROM [dbo].[Posts] AS [p]
WHERE [p].[OwnerUserId] = @UserId
AND [p].[LastActivityDate] <= @TodaysDate

END

GO
```

```
EXEC dbo.sp_BlitzFirst
    @UserStoredProc = 'StackOverflow.dbo.GetUserCrap 100',
```

Test ad hoc
```
EXEC dbo.sp_BlitzFirst
    @UserAdHoc = 'SELECT * FROM StackOverflow.dbo.Users AS u WHERE u.Id >= 1 and u.Id < 10000 SELECT * FROM StackOverflow.dbo.Posts AS p WHERE p.Id >= 1 and p.Id < 10000'
```

Fails if both aren't null
```
EXEC dbo.sp_BlitzFirst
    @UserStoredProc = 'StackOverflow.dbo.GetUserCrap 100',
    @UserAdHoc = 'SELECT * FROM StackOverflow.dbo.Users AS u WHERE u.Id >= 1 and u.Id < 10000 SELECT * FROM StackOverflow.dbo.Posts AS p WHERE p.Id >= 1 and p.Id < 10000'
```

BlitzErik:
Sample screen shots:

image

image

image

image

BlitzErik:
Sample screen shots with ad hoc:

image

image

image

BrentOzar:
I tried it on a box that doesn't have Ad Hoc Distributed Queries enabled, and it's trying to run this:

CREATE PROC #47CC531B19F74F219E5B07CAF3446CE4
		  AS
			BEGIN 
  EXEC sp_configure 'Ad Hoc Distributed Queries', 1
					RECONFIGURE 
			SELECT *
					INTO #sp_BlitzFirst_Temp 
					FROM OPENROWSET('SQLNCLI', 
									'Server=SQL2008R2PROD1;Trusted_Connection=yes;',
									'SELECT * FROM sys.databases') 
  EXEC sp_configure 'Ad Hoc Distributed Queries', 0
					RECONFIGURE 
	END;

That doesn't work:

Msg 15281, Level 16, State 1, Procedure #47CC531B19F74F219E5B07CAF3446CE4, Line 6 [Batch Start Line 0]
SQL Server blocked access to STATEMENT 'OpenRowset/OpenDatasource' of component 'Ad Hoc Distributed Queries' because this component is turned off as part of the security configuration for this server. A system administrator can enable the use of 'Ad Hoc Distributed Queries' by using sp_configure. For more information about enabling 'Ad Hoc Distributed Queries', see "Surface Area Configuration" in SQL Server Books Online.

BrentOzar:
Known limitations (will edit this as I find stuff):

* All output fields have to be named (you can't do a SELECT COUNT(*) for example, or SELECT GETDATE() without naming it)

PR #645, opened 2016-12-13
sp_BlitzFirst & sp_BlitzWho - Allow for sleeping SPIDs

### LINKED ISSUE #576: BlitzWho -- option to show sleeping SPIDs
Feature

Right now, we automatically filter out sleeping SPIDs

Add a flag to show them if you suspect sleeping sessions are causing issues

All

RichBenner:
Fixes #576  .

 - Alter sp_BlitzWho to allow for a paramater to show sleeping SPIDs
 - Also alter sp_BlitzFirst to allow for this parameter to be used when it calls sp_BlitzWho

How to test this code:
 - For sp_BlitzWho run with and without new parameter to see that sleeping SPIDs appear when using it;

     EXEC sp_BlitzWho @ShowSleepingSPIDs = 1

 - For sp_BlitzFirst it requires expert mode to call sp_BlitzWho. Run with and without both of the parameters below. Also check when one is set to 1 and the other to zero

    EXEC sp_Blitzfirst @ExpertMode = 1, @ShowSleepingSPIDs = 1

 - SQL Server 2012 Enterprise (11.0.6020.0) 
 - SQL Server 2014 Enterprise (12.0.4213.0)
 - SQL Server 2016 Developer (13.0.4001.0)

Note: paramater uses TINYINT in declaration but we're having to convert to NVARCHAR(1) in order to use in the dynamic SQL. I'm reticent to do it (because it's a number) but would it be better as NVARCHAR(1) in the first place?

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -22,7 +22,7 @@ ALTER PROCEDURE [dbo].[sp_BlitzFirst]
     @CheckProcedureCache TINYINT = 0 ,
     @FileLatencyThresholdMS INT = 100 ,
     @SinceStartup TINYINT = 0 ,
-	@ShowSleepingSPIDs TINYINT = 1 ,
+	@ShowSleepingSPIDs TINYINT = 0 ,
     @VersionDate DATETIME = NULL OUTPUT
     WITH EXECUTE AS CALLER, RECOMPILE
 AS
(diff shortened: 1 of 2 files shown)


BrentOzar:
Looks good! Good work sir, and thanks. This will make the January release.

PR #647, opened 2016-12-13
sp_BlitzFirst - Check if CPU >= 25% for non SQL Server processes

### LINKED ISSUE #444: sp_BlitzFirst - alert when other processes are using >25% CPU
In the CPU checks, add a check for non-SQL-Server processes using >25% of CPU during the sample. (Can get this from the ring buffer in the same place I'm getting CPU usage from, I think.)

RichBenner:
Fixes #444  .

 - Add check 28 to see if CPU is over 25% for non SQL Server processes.
 - Add check to documentation

How to test this code:
 - Run a CPU intensive process on the SQL Server box and check in task manager that it's using over 25% of the available CPU. Then run sp_BlitzFirst with @seconds < 30 and this should appear as a level 50 error
 - Alternatively, change the threshold down to 1% (on line 977) to show the issue appearing as per below;

image

 - SQL Server 2012 Enterprise (11.0.6020.0)
 - SQL Server 2014 Enterprise (12.0.4213.0)
 - SQL Server 2016 Developer (13.0.4001.0)


BrentOzar:
Great idea, and it works perfectly!

And I gotta share how I figured out to test it - I don't have any high-CPU-using things on my VMs, so I did this in a couple windows of SSMS:

`WHILE 1 = 1 SELECT * FROM sys.messages;`

Presto:



This will be in the January release.

PR #649, opened 2016-12-13
Fixes error with database_id in exec sessions

### LINKED ISSUE #648: sp_BlitzWho -- database id should come from exec requests, not exec sessions
Bug

database_id is retrieved from  exec sessions, which apparently doesn't that that column prior to 2012(?)

Should get it from exec requests instead. It's always been there, afaik.

< 2012

BlitzErik:
Closes #648


BrentOzar:
Hooah! Thanks, sir. This will be in the January release.

PR #654, opened 2016-12-18
Blitzcache 633

### LINKED ISSUE #633: sp_BlitzCache - [DISCUSS] get number of NC index updates under a clustered index update?
Just a couple ideas to kick around. 

We could warn when a clustered index update has nonclustered index updates underneath it if there are more than... some number?

image

image

There could also be a warning for when updates get a wide plan, like this:

image

There are a bunch of RelOp nodes with this:

image

**Questions** 
How many NC index updates would make a good minimum threshold to warn about?

What would the warnings be?

What would we suggest as a resolution, or what would we expect a user to do with that information?

### LINKED ISSUE #652: sp_BlitzCache -- Warn for filters referencing scalar valued functions
I'm a broken record on this stuff, but I keep finding it. 

SVFs in Filter operators as catchable:

image

The path will be something like Relop/Filter/Predicate/UserDefinedFunction

Can probably use the code to find them in Compute Scalars as a base.

### LINKED ISSUE #653: sp_BlitzCache -- Session data will be analyzed across different users
If you run sp_BlitzCache in one SSMS window, and then right after in another window, the second session will analyze the first session's global temp table data.

That thing up there

We have a SPID column in our global temp tables that analysis queries should be filtering on

All

BlitzErik:
Closes #653
--Session data read across executions 

Closes #633 
--Adds check for number of indexes modified

Closes #652
--Adds checks for scalar UDFs in filters

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -1682,6 +1682,7 @@ RAISERROR(N'Attempting to aggregate stored proc info from separate statements', 
         SUM(b.LastReturnedRows) AS LastReturnedRows
     FROM ##bou_BlitzCacheProcs b
     WHERE b.QueryHash IS NOT NULL 
+	AND b.SPID = @@SPID
     GROUP BY b.SqlHandle
 )
 UPDATE b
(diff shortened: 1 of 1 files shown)


BlitzErik:
Notes on testing is in each issue.

Tested in the lab 2005-2014, locally on 2016.

BrentOzar:
Great job! This will be in the January release.

PR #664, opened 2016-12-20
Blitz 602

### LINKED ISSUE #602: sp_Blitz - more CPU deformity checks
Feature; idea from https://github.com/Microsoft/tigertoolbox/blob/master/BPCheck/Check_BP_Servers.sql

We get some CPU info and warn if they're offline. There are additional checks we could do to see if nodes have a single core, or odd number of cores, etc.

Diagnose more CPU maladies

The Tiger Toolbox is licensed under MIT's license. The code involved is here:

```
	WITH ncpuCTE (ncpus) AS (SELECT COUNT(cpu_id) AS ncpus from sys.dm_os_schedulers WHERE is_online = 1 AND scheduler_id < 255 AND parent_node_id < 64 GROUP BY parent_node_id, is_online HAVING COUNT(cpu_id) = 1),
	cpuCTE (node, afin) AS (SELECT DISTINCT(parent_node_id), is_online FROM sys.dm_os_schedulers WHERE scheduler_id < 255 AND parent_node_id < 64 GROUP BY parent_node_id, is_online)
	SELECT 'Processor_checks' AS [Category], 'Affinity_NUMA' AS [Check],
		CASE WHEN (SELECT COUNT(*) FROM ncpuCTE) > 0 THEN '[WARNING: Current NUMA configuration is not recommended. At least one node has a single assigned CPU]' 
			WHEN (SELECT COUNT(DISTINCT(node)) FROM cpuCTE WHERE afin = 0 AND node NOT IN (SELECT DISTINCT(node) FROM cpuCTE WHERE afin = 1)) > 0 THEN '[WARNING: Current NUMA configuration is not recommended. At least one node does not have assigned CPUs]' 
			ELSE '[OK]' END AS [Deviation]
	FROM sys.dm_os_sys_info (NOLOCK) 
	OPTION (RECOMPILE);
```

BlitzErik:
Closes #602 
Tested in lab 2008-2014
Locally 2016

### DIFF
--- a/Documentation/sp_Blitz Checks by Priority.md
+++ b/Documentation/sp_Blitz Checks by Priority.md
@@ -26,6 +26,7 @@ If you want to change anything about a check - the priority, finding, URL, or ID
 | 10 | Performance | Auto-Close Enabled | http://BrentOzar.com/go/autoclose | 12 |
 | 10 | Performance | Auto-Shrink Enabled | http://BrentOzar.com/go/autoshrink | 13 |
 | 10 | Performance | CPU Schedulers Offline | http://BrentOzar.com/go/schedulers | 101 |
+| 10 | Performance | CPU w/Odd Number of Cores | http://BrentOzar.com/go/oddity | 198 |
 | 10 | Performance | High Memory Use for In-Memory OLTP (Hekaton) | http://BrentOzar.com/go/hekaton | 145 |
 | 10 | Performance | Memory Nodes Offline | http://BrentOzar.com/go/schedulers | 110 |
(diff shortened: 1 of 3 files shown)


BlitzErik:
This was added with CheckID 198. It was free in the checklist.

BrentOzar:
I added the shortcode http://BrentOzar.com/go/oddity to the code & documentation, but it's not linked to anything. Can you point that to a page by MS explaining why it's a bad thing, or if not, create one? Then go into BrentOzar.com, Tools, GoCode and add that short link to point to the URL. (I know this is gonna be the first question we get.)

BlitzErik:
Page and Go code added. Couldn't find any MS documentation. I kept it short, with a couple useful links.

Issue #671, opened 2016-12-23
sp_blitzIndex Exclusion ReportServer and ReportServerTempDB not always correct


JanSchreuder:
When a database is configured with named instances (eg: MyServer\INS01) then the ReportServer database might look like this ReportServer$INS01, ReportServer$INS01TempDB. In those cases the following clause will not be enough:
        AND DB_NAME(database_id) NOT IN ('ReportServer','ReportServerTempDB')

Replacing it with the following code resolved it for me:

        AND DB_NAME(database_id) NOT like 'ReportServer%'

BrentOzar:
Oh, cool! Great point, and yeah, let's go for it.

Here's how to contribute the code:
https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/CONTRIBUTING.md

If you don't want to make that change yourself (and I'd totally understand, your first contribution can be kinda intimidating because Github isn't exactly easy to use), just let me know and I'll do it. Just wanted to give you the chance to get your name into lights in the release notes first.

JanSchreuder:
Please make the change :-) The thought of my name in the release notes is tempting. But Git is not my thing, we use VSTS. I'm happy enough to contribute in this way

BrentOzar:
OK, done, thanks!

PR #672, opened 2016-12-27
#671 sp_BlitzIndex now ignores more ReportServer% databases

BrentOzar:
To catch named instances of Reporting Server.


BrentOzar:
Closes #672.

Issue #677, opened 2017-01-11
sp_Blitz - Query Store Cleanup check not checking build numbers


BrentOzar:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Check 182 looks for Query Store Cleanup Disabled by using this code, and it doesn't look at the build number:

						IF NOT EXISTS ( SELECT  1
										FROM    #SkipChecks
										WHERE   DatabaseName IS NULL AND CheckID = 182 )
                            AND EXISTS(SELECT * FROM sys.all_objects WHERE name = 'database_query_store_options')
							AND CAST(SERVERPROPERTY('edition') AS VARCHAR(100)) NOT LIKE '%Enterprise%'
							AND CAST(SERVERPROPERTY('edition') AS VARCHAR(100)) NOT LIKE '%Developer%'

**What is the expected behavior?**
Look at the build number too to make sure they're on a fixed build. (This check also shouldn't return a row per database.)

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
SQL 2016 SP1 Standard Edition.

Issue #650, opened 2016-12-14
Standard edition vs Enterprise edition


HenrikStaunPoulsen:
**Do you want to request a *feature* or report a *bug*?**
this must be considered a bug now.

**What is the current behavior?**
sp_Blitz reports a warning; The [mydb] database is using Compression.  If this database is restored onto a Standard Edition server, the restore will fail.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via http://sqlfiddle.com
run Exec sp_Blitz   :-)

**What is the expected behavior?**
No warning if SQL Server is 2016 SP1 and above

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
only SQL Server 2016 SP1 affected so far. It works in old versions.

BlitzErik:
I think for this we'd just change the wording of the message.

BlitzErik:
Closed in #683

HenrikStaunPoulsen:
Hi Erik, 
Thank you very much for fixing this.
Best regards,
Henrik

BlitzErik:
My pleasure!

Issue #628, opened 2016-12-07
sp_Blitz - Product Version Checks - Bug?


RichBenner:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

We check if @ProductVersionMajor >= 10 and @ProductVersionMinor >= 50 in a lot of places. It looks like this should be looking for anything on 2008 R2 and newer but this isn't how the minor product version is working. 

For SQL Server 2016 SP1 (13.00.4001.00) the major version is 13.00 and the minor version is 4001.00. The only version of SQL Server with a minor version less than 50 was build 8.00.047 which was for SQL Server 2000 EAP4 (source: https://sqlserverbuilds.blogspot.co.uk/). 

Is this check working correctly (checking for SQL 2008+ boxes) or should it be checking for 2008 R2 and newer? If the former, do we even need the minor version check?

To find these lines search for 'IF @ProductVersionMajor >= 10 AND @ProductVersionMinor >= 50 ' in sp_Blitz.

BlitzErik:
Oh, barf, those are all mine. What the heck was on my mind? Or rather, my clipboard?

Yeah, all the DMVs it's hitting are 2008 and up. Given the amount of 2008 boxes this has been run on, I'd say they're "working" but, yeah.

RichBenner:
Worth a quick fix to remove the minor version check or leave as-is?

BlitzErik:
I'll take care of this one. It's my goof. I can test it against our 2008 instance to make sure it doesn't break anything.

Thanks!

PR #683, opened 2017-01-16
sp_Blitz 628

BlitzErik:
Closes three issues:

#628: useless version checks
#650: Wording for EE feature checks
#677: Insufficient checks for broken QS cleanup


BlitzErik:
This was tested in AWS on 2008 - 2014, and 2016 locally.

PR #705, opened 2017-02-09
#670 sp_Blitz recompiles to avoid plan cache bloat

### LINKED ISSUE #670: sp_Blitz -- Should recompile at proc level, and hints need to go in dynamic sql
Change. BlitzIndex and BlitzCache both recompile at the proc level and in dynamic sql. Blitz should too. Now I'm thinking about BlitzFirst and I'm worried.

We don't recompile stuff. We shouldn't be adding to the plan cache etc.

RECOMPILE GOD AREN'T YOU LISTENING?

All

BrentOzar:
Slapped recompile on the proc and in all the dynamic SQL. Closes #670.


BrentOzar:
I wouldn't close my own PRs, but this one touches so many lines of the code that I wanna get it into the dev branch in case anybody plays around with the code when today's blog post launches.

PR #745, opened 2017-03-03
sp_foreachdb - create + alter procedure pattern

### LINKED ISSUE #743: sp_foreachdb, sp_BlitzTrace - drop + create procedure instead of create if not exists + alter procedure
Bug? More inconsistent behavior with the potential for errors in the right conditions. (See http://www.codeofhonor.com/blog/a-better-way-to-update-sql-stored-procedures)

sp_foreachdb & sp_blitztrace both follow the "drop procedure then create procedure" pattern for deployment of updated versions of sp. 

sp_blitz, et. al. employ a "create if not exists and then alter procedure" pattern.  

dnc-msimons:
Fixes #743 for sp_foreachdb.

switching to "create if not exists and then alter procedure" pattern instead of drop/create
How to test this code:

Tested in an scenario where the procedure doesn't exist yet and then also when it already does.

SQL Server 2008 R2

### DIFF
--- a/sp_foreachdb.sql
+++ b/sp_foreachdb.sql
@@ -1,11 +1,11 @@
 USE [master];
 GO
 
-IF OBJECT_ID('dbo.sp_foreachdb') IS NOT NULL
-    DROP PROCEDURE dbo.sp_foreachdb
+IF OBJECT_ID('dbo.sp_foreachdb') IS  NULL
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks sir!

dnc-msimons:
You're welcome.  And thanks for all you do too!

PR #767, opened 2017-03-09
Issue 766/movefiles fails with long paths

### LINKED ISSUE #766: DatabaseRestore - Restores fail when using @MoveFiles=1 with long paths / multiple files

Report a bug

When restoring a database that is backed up to multiple files, the restore fails if you use @MoveFiles=1 and the combined length of the new paths is over 1,000 characters. The error message will look something like this, although it will vary depending on the exact file names you use:

> Msg 105, Level 15, State 1, Line 1
> Unclosed quotation mark after the character string 'C:\Pr

Backup up a database to multiple files (i.e. 8) to a directory with a relatively long path (i.e. C:\Program Files\Microsoft SQL Server\MSSQL13.MSSQLSERVER\MSSQL\Backup). Then attempt to restore using @MoveFiles=1.

The @MoveFiles=1 option should be able to accommodate long paths and multiple files.

All supported versions, which at present is just SQL Server 2016.

jeffrosenberg:
Fixes #766 (DatabaseRestore - Restores fail when using @MoveFiles=1 with long paths / multiple files).

 - Remove 4 unused variables
 - Change @MoveOption variable to nvarchar(MAX)

How to test this code:
 - Backup up a database to multiple files (i.e. 8) to a directory with a relatively long path (i.e. C:\Program Files\Microsoft SQL Server\MSSQL13.MSSQLSERVER\MSSQL\Backup). Then attempt to restore using @MoveFiles=1. This should now work.

 - SQL Server 2016

### DIFF
--- a/DatabaseRestore.sql
+++ b/DatabaseRestore.sql
@@ -47,8 +47,6 @@ SET NOCOUNT ON;
 DECLARE @cmd NVARCHAR(4000), @sql NVARCHAR(MAX), @LastFullBackup NVARCHAR(500), @BackupFile NVARCHAR(500);
 DECLARE @FileList TABLE (BackupFile NVARCHAR(255));
 
-DECLARE @MoveDataLocation AS NVARCHAR(500), @MoveDataLocationName AS NVARCHAR(500), @MoveLogLocation AS NVARCHAR(500), @MoveLogLocationName AS NVARCHAR(500);
-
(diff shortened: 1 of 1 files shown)


jeffrosenberg:
I'll hopefully stop bombarding you with pull requests for this procedure soon. What I really wanted to do was to work on making it compatible with versions earlier than 2016, but I kept running into bugs! :grimacing:

PR #772, opened 2017-03-10
Issue 221

### LINKED ISSUE #221: sp_BlitzIndex - add OutputDatabaseName, SchemaName, TableName
Write the output to table like we do with sp_Blitz and sp_BlitzFirst. Can rip the code straight out of there.

rwhoward:
Fixes #221  .

 - Added Output parameters to sp_BlitzIndex to send Mode 2 results to permanent repository

How to test this code:
 - EXEC sp_BlitzIndex @Mode=2, @OutputDatabaseName='dbName', @OutputSchemaName='schemaName', @OutputTableName='tableName', @OutputServerName='linkedServerName'
- OutputServerName is optional, the rest are all or nothing
- Works with global temp tables (##) but not local temp tables (#). No temp tables when using linked server.
- Only works on Mode 2 for now
- Should error out if DB, schema, table, or linked server are invalid or are not all provided

 - SQL Server 2008
 - SQL Server 2014
 - SQL Server 2016
 - SQL Server 2016 reporting to linked SQL 2008
 - SQL Server 2008 reporting to linked SQL 2016

Notes:
I hate building dynamic SQL using a bunch of string concatenation. My opinion - it makes for complicated copy/paste and sections of this will require copy/paste each time it is updated with new checks and columns. Right now it's set to do a REPLACE on variable names. I think it is easier to read and maintain. Definitely worse performance, but this isn't a high volume procedure so I don't think that matters.
Obviously that makes this pattern different than most of the other code in here though. So if it's going to bother people let me know and I'll make it fit the existing pattern.


BrentOzar:
Looks good! Thanks for the detailed explanation. I'm fine with the variable approach here, and I agree that it doesn't run often enough where the performance would be an issue. Good work!

BlitzErik:
Howdy @rwhoward!

When I run BlitzIndex in Mode 2, I get an error:

image

Since that's a custom RAISERROR you wrote in, any chance you could point me in the right direction for debugging?

Thanks!

rwhoward:
That was kind of a generic catch-all error. It should only be thrown if someone was attempting to output the results if either some or all of the output destination parameters were invalid or if some were specified and others were missing. LIke if someone ran EXEC sp_BlitzIndex @Mode=2, @OutputDatabaseName='IndexStuff' but didn't set an output table or schema.

Looks like something changed between when I merged dev into my branch and went Brent merged my branch into dev. My version can run EXEC sp_BlitzIndex @Mode=2 but the dev branch post-merge can't. I'll do some digging to see if it needs more of a fix than what you put in to get it shipped. I want to make sure there aren't other downstream problems caused by the same mystery change.

rwhoward:
Never mind, I'm just being dumb. When testing just now I was using the wrong branch - my version never worked right. Apparently I never thought to make sure that running it without any output worked. 

But I think what you changed was all it needed - just removing the manual RAISERROR. Mystery problem solved.
Note to future self: Make sure test plan includes the 99% use case.

Issue #792, opened 2017-04-01
sp_Blitz -- CheckID 100 not checking for all prerequisites when using Express Edition


Totty1973:
**Do you want to request a *feature* or report a *bug*?**
Small Bug / Feature Request

**What is the current behavior?**
CheckID only checks sys.configurations WHERE   name = 'remote admin connections' AND value_in_use = 0 
it says in the SQL server errorlog "Dedicated administrator connection support was not started because it is disabled on this edition of SQL Server. If you want to use a dedicated administrator connection, restart SQL Server using the trace flag 7806. This is an informational message only. No user action is required."

**What is the expected behavior?**
It could check for TF7806, because this is required when running Express Edition.
https://www.mssqltips.com/sqlservertip/2538/enabling-dedicated-administrator-connection-in-sql-server-2008-express-edition/
Maybe CheckID 74 should report a missing TF7806 when running express edition

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Tested only on 2012sp3express which seems affected.

BlitzErik:
Howdy -- It's been a month, gonna close this down. If you want to pick it back up, please reopen it.

PR #795, opened 2017-04-02
Update #792 to SP_BLITZ, because SQL express editions need an extra traceflag to fully enable DAC

Totty1973:
Update #792 to SP_BLITZ, because SQL express editions need an extra traceflag to fully enable DAC

Fixes # 792

 - new warning 'Remote DAC switched on but needs traceflag 7806'  on express edition

How to test this code:
 - have enabled TF7806 and DAC enabled when using Express edition: no warning
 - have disabled TF7806 (default) and DAC enabled when using Express edition: new warning 'Remote DAC switched on but needs traceflag 7806'  
 - have DAC disabled: only old warning

 - SQL 2012 sp3 express
 - SQL 2014 enterprise

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -2482,7 +2482,6 @@ AS
 										AND j.category_id <> 100 /* Exclude SSRS category */
 					END
 
-
 				IF EXISTS ( SELECT  1
 							FROM    sys.configurations
 							WHERE   name = 'remote admin connections'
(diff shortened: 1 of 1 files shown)


Totty1973:
I don't see how populatin #tracestatus just once can be achieved if a user can elect not to run all checks

BrentOzar:
By doing it once at the beginning of the proc, or prior to the check you're doing.

I totally understand where you're coming from, I just don't want to redo work in the proc if necessary. A good example might be doing your code later in the proc, after #TraceStatus has been populated, and then checking to see if #TraceStatus already has data in it. If not, run your query.

Totty1973:
> after #TraceStatus has been populated, 
It could be empty because the user choose not to enable any flags, my 2014web-instance does not have any flags set. 
I do not see any flag in version 20170307 saying "i did my query but i still may be empty". Is that the best way?
I could move the procedure to fill the #tracestatus into a global section, so that it gets executed every single time. is that the best way?

BlitzErik:
Haven't seen anything on this in a while, so going to close out for now. If you want to work on it in the future, you can reopen.

Thanks!

PR #839, opened 2017-04-07
Check for plans with >128 levels of nesting

### LINKED ISSUE #838: sp_BlitzCache - check for plans in sys.dm_exec_text_query_plan
Feature

If a plan isn't in sys.dm_exec_query_plan, we say it's missing. However, it might just be too big for SQL Server to render as XML. Let's check for the existence of it in sys.dm_exec_text_query_plan, and just tell the user to go get it from there, but that we haven't done any analysis of it because it's so big.

BlitzErik:
Closes #838


BrentOzar:
Love it.

PR #840, opened 2017-04-07
Issue 814 - Added Differential Restore Feature to sp_DatabaseRestore 

### LINKED ISSUE #814: DatabaseRestore -- Add Differential Restores
**Do you want to request a *feature* or report a *bug*?** Feature

**What is the current behavior?** Will restore Full backups, and Log backups, but not differential

**What is the expected behavior?** After restoring full backups, restore the differential backups and then the log backups.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?** All

digitalohm:
Fixes #814 

added ContinueDiff bit parameter to signal that the procedure should
check for diffs

added LastDiffBackup variable and moved the BackupFile and
BackupDateTime variables to be global so that they can be used across
the full and diff functions

cut n pasted the full restore feature, modified it to handle diffs and
set conditional logic so that the most recent diff date is checked
against the full backup date, this will hopefully prevent a situation
where there are no diffs or diffs older than the full.

How to test this code:

EXEC dbo.sp_DatabaseRestore 
  @Database = 'YourDBName', 
  @BackupPathFull = '\\YourPath\FULL\', 
  @BackupPathDiff = '\\YourPath\DIFF\',
  @BackupPathLog = '\\YourPath\LOG\', 
  @RestoreDiff = 1,
  @ContinueLogs = 0, 
  @RunRecovery = 1;

 - SQL Server 2014
 - SQL Server 2012

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -1,33 +1,33 @@
 /*
-EXEC dbo.DatabaseRestore 
+EXEC dbo.sp_DatabaseRestore 
 	@Database = 'LogShipMe', 
 	@BackupPathFull = 'D:\Backup\SQL2016PROD1A\LogShipMe\FULL\', 
 	@BackupPathLog = 'D:\Backup\SQL2016PROD1A\LogShipMe\LOG\', 
 	@ContinueLogs = 0, 
 	@RunRecovery = 0;
 
-EXEC dbo.DatabaseRestore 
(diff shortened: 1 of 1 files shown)


dbassassin:
Updated branch to also address #849

dbassassin:
Updated branch to address issues: #850 , #851 and to provide a fix to #814 so that the LSNs are checked against the differential file that gets restored.

Tested against: SQL Server 2014

PR #848, opened 2017-04-11
Table variables ftw

### LINKED ISSUE #847: sp_BlitzWho is slow af sometimes
Bugger

sp_BlitzWho is sometimes quite slow.

It shouldn't be. The big difference in this version is finding the additional blocking nonsense. That probably needs to get broken out to a temp table or table variable.

All

BlitzErik:
Fixing slow performance issue (maybe?).
Works but needs testing under more load.
Closes #847


BlitzErik:
Tested head to head vs. who is active with 100 blocking sessions. Both finish in ~1 second.

image

PR #878, opened 2017-05-15
Higher priority for maintenance plans and jobs with shrinkdatabase or shrinkfile commands but disabled

JohnKNess:
Fixes # 776

 - Higher priority (200) for maintenance plans and jobs with shrinkdatabase or shrinkfile commands whose schedules have been disabled or not run in the past 30 days
 - If the job or maintenance plan has a schedule then it will be displayed at the end of the Details column

`In the [Shrink Database AdminDB2] job, step [Shrink Database AdminDB2] has SHRINKDATABASE or SHRINKFILE, which may be causing database fragmentation. (Schedule: [TEST TJ Daily 07:00])`

How to test this code:
 - Run sp_Blitz against SQL Server instance with maintenance plans that contain SHRINK DATABASE building block  
 - Run sp_Blitz against SQL Server instance that has job(s) that contains a SHRINKDATABASE or SHRINKFILE command in the job step

 - SQL Server 2008
 - SQL Server 2008 R2
 - SQL Server 2012
 - SQL Server 2014

I currently don't have a SQL Server 2016 instance that I could test against.


sp_Blitz.sql
```diff
@@ -3687,13 +3688,36 @@ IF @ProductVersionMajor >= 10
...
+						+ CASE WHEN COALESCE(ssc.name,'0') != '0' THEN + ' (Schedule: [' + ssc.name + '])' ELSE + '' END AS [Details]
 						FROM [maintenance_plan_steps] [mps]
 							CROSS APPLY [maintenance_plan_xml].[nodes]('//dts:Executables/dts:Executable') [t]([c])
+                    	join msdb.dbo.sysmaintplan_subplans as sms 
+                    		on mps.id = sms.plan_id 
+                    	JOIN msdb.dbo.sysjobs j 
+                    		on sms.job_id = j.job_id
+                    	LEFT OUTER JOIN msdb.dbo.sysjobsteps AS step
+                    		ON j.job_id = step.job_id
+                    	LEFT OUTER JOIN msdb.dbo.sysjobschedules AS sjsc
```
BrentOzar: If a job has multiple schedules, I think this is going to produce multiple rows. Putting a select distinct on here won't fix it since the schedule is included in the details notes. If you wanna remove the schedule name from the details and then do a select distinct, I'm cool with that.

sp_Blitz.sql
```diff
@@ -3687,13 +3688,36 @@ IF @ProductVersionMajor >= 10
 										[Details] )									  
 						SELECT
 						180 AS [CheckID] ,
-						100 AS [Priority] ,
+						-- sp_Blitz Issue #776
+						-- Job has history and was executed in the last 30 days
+						CASE WHEN (cast(datediff(dd, substring(cast(sjh.run_date as nvarchar(10)), 1, 4) + '-' + substring(cast(sjh.run_date as nvarchar(10)), 5, 2) + '-' + substring(cast(sjh.run_date as nvarchar(10)), 7, 2), GETDATE()) AS INT) < 30) OR (j.[enabled] = 1 AND ssc.[enabled] = 1 )THEN
```
BrentOzar: Can we just do a single datediff here instead of string parsing? I hate to be picky, but I'm worried about how this is going to go down in international servers - just do a plain old datediff(dd, job date, today).

sp_Blitz.sql
```diff
@@ -5619,16 +5643,33 @@ IF @ProductVersionMajor >= 10
...
 								  Details
 								)
-								SELECT  79 AS CheckID ,
-										100 AS Priority ,
+								SELECT  DISTINCT 79 AS CheckID ,
+										-- sp_Blitz Issue #776
+										-- Job has history and was executed in the last 30 days OR Job is enabled AND Job Schedule is enabled
+                						CASE WHEN (cast(datediff(dd, substring(cast(sjh.run_date as nvarchar(10)), 1, 4) + '-' + substring(cast(sjh.run_date as nvarchar(10)), 5, 2) + '-' + substring(cast(sjh.run_date as nvarchar(10)), 7, 2), GETDATE()) AS INT) < 30) OR (j.[enabled] = 1 AND ssc.[enabled] = 1 )THEN
+                						    100
+                						ELSE -- no job history (implicit) or job not run in the past 30 days OR Job disabled OR Job Schedule disabled
```
BrentOzar: Small tweak - I think that ELSE description should be:

no job history (implicit) AND job not run in the past 30 days AND (Job disabled OR Job Schedule disabled)

Not positive though, correct me if I'm wrong.

sp_Blitz.sql
JohnKNess: Thank you for the feedback.

The column run_date (and run_time) is defined as int according to the TechNet Article [dbo.sysjobhistory (Transact-SQL)](https://technet.microsoft.com/en-us/library/ms174997(v=sql.110).aspx). 

Values in the column are stored for example as : 

- 20171201 
- 20170524
- YYYYMMDD

... without any dashes, stops, hyphens, etc.

If I perform a simple `select datediff(dd,run_date,GETDATE()) from msdb..sysjobhistory` on an English installation of SQL Server 2014 running on a German Windows 7 or on a German installation of SQL Server running on a German Windows 2012 Server then I will receive the following error message:

`Msg 8115, Level 16, State 2, Line 4
Arithmetic overflow error converting expression to data type datetime.`

(it get's even worse once you want to start analysing job durations and use the run_time column where `1` is a valid time)

If I don't parse it like I did, then it is possible for incorrect values to be returned or for errors to occur depending on language settings of the SQL Server. I have previously tried to CAST and CONVERT the run_date to a DATETIME format for previous projects, but it never seemed to work out pretty well.

I am open for other suggestions.

sp_Blitz.sql
JohnKNess: Yes, you are correct twice.

My query will produce results like this:

`CheckID	Priority	FindingsGroup	Finding	URL	Details
79	100	Performance	Shrink Database Job	https://BrentOzar.com/go/autoshrink	In the [shrink_anonymous_daily] job, step [shrink_it] has SHRINKDATABASE or SHRINKFILE, which may be causing database fragmentation. (Schedule: [night])
79	200	Performance	Shrink Database Job	https://BrentOzar.com/go/autoshrink	In the [shrink_anonymous_daily] job, step [shrink_it] has SHRINKDATABASE or SHRINKFILE, which may be causing database fragmentation. (Schedule: [day])
79	200	Performance	Shrink Database Job	https://BrentOzar.com/go/autoshrink	In the [anonymous_index_rebuild] job, step [shrink_first] has SHRINKDATABASE or SHRINKFILE, which may be causing database fragmentation. (Schedule: [index_rebuild.Subplan_1])
79	200	Performance	Shrink Database Job	https://BrentOzar.com/go/autoshrink	In the [anonymous_index_rebuild] job, step [shrink_last] has SHRINKDATABASE or SHRINKFILE, which may be causing database fragmentation. (Schedule: [index_rebuild.Subplan_1])`

My code will differentiate between each schedule as an individual finding, as they can be individually enabled/disabled depending on the phase of the moon or the constellation of the stars. Enabled and running schedules will have a priority of 100 and disabled and no longer running schedules will have a priority of 200.

...and, the DISTINCT can be removed and will be removed.

(I was lucky with this one, because I had a SQL Server that was being maintained by a 3-rd party DBA who had installed multiple shrink jobs. Some schedules disabled, some enabled and with various schedules.)

The DISTINCT was possibly a leftover from testing. I'm sorry for that one.

sp_Blitz.sql
JohnKNess: Yup. Fixed.

BlitzErik:
@BrentOzar is this one good to go now?

BrentOzar:
Yep, looks good!

BlitzErik:
Hahaha, alright. It'll be in next month.

JohnKNess:
Awesome. Thanks. My first contribution.

BlitzErik:
Now you get to support it forever :D

Issue #906, opened 2017-05-30
sp_BlitzBackups Error


jsetliffe:
All, I am getting an error when running sp_BlitzBackups on my production SQL 2012 ( build 11.0.6544.0) instance.  It works fine on my QA one, same build number.  What could be the issue?  Here is the error:

Inserting to #Backups
Updating #Backups with worst RPO case
Warning: The join order has been enforced because a local join hint is used.
Updating #Backups with worst RPO case queries
Gathering RTO information
Updating #RTORecoveryPoints
Add any full backups in the StartDate range that weren't part of the above log backup chain
Fill out the most recent log for that full, but before the next full
Fill out a diff in that range
Get time & size totals for full & diff
Get time & size totals for logs
Gathering RTO worst cases
Returning data
Rules analysis starting
Warning: Null value is eliminated by an aggregate or other SET operation.
Msg 105, Level 15, State 1, Line 7
Unclosed quotation mark after the character string ' backups, and the last time a certificate was backed up is Aug 24 2015  6:14PM. AS [Warning]
							FROM   [msdb].dbo.backupset AS b
							WHERE b.encryptor_type IS NOT NULL
							GROUP BY b.database_name, b.encryptor_type;
'.
Msg 102, Level 15, State 1, Line 7
Incorrect syntax near ' backups, and the last time a certificate was backed up is Aug 24 2015  6:14PM. AS [Warning]
							FROM   [msdb].dbo.backupset '.
Rules analysis starting on temp tables
Rules analysis finished

For both instances I run sp_BlitzBackups with the master database selected.

BlitzErik:
Howdy,

Can you run it with @Debug = 1 on, and paste in the query the full text of the query it errors out on? Looks like CheckId 9.

Thanks!

jsetliffe:
[debug_for_Erik.txt](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/files/1039115/debug_for_Erik.txt)
error file attached

BlitzErik:
Thanks for the quick reply!

Can you try the version here? 

https://raw.githubusercontent.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/sp_BlitzBackups_906/sp_BlitzBackups.sql

jsetliffe:
Different error this time:
Inserting to #Backups
Updating #Backups with worst RPO case
Warning: The join order has been enforced because a local join hint is used.
[...]

PR #907, opened 2017-05-30
Sp blitzbackups 906

BlitzErik:
Fix dynamic SQL looking for encrypted backups
Fix product version variable population
Fix check for encrypted backups to only run on 2014+
Fix dynamic SQL to add the encryptor_type column to the fake msdb table if it doesn't already exist (fun fact: try/catch doesn't work with alter table)
Fix dynamic SQL in data mover to exclude the encryptor_type column based on production version


BlitzErik:
Closes #906

Issue #536, opened 2016-10-16
sp_foreachdb - handle database names the same way as sp_MSforeachdb


BrentOzar:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
Run this:
EXEC dbo.sp_MSforeachdb @command1 = 'USE [?]; SELECT DB_NAME();'
and it works fine. Run this:
EXEC dbo.sp_foreachdb @command = 'USE [?]; SELECT DB_NAME();'

And you get errors:

```
Msg 105, Level 15, State 1, Line 1
Unclosed quotation mark after the character string '[AdventureWorks2014]; SELECT DB_NAME();'.
Msg 102, Level 15, State 1, Line 1
Incorrect syntax near '[AdventureWorks2014]; SELECT DB_NAME();'.
Msg 105, Level 15, State 1, Line 1
Unclosed quotation mark after the character string '[ClearTrace]; SELECT DB_NAME();'.
Msg 102, Level 15, State 1, Line 1
Incorrect syntax near '[ClearTrace]; SELECT DB_NAME();'.
```

It looks like sp_foreachdb is adding brackets. This works fine:
EXEC dbo.sp_foreachdb @command = 'USE ?; SELECT DB_NAME();'

We should probably work the same way sp_MSforeachdb does by default.

t0mburton:
I was just caught out by this myself. I've defaulted the suppress_quotename param to true in this [commit](https://github.com/t0mburton/SQL-Server-First-Responder-Kit/commit/755ef74dead74e8dba265fe73c30d3b297a243bf).

I can submit a PR if you would like or were you looking for a different solution?

PR #912, opened 2017-05-31
Default @suppress_quotename to true

t0mburton:
Fixes [#536](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/536)

Bring sp_foreachdb inline with sp_MSforeachdb. Default @suppress_quotename param to true.

How to test this code:
EXEC dbo.sp_foreachdb @command = 'USE [?]; SELECT DB_NAME();' should work the same as
EXEC dbo.sp_MSforeachdb @Command1 = 'USE [?]; SELECT DB_NAME();'

 - SQL Server 2008
 - SQL Server 2008 R2
 - SQL Server 2012
 - SQL Server 2014

### DIFF
--- a/sp_foreachdb.sql
+++ b/sp_foreachdb.sql
@@ -10,7 +10,7 @@ CREATE PROCEDURE dbo.sp_foreachdb
     @replace_character NCHAR(1) = N'?' ,
     @print_dbname BIT = 0 ,
     @print_command_only BIT = 0 ,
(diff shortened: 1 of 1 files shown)


BlitzErik:
I'm not sure this is quite what we had in mind. Can you describe your changes a bit?

t0mburton:
No problem, like I said probably better solutions but this did the trick for me use case. Will close the PR.

PR #927, opened 2017-06-05
Adds tempdb allocations

BlitzErik:
Yet another outer apply


BlitzErik:
To test this branch:

Run this in one window:

```
SELECT p.Body
INTO #crap
FROM dbo.Users u
JOIN Posts p
ON p.OwnerUserId = u.Id

DROP TABLE #crap
```

Run this in another window:

`sp_BlitzWho`

And then you should see something like this:

image

And it grows over time!

image

PR #938, opened 2017-06-14
Sp databaserestore erikpoking

BlitzErik:
FIVE SCOOPS C'MON


digitalohm:
@BlitzErik this smells great!  It's a good example and notes for me to use moving forward.  Sorry I didn't get to do some of the cleanup myself, the real world called :(

BlitzErik:
@digitalohm no problem at all. That's what slackers like me are here for!

PR #944, opened 2017-06-16
Closes #937

### LINKED ISSUE #937: sp_BlitzIndex -- Take table creation date into missing index benefit calculation
When we judge missing index benefit, we look at the 'magic benefit number' divided by days of uptime. This makes sense as long as the table has been there the whole time, but less sense if it was created more recently.

We collect `create_date` in #IndexSanity, which is where we grab missing index details from. To avoid grouping awkwardness, I'm just going to grab `ISNULL(NULLIF(MAX(DATEDIFF(DAY, t.create_date, SYSDATETIME())), 0), 1)` here, and then in `WHERE` clause, it will look like

```
WHERE ( @Mode = 4 AND (magic_benefit_number / CASE WHEN create_days < @DaysUptime THEN create_days ELSE @DaysUptime END) >= 100000 ) 
OR (magic_benefit_number / CASE WHEN create_days < @DaysUptime THEN create_days ELSE @DaysUptime END) >= 100000
```

BlitzErik:
This changes the logic of the missing index filter to take into account
table creation time. If a table or partition was created more recently
than the last server reboot, we shouldn't judge missing index benefit
going that far back. For example, if someone uses partitioned views and
they have daily or weekly tables, etc.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1891,23 +1891,36 @@ BEGIN
     OPTION    ( RECOMPILE );                        
 
     IF (SELECT TOP 1 [object_id] FROM    #MissingIndexes mi) IS NOT NULL
-    BEGIN  
+    BEGIN
+
+	WITH create_date AS (
+						SELECT i.database_id,
+							   i.schema_name,
+							   i.[object_id], 
+							   ISNULL(NULLIF(MAX(DATEDIFF(DAY, i.create_date, SYSDATETIME())), 0), 1) AS create_days
+						FROM #IndexSanity AS i
+						GROUP BY i.database_id, i.schema_name, i.object_id
+						)
         SELECT  N'Missing index.' AS Finding ,
                 N'http://BrentOzar.com/go/Indexaphobia' AS URL ,
                 mi.[statement] + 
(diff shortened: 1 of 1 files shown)


sp_BlitzIndex.sql
```diff
@@ -1891,23 +1891,36 @@ BEGIN
...
-        WHERE   [object_id] = @ObjectID
-                /* Minimum benefit threshold = 100k/day of uptime */
-        AND (magic_benefit_number/@DaysUptime) >= 100000
+		LEFT JOIN create_date AS cd
+		ON mi.[object_id] =  cd.object_id 
+		AND mi.database_id = cd.database_id
+		AND mi.schema_name = cd.schema_name
+        WHERE   mi.[object_id] = @ObjectID
+                /* Minimum benefit threshold = 100k/day of uptime OR since table creation date, whichever is lower*/
+        AND (magic_benefit_number / CASE WHEN cd.create_days < @DaysUptime THEN cd.create_days ELSE @DaysUptime END) >= 100000
```
BrentOzar: Will this give a divide-by-zero error for newly created indexes? (Same thing on the other two divided-by-create-days changes below.

PR #1020, opened 2017-07-10
Adds debug modes to sp_Blitz

### LINKED ISSUE #1008: sp_Blitz -- Add a Debug mode
Troubleshooting errors in sp_Blitz is exceptionally difficult, especially when they crop up in dynamic SQL.

We should have a `@Debug` mode, which will, when set to 1:

Shoot out a RAISERROR noting the CheckId
If block is dynamic, print query
If block is dynamic but `@StringToExecute` is `NULL`, print a message that it's `NULL`

BlitzErik:
Closes #1008

--Print only CheckIds
`EXEC dbo.sp_Blitz @CheckUserDatabaseObjects = 1, @CheckServerInfo = 1, @Debug = 1`

--Print CheckIds and any dynamic SQL used in a variable. Some of these are not printed because they're passed in as strings to MS_foreachdb.
`EXEC dbo.sp_Blitz @CheckUserDatabaseObjects = 1, @CheckServerInfo = 1, @Debug = 2`

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -1,5 +1,5 @@
 IF OBJECT_ID('dbo.sp_Blitz') IS NULL
-  EXEC ('CREATE PROCEDURE dbo.sp_Blitz AS RETURN 0;')
+  EXEC ('CREATE PROCEDURE dbo.sp_Blitz AS RETURN 0;');
 GO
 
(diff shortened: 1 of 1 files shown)


BrentOzar:
This looks great! Nice work, sir. (And a lot of work it was.)

PR #1022, opened 2017-07-12
sp_Blitz - Fix for #962 - Failure of sys.fn_trace_gettable

### LINKED ISSUE #962: sp_Blitz - sys.fn_trace_gettable Failure
Bug

If you've been messing about with your trace files then sys.fn_trace_gettable is going to fail. Because we don't have error catching then the whole of sp_Blitz fails.

Delete one of the older trace files, to test you can use this extract of the code for testing;

DECLARE @base_tracefilename NVARCHAR(500),@curr_tracefilename NVARCHAR(500),@indx int

select @curr_tracefilename = [path] from sys.traces where is_default = 1 ;
set @curr_tracefilename = reverse(@curr_tracefilename);
select @indx = patindex('%\%', @curr_tracefilename) ;
set @curr_tracefilename = reverse(@curr_tracefilename) ;
set @base_tracefilename = left( @curr_tracefilename,len(@curr_tracefilename) - @indx) + '\log.trc' ;

BEGIN TRY
SELECT * FROM sys.fn_trace_gettable(@base_tracefilename, DEFAULT)
END TRY
BEGIN CATCH
SELECT 'Yep, its broken'
END CATCH

It would be nice if failure didn't result in the whole of sp_Blitz failing

I believe it's always going to have been a problem. People messing with trace files transcends SQL Server versions.

RichBenner:
Fixes #962  .

 - Added in a new temp table #fnTraceGettable so that we're not pulling data multiple times from sys.fn_trace_gettable
 - Also added @TraceFileIssue for error checking
 - Added a try/catch block to insert data into #fnTraceGettable from sys.fn_trace_gettable using default trace file. If this succeedes then set @TraceFileIssue = 0, if it fails set @TraceFileIssue = 1
 - Checks that used to use sys.fn_trace_gettable now use #fnTraceGettable.
 - Skip any checks that depend on #fnTraceGettable if @TraceFileIssue = 1
How to test this code:
 - On line 242 where we set @base_tracefilename, break the file name (e.g. change "+ '\log.trc' ;" to "+ '\log. trc' ;") by adding a space
 - Once that trace file name is broken then execute sp_Blitz, you should then get a new lvl 50 error for CheckID 199 that states there has been a problem with the log files.

 - Case-sensitive SQL Server instance (12.0.5203.0 - SQL 2014 SP2-GDR Enterprise Edition)
 - SQL Server 2014 (Non Case-sensitive)

### DIFF
--- a/Documentation/sp_Blitz Checks by Priority.md
+++ b/Documentation/sp_Blitz Checks by Priority.md
@@ -55,6 +55,7 @@ If you want to change anything about a check - the priority, finding, URL, or ID
 | 50 | Reliability | Remote Admin Connections Disabled | https://www.BrentOzar.com/go/dac | 100 |
 | 50 | Reliability | TempDB File Error | https://www.BrentOzar.com/go/tempdboops | 191 |
 | 50 | Reliability | Transaction Log Larger than Data File | https://www.BrentOzar.com/go/biglog | 75 |
+| 50 | Reliability | Default Trace File Error | https://BrentOzar.com/go/defaulttrace | 199 |
 | 100 | In-Memory OLTP (Hekaton) | Transaction Errors | https://www.BrentOzar.com/go/hekaton | 147 |
 | 100 | Features | Missing Features | https://www.BrentOzar.com/ | 189 |
 | 100 | Performance | Change Tracking Enabled | https://www.BrentOzar.com/go/tracking | 112 |
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -3443,8 +3443,8 @@ AS
 								LoginName ,
 								DBUserName
 							)
-							SELECT
-								t.TextData ,
+							SELECT TOP 20000
+								CONVERT(NVARCHAR(4000),t.TextData) ,
 								t.DatabaseName ,
 								t.EventClass ,
 								t.Severity ,
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -4447,6 +4447,7 @@ IF @ProductVersionMajor >= 10
 			IF NOT EXISTS ( SELECT  1
 								FROM    #SkipChecks
 								WHERE   DatabaseName IS NULL AND CheckID = 199 )
+							AND @TraceFileIssue = 0
 					BEGIN
 						  
 						  IF @Debug IN (1, 2) RAISERROR('Running CheckId [%d].', 0, 1, 199) WITH NOWAIT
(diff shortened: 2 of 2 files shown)


BlitzErik:
Hello there!

Cool, thanks. I like everything, but I'm a tiny bit concerned about what happens to that new temp table when there's a lot of data in the default trace. Can you give me a rough idea how many rows get inserted on servers you've tested on?

Is this a more sensible approach than just adding a safeguard check on the variable that holds the file name? CheckId 106 has this:

```
IF NOT EXISTS ( SELECT  1
FROM    #SkipChecks
WHERE   DatabaseName IS NULL AND CheckID = 106 )
AND (select convert(int,value_in_use) from sys.configurations where name = 'default trace enabled' ) = 1
AND DATALENGTH( COALESCE( @base_tracefilename, '' ) ) > DATALENGTH('.TRC')

```

RichBenner:
Hi Erik

I've tested on a few different databases (test and live) and the largest amount of data that got inserted into that temp table was 4900 rows. I thought that, considering we were pulling the same amount of data from sys.fn_trace_gettable (the same where clauses) throughout the code it would be similar.

Another check would be to do a select count(*) from sys.fn_trace_gettable which would also fail if there was a problem with the trace files and leave the old FROM clauses to use sys.fn_trace_gettable rather than putting this data into a temp table.

BlitzErik:
Just running this pulled out code:

Brings back just shy of 22k rows on My Computer® -- I can imagine this being much higher on a system that's been used and abused for a while. I do see your point about loading up a temp table once vs. going to the trace file multiple times though. 

So, up for discussion it is! @BrentOzar -- any religious beliefs on the subject?

```
DECLARE @StringToExecute NVARCHAR(4000),
        @curr_tracefilename NVARCHAR(500),
        @base_tracefilename NVARCHAR(500),
        @indx INT;

SELECT @curr_tracefilename = path
FROM   sys.traces
WHERE  is_default = 1;
SET @curr_tracefilename = REVERSE(@curr_tracefilename);
SELECT @indx = PATINDEX('%\%', @curr_tracefilename);
SET @curr_tracefilename = REVERSE(@curr_tracefilename);
SET @base_tracefilename = LEFT(@curr_tracefilename, LEN(@curr_tracefilename) - @indx) + '\log.trc';

SELECT COUNT(*)
FROM   sys.fn_trace_gettable(@base_tracefilename, DEFAULT) AS t
WHERE  (
           t.EventClass = 22
           AND t.Severity >= 17
           AND t.StartTime > DATEADD(dd, -30, GETDATE()))
       OR (
              t.EventClass IN ( 92, 93 )
              AND t.StartTime > DATEADD(dd, -30, GETDATE())
              AND t.Duration > 15000000 )
       OR ( t.EventClass IN ( 94, 95, 116 ));

```

BrentOzar:
Yeah, how about we stop at, say, 20k rows by default?

PR #1026, opened 2017-07-13
First stab at full backup failure detection

### LINKED ISSUE #1000: sp_AllNightLog doesn't retry the first full restore if it fails
Bug.

If sp_AllNightLog @Restore = 1 finds a new database, and tries to restore it from backup, and that restore fails because no file was found (like if there's a /databasename/full folder, but the backup file isn't there):

- No errors are thrown
- The magic time values in msdb.dbo.restore_worker are changed away from their 1900-01-01 00:00:00.000 and 9999-12-31 00:00:00.000 values, reflecting the time of the attempted restore
- That means the full restore is never tried again

It should retry the full restore on the next pass. (Not sure on implementation details yet regarding the error number column or start/finish time hacks.)

BlitzErik:
Closes #1000


BlitzErik:
@BrentOzar do we still need this PR?

BrentOzar:
Dang conflicts. Yeah, it'd be nice to have this long term, but if you want you can back-burner it. No need to jump into it now.

PR #1060, opened 2017-08-09
Fixed scoping of RAISERROR('Running CheckId [%d].') statements.

philipthatcher:
Fixed scoping of RAISERROR('Running CheckId [%d].') statements to indicate that the check was performed, not solely if the check produced a result.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -39,7 +39,7 @@ AS
 	IF @Help = 1 PRINT '
 	/*
 	sp_Blitz from http://FirstResponderKit.org
-	
+
 	This script checks the health of your SQL Server and gives you a prioritized
 	to-do list of the most urgent things you should consider fixing.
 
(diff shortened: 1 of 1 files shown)


BlitzErik:
HI Phillip.

Appreciate the effort, but a couple things:

1. We have a fix for this issue in #1055, which as been merged to dev.
2. Before adding code, you should open an issue to avoid this kind of double work. We would have closed it and let you know that we already have a fix in place.

Please read the [contributing](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/CONTRIBUTING.md) guidelines first.

Thanks,
Erik

philipthatcher:
My edits are an extension to #1055.

BlitzErik:
Okay, so take a step back, open an issue for your proposed changes, and read the contributing guidelines. 

Thanks!

PR #1080, opened 2017-08-24
sp_BlitzCache -- row estimate mismatch and forced serial issues

BlitzErik:
See commit notes


BlitzErik:
This also fixes the issues in BQS

PR #1156, opened 2017-10-12
sp_foreachdb - handle database names the same way as sp_MSforeachdb #536

MisterZeus:
Fixes issue #536

### DIFF
--- a/sp_foreachdb.sql
+++ b/sp_foreachdb.sql
@@ -169,6 +169,8 @@ AS
             BEGIN
                 SET @sql = REPLACE(@command, @replace_character, @db);
 
+                IF @suppress_quotename = 0 SET @sql = REPLACE(REPLACE(@sql,'[[','['),']]',']');
+
                 IF @print_command_only = 1
                     BEGIN
                         PRINT '/* For ' + @db + ': */' + CHAR(13) + CHAR(10)


BlitzErik:
Closes #536

amtwo:
@BlitzErik It looks like this is a breaking change. Is that intentional? I see the goal was to make sp_foreachDB a drop-in replacement for sp_MSforeachdb (which I commend!). But the *replacement* of @command with @command1 is going to break a bunch of stuff. 

What if sp_foreachdb maintained the existing @command for backwards compatibility, and also added the new "drop in params"?  Maintaining both for at least a short period would seem better than breaking code suddenly & without warning. 

This would require a few lines of code to check that you only supplied *either* @command1 or @command. Supplying both should throw an error.



```
 -- Original fields from sp_MSforeachdb...
    @command1 NVARCHAR(MAX) = NULL ,
    @replacechar NCHAR(1) = N'?' ,
    @command2 NVARCHAR(MAX) = NULL ,
    @command3 NVARCHAR(MAX) = NULL ,
    @precommand NVARCHAR(MAX) = NULL ,
    @postcommand NVARCHAR(MAX) = NULL ,
    -- Additional fields for our sp_foreachdb!
    @command NVARCHAR(MAX) = NULL
   <snip>
--error checking 
IF ( (@command1 IS NOT NULL AND @command IS NOT NULL)
      OR (@command1 IS NULL AND @command IS NULL) )
  BEGIN
    RAISERROR('You must supply either @command1 or @command, but not both.',16,1);
  END;
```

BlitzErik:
I'm totally cool with that. I hardly touch this one, so whatever makes sense for people who do is the best course.

Thanks!

PR #1157, opened 2017-10-13
#1132 sp_BlitzIndex - skip >100 partitions

### LINKED ISSUE #1132: sp_BlitzIndex - turn off @SkipPartitions unless @BringThePain = 1
Feature

@SkipPartitions defaults to 0. If there are more than, say, 100 partitions in the database, set @SkipPartitions = 1 unless @BringThePain is set to 1.

This will make our data collection run way faster by default (and we don't care about partition-level details initially.)

BrentOzar:
If there’s over 100 partitions and @BringThePain wasn’t set to 1, set
@SkipPartitions = 1. Closes #1132.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -985,6 +985,29 @@ BEGIN TRY
                 EXEC sp_executesql @dsql;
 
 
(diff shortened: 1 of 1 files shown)


BrentOzar:
@BlitzErik FYI, I merged this - getting a Collector build out for a client to run on the weekend.

PR #1161, opened 2017-10-13
#1160 sp_BlitzIndex skip non-readable AG secondary databases

### LINKED ISSUE #1160: sp_BlitzIndex @GetAllDatabases = 1 fails on AGs w/non-readable databases
Bug

On an AG secondary with databases set to be non-readable, run sp_BlitzIndex @GetAllDatabases = 1, and it fails on the databases it can't access.

All burritos should be available at breakfast time, not just the breakfast ones.

Never worked.

BrentOzar:
Closes #1160.


BrentOzar:
@BlitzErik FYI, I merged this - getting a Collector build out for a client to run on the weekend.

PR #1189, opened 2017-10-18
Issue 1180

### LINKED ISSUE #1180: sp_DatabaseRestore: Add option to force simple recovery model
Add option to force simple recovery model after restore.

If the use case is to restore from Live to non-live, typically we don't require full recover model in non-live

ShawnCrocker:
Fixes #1180 

Force simple recovery model after restore

How to test this code:
Restore a database that is FULL or BULKLOGGED
 - 

I need to push here and test elsewhere sorry. 
Don't have SQL Server installed here (yet)


ShawnCrocker:
Debugging comment wrong!

PR #1260, opened 2017-11-29
1255 Add context_info to output of sp_BlitzWho

SQLGumbo:
Enhancement to add the context_info field to the output
of sp_BlitzWho.  Field is obtained from the sys.dm_exec_requests
view.  CONVERT(VARCHAR(128), r.context_info)  AS [context_info]
was added to line 357.

Tested on SQL Server 2014 and 2016.

Fixes # .

 - 
 - 
 - 

How to test this code:
 - 
 - 
 - 

 - SQL Server 2008
 - SQL Server 2008 R2
 - SQL Server 2012
 - SQL Server 2014
 - SQL Server 2016
 - Amazon RDS
 - Azure SQL DB

### DIFF
--- a/sp_BlitzWho.sql
+++ b/sp_BlitzWho.sql
@@ -354,6 +354,7 @@ SELECT @StringToExecute = N'SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
 						END
 						+' 
 			            s.status ,
+                     CONVERT(VARCHAR(128), r.context_info)  AS [context_info],
 			            ISNULL(SUBSTRING(dest.text,
 			                             ( query_stats.statement_start_offset / 2 ) + 1,
 			                             ( ( CASE query_stats.statement_end_offset


SQLGumbo:
I could see this coming back.  Quite a few learning curves.  I tested the changes on 2014 and 2016.  Will keep an eye on this.  Thanks for giving me a chance to go through your process.   Also found out my AWS RDS instance was still up and running (no charges yet) when I went to test 2016 :-).

BlitzErik:
Thanks for putting this in! Look pretty good, but a few questions/comments:

1. You only put this in the 2012+ block -- is this version dependent? I checked the BoL page for exec_requests, and it didn't seem to be.

2. I'm not sure I'd put it that early in the select list, because this isn't something a lot of folks are gonna use. I'd probably put it towards the end with the resource governor info.

3. The column in the DMV is VARBINARY(128) -- what's the reason casting it to VARCHAR(128)?

SQLGumbo:
Thanks!  
0) Since I am going to be starting over with this...resubmitting.  Should I just remove the pull request?
1) Not sure about the 2012 block question.  Think I am missing something, but have an idea what it is and will check..   I did test on 2014 and 2016.  Searched BOL as well and didn't foresee any issues.
2) Roger that..with moving the field towards the Resource Governor.  Wondered about this.
3) I used VARCHAR 128 because it is my understanding that it is a VARBINARY 128. I cast it to make it readable.  Also want to bring up the [Phase out of CONTEXT_INFO](https://www.mssqltips.com/sqlservertip/4094/phase-out-contextinfo-in-sql-server-2016-with-sessioncontext/) which I do not believe will have an effect from everything I researched, but think this needs to be out there.
4) Also noticed after the fact that my formatting w/ tabs in SSMS is different than what was previously used.  Little OCD w/ code formats and caught that after the pull request was placed.

eaglejohn81:
Tested this on 2012 (11.0.5058.0):

image

works like a charm!

Regarding [session_context()](https://www.mssqltips.com/sqlservertip/4094/phase-out-contextinfo-in-sql-server-2016-with-sessioncontext/), we opened a MS case for this, but this is not shown in any DMV, so not useful here.

Thanks Jeff, Eric!

SQLGumbo:
Closing pull request.  Needs modifications.

PR #1319, opened 2017-12-05
#1318 - Make sp_foreachdb AG-aware

### LINKED ISSUE #1318: sp_foreachdb Breaks when an instance has AG non-readable secondaries
*Bug*
Version 1.9.5 - 20171115

If you run sp_foreachdb on an instance with AG secondary replicas that are *not* readable, it fails with 
Msg 976, Level 14, State 1, Line 35
The target database, 'IAmInAnAvailabilityGroup', is participating in an availability group and is currently not accessible for queries. Either data movement is suspended or the availability replica is not enabled for read access. To allow read-only access to this and other databases in the availability group, enable read access to one or more secondary availability replicas in the group.  For more information, see the ALTER AVAILABILITY GROUP statement in SQL Server Books Online.

The database state is "ONLINE" so you cannot use the existing filter to ignore non-readable secondary replicas. The pertinent information is not contained in 

Configure an AG where the secondary replicas do not accept connections, then run this query on a server with secondary replicas:
EXEC sp_foreachdb @suppress_quotename = 1, @state_desc = 'ONLINE', @command = 'USE [?]'

Using either existing or new parameters, I should be able to run sp_foreachdb on an instance with non-readable secondary replicas. I would be inclined to use the existing @state_desc = 'ONLINE' filter. Technically the DB's state is ONLINE, but it is effectively offline, since it is unavailable for querying. 

This fix would need to include version-aware logic, since the necessary AG DMOs are not present in all supported versions.

sp_foreachdb has never been AG-aware

amtwo:
Also sneak in fix for 1156 to maintain backwards compatibility of command params

Fixes #1318, #1156 

 - #1318 - When using existing `@state_desc` check for ONLINE database, ignore AG secondaries that are not accessible. (Instances without AGs enabled will no nothing)
 - #1156 - Maintain `@command` in addition to `@command1` for backwards compatibility. Do some checking to make sure exactly 1 value is supplied

How to test this code:
 - On a secondary AG replica that is *not* enabled for data access, run something like:
```
EXEC sp_foreachdb @suppress_quotename = 1, @state_desc = 'ONLINE', @command = 'USE [?]; SELECT db_name();'
EXEC sp_foreachdb @suppress_quotename = 1, @state_desc = 'ONLINE', @command1 = 'USE [?]; SELECT db_name();'
```
- Confirm that both return identical results, and does not error on the non-readable secondary replicas.
- Run something like this, and confirm that errors occur if both or neither `@command` and `@command1` are supplied:
```
EXEC sp_foreachdb @suppress_quotename = 1, @state_desc = 'ONLINE', @command1 = 'USE [?]; SELECT db_name();', @command = 'USE [?]; SELECT db_name();';
EXEC sp_foreachdb @suppress_quotename = 1, @state_desc = 'ONLINE';
```

 - SQL Server 2008
 - SQL Server 2008 R2
 - SQL Server 2012
 - SQL Server 2014
 - SQL Server 2016
 - SQL Server 2017

### DIFF
--- a/sp_foreachdb.sql
+++ b/sp_foreachdb.sql
@@ -7,13 +7,14 @@ GO
 
 ALTER PROCEDURE dbo.sp_foreachdb
     -- Original fields from sp_MSforeachdb...
-    @command1 NVARCHAR(MAX) ,
+    @command1 NVARCHAR(MAX) = NULL,
     @replacechar NCHAR(1) = N'?' ,
     @command2 NVARCHAR(MAX) = NULL ,
     @command3 NVARCHAR(MAX) = NULL ,
     @precommand NVARCHAR(MAX) = NULL ,
     @postcommand NVARCHAR(MAX) = NULL ,
     -- Additional fields for our sp_foreachdb!
+    @command NVARCHAR(MAX) = NULL,  --For backwards compatibility
     @print_dbname BIT = 0 ,
     @print_command_only BIT = 0 ,
     @suppress_quotename BIT = 0 ,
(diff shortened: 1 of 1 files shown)


BlitzErik:
Andy sez: This is the wrong code. Submitting a new request with the right code.

PR #1345, opened 2017-12-17
Fixes #1344

### LINKED ISSUE #1344: sp_BlitzIndex: Check NC indexes for bad read:write ratios
We flag indexes for lots of stuff, including not being used. It would be helpful to flag indexes with poor read to write ratios as well. I think 10:1 is a good starting point.

Things to keep in mind:
- Index created recently
- Index rebuilt recently
- Index is unique, on a FK, etc.
- Index is a PK/CX

It would be helpful in analysis of dupe/borderline dupe indexes, so maybe...

Have an is_low_use column that gets updated or is computed, and tack a case expression onto the dupe/borderline dupe check to add the wording in. It wouldn't have to be a separate check at all in this case.

BlitzErik:
Adds check for indexes with bad read:write ratio

Also using this PR to make sure everything has an approproate semi-colon. It's not fun to write code that errors our because one is missing.

Fixes #1344 

 - Adds check for NC indexes with 10x write:read ratio
 - Properly terminates all statements with a semicolon

How to test this code:
 - Run against a database that has indexes which are heavily written to, but not heavily reads from:

image

 - SQL Server 2016

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -9,7 +9,7 @@ SET STATISTICS TIME OFF;
 GO
 
 IF OBJECT_ID('dbo.sp_BlitzIndex') IS NULL
-  EXEC ('CREATE PROCEDURE dbo.sp_BlitzIndex AS RETURN 0;')
+  EXEC ('CREATE PROCEDURE dbo.sp_BlitzIndex AS RETURN 0;');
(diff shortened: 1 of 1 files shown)


BlitzErik:
I also updated the URL associated with this warning to give more info on it.

Issue #1349, opened 2017-12-19
sp_BlitzCache: some warnings aren't firing consistently


BrentOzar:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Some of the warnings (not sure how many yet) are not firing consistently. Here's how to reproduce it:

```
DBCC FREEPROCCACHE
GO
/* Build a batch that has both regular tables and table variables */
DROP TABLE IF EXISTS ##bou_BlitzCacheProcs;
GO
DBCC FREEPROCCACHE
GO
DECLARE @Crap TABLE
	    (
	        Id INT NOT NULL
	    );
	
INSERT @Crap (Id )
SELECT database_id
FROM   sys.databases;
	
SELECT *
FROM   @Crap AS c;
GO
sp_BlitzCache
GO
```

Sometimes you'll get the table variable warning, and sometimes you don't.

The root cause is in the repro.txt script attached. The last statement is the table variables check, but does a select instead of an update so you can see the data.
[repro.txt](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/files/1573384/repro.txt)

The last result set shows the problem:
![1](https://user-images.githubusercontent.com/245462/34180638-1c854872-e4d5-11e7-822a-58cede5a7acc.png)

Some of the objects start with @, and some don't. As a result, it's kinda wild luck as to whether or not the is_table_variable field is set to 1 or 0.

I think the easiest fix is to add a where clause to the update so that it only fires when x.first_char = '@', and ignores the rest of the rows. That one is easy - then I also have to go through the rest of the checks looking for this same logic gotcha.

BrentOzar:
There's one query I didn't check because I don't have a quick repro of it - look for this line in the code:

/* BGO #1349 2017/12/19 I think the below won't always reliably fire if there are multiple scans in a plan, and some are forced and some aren't */

The rest are much better, though.

PR #1351, opened 2017-12-19
#1349 sp_BlitzCache make warnings fire more consistently

BrentOzar:
Calling it quits on this for now, but may need to leave that issue open
if Erik wants to take another look.


BlitzErik:
I think I have to revert this. The change for table variables is right, but the TOP 1s look like they're just grabbing 1 row, not 1 row per sqlhandle/queryhash.

BrentOzar:
OH NOES

Issue #1322, opened 2017-12-06
sp_BlitzCache - String or binary data would be truncated


rossw99:
Running 6.0 and hit a problem.  Running on SQL Server 13.0.4435.

I receive the error:
Inserting variables
Msg 8152, Level 16, State 13, Procedure sp_BlitzCache, Line 2980 [Batch Start Line 0]
String or binary data would be truncated.
The statement has been terminated.
Updating procs

I tracked it down to this:
INSERT #stored_proc_info ( SPID,    SqlHandle,    QueryHash,    variable_name,    variable_datatype,    compile_time_value,    proc_name )
SELECT                     vi.SPID, vi.SqlHandle, vi.QueryHash, vi.variable_name, vi.variable_datatype, vi.compile_time_value, vi.proc_name
FROM #variable_info AS vi

I believe the problem is related to the mismatched data types in the #stored_proc_info and #variable_info tables for:
compile_time_value NVARCHAR(128),
proc_name NVARCHAR(300),
variable_name NVARCHAR(128),

vs.

compile_time_value NVARCHAR(4000)
proc_name NVARCHAR(128),
variable_name NVARCHAR(200),

BlitzErik:
Great catch! I'll kick out a fix for you in a minute.

rossw99:
FYI, this is what I was running.

exec dba.dbo.sp_BlitzCache @Top = 20, @SortOrder='reads', @OutputDatabaseName  = 'DBA', @OutputSchemaName = 'dbo', @OutputTableName  = 'sp_BlitzCache_ReadsResults'

BlitzErik:
Cool, thanks. Just to make sure, does it only error when you output to a table?

rossw99:
Just tried without output to table and produced same error.

BlitzErik:
Can you grab the version of BlitzCache [here](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/Issue_1322/sp_BlitzCache.sql) and let me know if it produces an error?

rossw99:
Confirmed working with and without output to table option!

Thanks for the super fast response ;-)

BlitzErik:
Great! Thanks for letting me know, and sorry about the troubles.

rossw99:
All good.  Discovered and fixed in under an hour!

HenryLee19:
[...]

PR #1353, opened 2017-12-21
spBlitzCache Error Updating Procs

johnhoven:
8152: String or binary data would be truncated.

Appears to be issue with CharIndex called with varchar and nvarchar columns.
https://connect.microsoft.com/SQLServer/feedback/details/537610/charindex-patindex-does-not-handle-nvarchar-max-or-varchar-max-correctly-generates-error-msg-8152-when-used-on-expression-1

If I re-load the dev branch of sp_BlitzCache the error comes back w/ my current cache.  It appears there is no data in #stored_proc_info when the procedure fails.

 - Use same data type for finding CharIndex of s.variable_datatype

How to test this code:
 - I don't know exactly how to reproduce, just that it is happening in my environment at times, probably based on the stored procedures in cache.

 - SQL Server 2014

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -3206,7 +3206,7 @@ OPTION ( RECOMPILE );
 RAISERROR(N'Updating procs', 0, 1) WITH NOWAIT;
 UPDATE s
 SET    s.variable_datatype = CASE WHEN s.variable_datatype LIKE '%(%)%' THEN
-                                      LEFT(s.variable_datatype, CHARINDEX('(', s.variable_datatype) - 1)
+                                      LEFT(s.variable_datatype, CHARINDEX(N'(', s.variable_datatype) - 1)
 								  ELSE s.variable_datatype
                              END,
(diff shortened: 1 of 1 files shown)


johnhoven:
Disregard the Connect comment, that was regarding NVARCHAR(MAX) and this column is NVARCHAR(128).  I'm not sure what the issue is or why this fixes it for me yet.

BlitzErik:
Hi John,

I appreciate the attention, but a few things:

1. Are you using the newest (or the version in the Dev branch) of sp_BlitzCache?
2. Before submitting pull requests, please search for issues, and only open one if [you can't find one](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/1322) that addressed your problem.
3. When submitting issues, please let us know which version or version date of the proc you're using.

We work really hard to document the best way to [contribute code](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/CONTRIBUTING.md), so that the process is transparent and trackable.

Thanks!

PR #1361, opened 2017-12-29
Fixes checks for active and inactive heaps

russellhart:
Fixes checks for active and inactive heaps, checks 38, 39.

The checks on the user columns of sys.dm_db_index_usage_stats checks for nulls however a row can exist due to a system scan. In this case the user columns are defaulted to zeros. This change correctly handles zeros. 

 - SQL Server 2008 R2
 - SQL Server 2012
 - SQL Server 2016


BlitzErik:
I appreciate the attention, but, we work really hard to document the best way to [contribute code,](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/CONTRIBUTING.md) so that the process is transparent and trackable.

Thanks!

PR #1394, opened 2018-01-19
Issue 1393 corrected name of default XE session

### LINKED ISSUE #1393: sp_Blitz incorrectly reporting the default XE diagnostics session as an extra one
sp_Blitz Version 6.1 (20180101 Date) lines 4239-4260 check for extended event sessions that are not standard and are running.

`SELECT name
FROM sys.dm_xe_sessions;`
returns several rows that include

- system_health
- AlwaysOn_health
- sp_server_diagnostics session
- hkenginexesession
- telemetry_xevents

Line 4258 excludes names that are not in ( 'AlwaysOn_health', 'system_health', 'telemetry_xevents', '**sp_server_diagnostics**', 'hkenginexesession' )

If I have 1 additional extended event session then I will trip CheckId 176.

I would expect that **sp_server_diagnostics session** would be excluded from the check as well.

I've tested this on SQL Server 2016 SP1 CU2 and CU7 as well as SQL Server 2008R2 SP3. It looks like this was broken by @BlitzErik commit 1c7c8ee4443dfdaac5bf4ca8cb6a3d52c38cd324 (which was supposed to fix issue #1116 

In the process of adding the exclusion for AlwaysOn_health session, the name of the sp_server_diagnostics session event was altered (because it's also the only extended event session that has a space followed by the word session).

elsimer:
Fixes #1393  .

 - correct name of default XE session

How to test this code:
 - start one XE session
 - verify single non-default XE session is running by executing` SELECT name FROM sys.dm_xe_sessions`
 - run sp_Blitz
 - look for CheckID 176 (it should not be there)
 - start one additional XE session
 - verify there are two XE sessions running
 - run sp_Blitz
 - verify that CheckId 176 is present with Priority 200

 - SQL Server 2008 R2
 - SQL Server 2016

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -4251,7 +4251,7 @@ IF @ProductVersionMajor >= 10
 													'Hey big spender, you have ' + CAST(COUNT_BIG(*) AS VARCHAR(30)) + ' Extended Events sessions running. You sure you meant to do that?' AS Details
 											    FROM sys.dm_xe_sessions
 												WHERE [name] NOT IN
-												( 'AlwaysOn_health', 'system_health', 'telemetry_xevents', 'sp_server_diagnostics', 'hkenginexesession' )
+												( 'AlwaysOn_health', 'system_health', 'telemetry_xevents', 'sp_server_diagnostics sessions', 'hkenginexesession' )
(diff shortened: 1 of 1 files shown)


BlitzErik:
Looks great, thanks!

PR #1408, opened 2018-02-05
sp_Blitz: Remove trailing spaces and surplus CR/LFs (newlines)

MisterZeus:
Thought I'd attempt a proto-cleanup of the sp_Blitz.
I know that whitespace styles are hotly debated and up to each person's preference, but guessed I could get rid of the worst offenders that most people tend to agree on: trailing spaces and extra newlines.
Let's see how this goes down... :-)

 - Remove trailing spaces
 - find places with 3 or more CR/LFs in a row and cut them down to 2

How to test this code:
 - run it to replace normal sp_Blitz
 - carry out as many different runs of sp_Blitz as human-machinely possible

 - SQL Server 2016

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -63,7 +63,6 @@ AS
      Changes - for the full list of improvements and fixes in this version, see:
      https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/
 
-
 	Parameter explanations:
 
 	@CheckUserDatabaseObjects	1=review user databases for triggers, heaps, etc. Takes more time for more databases and objects.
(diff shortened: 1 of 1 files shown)


BrentOzar:
Howdy sir. Thanks for the contribution.

Here's the thing: manually reviewing this is a little bit of a pain, going through every line of changes to make sure it's only spacing issues. I appreciate the enthusiasm for perfect code, but keep in mind that it's volunteers on the other side of the screen, too. I have to be careful about what work we take on, and make sure that it's providing big value in exchange for our time.

So with that in mind, going forward, can you follow the contribution guide? Before you make a change, file an issue with the changes you're considering, talk about it with the other contributors in Slack, and make sure that it's a good fit. (This would probably be an example of something we wouldn't recommend doing just because we only have so many hours in the day to review code.)

I'll put in the volunteer time to review this one just this one time, but going forward, I need you to follow the contribution guide process.

Thanks!

MisterZeus:
Brent, that sounds fair enough. I appreciate the one-off review and merge, and I'll go have a read of the contribution guidelines, since I'm hoping to contribute more stuff in the future!

Cheers!

Issue #1407, opened 2018-02-04
Sp_blitz New feature request to check if the owner of a database is an existing user


parlevjo2:
If I execute sp_helpdb i see many times databases which have no current database owner.
Mostly because the owner of the database was an administrator which has left the company, eg:

name | db_size | owner
--------- | --------- | ---------
dbname | 2284.13 MB | \~~ UNKNOWN ~~

Those can easily be identified:
`SELECT * FROM sys.databases WHERE SUSER_SNAME(owner_sid) IS NULL`

It would be nice if sp_blitz could report these databases.
Note: They are not reported by CheckID 55 'Database Owner <> SA'

I already have written the code. It is almost the same as CheckID 55.
Part of the code which can also be easily executed on its own:
```sql
SELECT  NN AS CheckID ,
        [name] AS DatabaseName ,
        230 AS Priority ,
        'Security' AS FindingsGroup ,
        'Database Owner Unknown' AS Finding ,
        'https://BrentOzar.com/go/owndbunknown' AS URL ,
        ( 'Database name: ' + [name] + '   '
        + 'Owner name: ' + ISNULL(SUSER_SNAME(owner_sid),'') ) AS Details
FROM    sys.databases
WHERE   SUSER_SNAME(owner_sid) is NULL
```

[Link to SUSER_SNAME()](https://docs.microsoft.com/en-us/sql/t-sql/functions/suser-sname-transact-sql)
This is my first time on Github. I already have forked and cloned the project.

parlevjo2:
Hi @BlitzErik ,

How does it works with new CheckID's.
Can I use CheckID 58 for my code ?

BlitzErik:
@parlevjo2 - Please [look here](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/Documentation/sp_Blitz%20Checks%20by%20Priority.md)

parlevjo2:
I would like to make a new CheckID with these characteristics. Open for a debate:

Priority | FindingsGroup | Finding | URL | CheckID
-------- | ------------- | ------- | --------------------------------------- | ----------------
230 | Security | Database Owner is Unknown | https://BrentOzar.com/go/owndbunknown | Current High CheckID + 1

It is based on CheckID 55 which is in the same category.

Priority | FindingsGroup | Finding | URL | CheckID
-------- | ------------- | ------- | --------------------------------------- | ----------------
[...]

PR #1423, opened 2018-02-12
Issue 1407/parlevjo2

parlevjo2:
Enhancement #1407.

 - Check if the owner of a database is an existing user
Sometimes if you execute sp_helpdb you see databases which have no current database owner.
Mostly because the owner of the database was an administrator which has left the company, eg:

name | db_size | owner
--------- | --------- | ---------
dbname | 2284.13 MB | \~~ UNKNOWN ~~

This check reports those databases so you can give the database a proper owner like the 'sa' user.

How to test this code:
 - Create a new Windows account. 
 - Create a new SQL account for this Windows account. 
 - Create a test database with this SQL account as owner of the test database.
 - Delete the new SQL account.
 - Delete the new Windows account.
 - Execute sp_helpdb to see that the owner is Unknown.
 - Execute sp_blitz on this sql instance.

 - SQL Server 2008 R2
 - SQL Server 2012


BlitzErik:
Oh boy, something very strange happened. Are you working with a recent sync of the dev branch?

For starters, more than just sp_Blitz was modified:

image

And the changes to Blitz don't match what you were working on:

image

PR #1431, opened 2018-02-15
Issue 1181 single kill drop

### LINKED ISSUE #1181: sp_DatabaseRestore: add options to kill user connections first and/or drop database
Developers won't close SSMS so we need to force majeure this.
To stop reconnection, DROP may be needed

```
    SET @DynamicCommand = '/* Kill connections */';
    SELECT 
        @DynamicCommand = CONCAT(@DynamicCommand, '
    KILL ', spid, ';')
    FROM
        master.sys.sysprocesses  
    WHERE
        dbid = DB_ID(@DBName)
    EXEC (@DynamicCommand);
```

ShawnCrocker:
Fixes #1181

The database to be restored may already exist, and be in-use.

In my personal use case I have spammy 3rd party apps that run as db_owner. 
The safest way (for me) to ensure a restore is to drop, but other use-cases exist: single user and kill connections

This adds an option via paramater @ExistingDBAction to 
- 0 do nothing
- 1 set single user
- 2 kill connections
- 3 drop database (implies kill connections)

How to test this code:
- Try each flag

SQL Server 2016


BlitzErik:
Howdy sir! It looks like you're working with a pretty old dev base. Do you mind updating and making another PR? I'm not nearly good enough at GitHub to sort out all these changes.

image

PR #1470, opened 2018-03-06
Fix sp_BlitzLock reports only 1 total deadlock for each object

### LINKED ISSUE #1469: sp_BlitzLock reports only 1 total deadlock for each object
spBlitzLock version 'Mar 1 2018 12:00AM'

Check_id 2 reports "This object was involved in 1 deadlock(s)." for each object detected, regardless of the actual number of deadlocks.

Here's a sqlfiddle that recreates the problem with the query (which seemed simpler than creating actual deadlocks):
http://sqlfiddle.com/#!18/e26d5/1

It should report the number of unique deadlock events that involved each objects.

This is the first time I've used sp_BlitzLock. I noticed this on SQL 2016 / Windows 2012 R2.

HolisticDeveloper:
Fixes #1469.

 - Count the number of distinct `event_date` values for each object, rather than the name (which will always be 1?).

How to test this code:
 - See this sqlfiddle: http://sqlfiddle.com/#!18/e26d5/1

 - SQL Server 2016

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -373,7 +373,7 @@ SET @VersionDate = '20180301';
 			   dow.object_name AS object_name,
 			   'Total object deadlocks' AS finding_group,
 			   'This object was involved in ' 
-				+ CONVERT(NVARCHAR(20), COUNT_BIG(DISTINCT dow.object_name))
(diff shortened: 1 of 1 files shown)


HolisticDeveloper:
This is my first PR, so let me know if I need to do anything differently.

PR #1476, opened 2018-03-15
#1474 sp_Blitz odd cores check

### LINKED ISSUE #1474: sp_blitz should ignore soft-NUMA for odd cores check
I am reporting what I believe to be a bug in version 6.3 of sp_blitz.

The check for odd number of cores in a NUMA node doesn't account for soft-NUMA. The comment for the section in question is:
/*Perf - Odd number of cores in a socket*/

Install SQL Server 2016 or later on a bare metal machine with 10 or 14 physical cores per socket with hyperthreading disabled or on a VM with access to 10 or 14 logical CPUs per memory node. Auto soft-NUMA will create NUMA nodes with odd numbers of schedulers and you'll see an error about odd numbers of schedulers per NUMA node.

Here's an example:

https://dba.stackexchange.com/questions/196327/automatic-soft-numa-results-in-odd-number-of-cores/

Based on the code and comments, the intention is to look at the number of schedulers per memory node instead of per soft-NUMA node. A query like the below should work, but I'm unable to test it at the moment:

SELECT  1
FROM    sys.dm_os_schedulers dos
INNER JOIN sys.dm_os_nodes don ON dos.parent_node_id = don.node_id
WHERE   dos.is_online = 1
AND dos.scheduler_id < 255
AND don.memory_node_id < 64
GROUP BY don.memory_node_id,
dos.is_online
HAVING  ( COUNT(dos.cpu_id) + 2 ) % 2 = 1;

This mostly affects SQL Server 2016 and 2017 with the release of auto soft-NUMA. There could have problems in previous versions with manual soft-NUMA.

BrentOzar:
Now look for uneven numbers of cores across nodes. Closes #1474.

Closes #1474.

- Instead of looking for odd numbers of cores, look for NUMA nodes with different numbers of cores, like where one node has 4 and another has 5, possibly leading to load imbalance issues (but also suggesting some herp-a-derp on the admin's part.)

How to test this code:
 - On a normally configured server or VM, the priority 10 "Uneven Core Distribution Across NUMA Nodes" warning shouldn't show up.
 - Create a VM with a low but odd number of cores, like 3 or 5, and it still shouldn't show up because you'd only have one NUMA node. (Even if you do something dorky like set it up as a single-core-per-CPU VM, then you'd end up with an odd number of NUMA nodes, but I'm not even going near that.)
 - Create a VM with a high and odd number of cores and multiple CPU sockets, like 19 or 31 or 1,973, and the priority 10 "Uneven Core Distribution Across NUMA Nodes" will show up.

 - SQL Server 2017

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -4454,19 +4454,19 @@ IF @ProductVersionMajor >= 10
 						'Mismatch between the number of TempDB files in sys.master_files versus tempdb.sys.database_files' AS [Details];
 				END;
 
-/*Perf - Odd number of cores in a socket*/
+/*Perf - Uneven core distribution across sockets */
 		IF NOT EXISTS ( SELECT  1
 		                FROM    #SkipChecks
 		                WHERE   DatabaseName IS NULL
 		                        AND CheckID = 198 )
-		   AND EXISTS ( SELECT  1
-		                FROM    sys.dm_os_schedulers
-		                WHERE   is_online = 1
-		                        AND scheduler_id < 255
-		                        AND parent_node_id < 64
-		                GROUP BY parent_node_id,
-		                        is_online
-		                HAVING  ( COUNT(cpu_id) + 2 ) % 2 = 1 )
+		   AND EXISTS ( SELECT MIN(schedulers) as bottom, MAX(schedulers) AS topper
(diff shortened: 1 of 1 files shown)


BlitzErik:
Closing this so Joe can work on it.

PR #1529, opened 2018-04-16
BlitzCache & Blitz - Version checks - 2017 but not yet CU3

### LINKED ISSUE #1528: sp_BlitzCache - throw more elegant error on SQL Version 2017 that isn't patched to CU3
I guess it's a bug, but it's definitely not a surprise

sp_BlitzCache chucks out syntax errors if people are on SQL 2017 but haven't patched up to CU3 because MS added cool new stuff into some dmv's that sp_BlitzCache takes advantage of.

With the amount of times this has been asked in the slack channel it's probably worth adding in a check for their version and throw up some sort of alert to say that they need to patch.

SQL Server 2017 before CU3

Discussed with Brent on slack, I'm happy to code this one as soon as i can find an early version of 2017

RichBenner:
Fixes #1528  .

 - Update BlitzCache so that we don't go looking for columns in DMV's that don't exist
 - Update Blitz to report if this is the case and nudge that they should patch
 - Update Blitz documentation

How to test this code:
 - Find a 2017 server that's not yet on CU3 
 - Run BlitzCache, should now not throw errors
 - Run Blitz, you should see a 100 error showing that user should patch.

 - SQL Server 2017 (14.0.1000.169)

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -3446,13 +3446,13 @@ AS
                         /* Check if SQL 2017 but not CU3 */
 						IF NOT EXISTS ( SELECT  1
 										FROM    #SkipChecks
-										WHERE   DatabaseName IS NULL AND CheckID = 189 )
+										WHERE   DatabaseName IS NULL AND CheckID = 216 )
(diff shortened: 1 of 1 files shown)


BlitzErik:
Looks great! Thanks @RichBenner !

PR #1572, opened 2018-05-11
Add priority to the output of sp_BlitzLock

### LINKED ISSUE #1568: BlitzLock - Add "Priority" to the output
Request a feature.

The first resultset of sp_BlitzLock does not include the deadlock priority of the processes / queries involved in the deadlock.

It would be nifty if the resultset *did* include the priority, since that's one of the main factors in victim choice.  This one (not "taskpriority" - I have no idea what that is):

    <deadlock>
      ...
      <process-list>
        <process ... priority="0" ... >

jadarnel27:
Put it right next to log_used, since these go together to determine victim choice.

Fixes #1568.

 - Add priority to the output of the first (detailed) resultset

How to test this code:
 - Run `exec sp_BlitzLock` on a machine that has deadlocks to be analyzed
 - Verify the "priority" column is there between 

 - ~~Case-sensitive SQL Server instance~~
 - SQL Server 2012
 - ~~SQL Server 2014~~
 - SQL Server 2016
 - ~~SQL Server 2017~~
 - ~~Amazon RDS~~
 - ~~Azure SQL DB~~

Note about testing: since this was just accessing one more property in the same XML document being read throughout the script, I think the risk is low that I introduced version-specific errors, so I only tested on versions that I have installed already


BlitzErik:
That's totally cool. This script only runs on 2012+ anyway, so the surface area is pretty small.

PR #1621, opened 2018-06-08
Get Power Plan value from proper location in registry

### LINKED ISSUE #1620: sp_Blitz - Power Plan is not retrieved correctly if set by Group Policy in Windows
sp_Blitz 6.5

Results incorrectly say that the server is on a power plan that it is not necessarily on.

Script needs to check a different spot in the registry to get the value set by group policy. If this key does not exist, it should then fall back to the local registry key.

AAvella:
If Power Plan was set by group policy, the value is in a different registry key. Look for this one first.

Fixes #1620 .

 - Grabs value from group policy registry location if it exists. Otherwise, grabs local.

GROUP POLICY LOCATION:
HKLM\SOFTWARE\Policies\Microsoft\Power\PowerSettings\ActivePowerScheme

LOCAL LOCATION:
HKLM\SYSTEM\CurrentControlSet\Control\Power\User\PowerSchemes\ActivePowerScheme

How to test this code:
 - Run against an instance that has a power plan of "high performance" set by GPO (Windows Group Policy), but "balanced" or "power saver" set locally. The fixed version will not tell you that you have a 'power saver' or 'balanced' plan when you have "high performance" set by GPO. 

SQL 2008 - SQL 2017

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -7324,12 +7324,19 @@ IF @ProductVersionMajor >= 10 AND  NOT EXISTS ( SELECT  1
 								
 								IF @Debug IN (1, 2) RAISERROR('Running CheckId [%d].', 0, 1, 211) WITH NOWAIT;
 
-								DECLARE @outval VARCHAR(36);							
+								DECLARE @outval VARCHAR(36);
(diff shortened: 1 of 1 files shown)


BlitzErik:
This looks great, thank you!

PR #1636, opened 2018-06-17
Add userstore token

### LINKED ISSUE #1633: sp_Blitz and sp_BlitzFirst: warn about userstore_tokenperm memory size
Feature.

We've seen situations where the plan cache is just continuously empty even though there's plenty of memory available. This feedback item suggests that it might be related to userstore_tokenperm size grows to 5GB, but I'm not going to be able to build a repro quickly:

https://feedback.azure.com/forums/908035-sql-server/suggestions/32903272-procedure-cache-empties-spontaneously-even-on-an-i

If we fire the alert for not much plan data in the cache, let's also include a check of the userstore_tokenperm cache size, and suggest clearing that to see if it alleviates the issue.

BlitzErik:
Fixes #1633 

 - Sum up amount of USERSTORE_TOKENPERM entries in dm_os_memory_cache_entries. If it's > 2 GB, warn about it with recent plan cache clearing.

How to test this code:
 - I have no idea how to get that cache entry to be really high, so I'm flying a little blind on this one.

 - SQL Server 2016
  - SQL Server 2017

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -3300,6 +3300,42 @@ AS
 							
 							IF @Debug IN (1, 2) RAISERROR('Running CheckId [%d].', 0, 1, 125) WITH NOWAIT;
 							
+								DECLARE @user_perm_sql NVARCHAR(MAX) = N'';
+								DECLARE @user_perm_gb_out DECIMAL(38,2);
+								
+								IF @ProductVersionMajor >= 11
+								
+								BEGIN
+								
+								SET @user_perm_sql += N'
+									SELECT @user_perm_gb = CASE WHEN SUM(pages_kb) >= (1048576 * 2)
+											THEN CONVERT(DECIMAL(38, 2), SUM(pages_kb))
+											ELSE NULL 
+										   END
+									FROM sys.dm_os_memory_cache_entries
+									WHERE type = ''USERSTORE_TOKENPERM''
+								';
+								
+								END
+								
+								IF @ProductVersionMajor < 11
+								
+								BEGIN
+								SET @user_perm_sql += N'
+									SELECT CASE WHEN SUM(pages_allocated_count) / 128.0 / 1024. >= (1048576 * 2)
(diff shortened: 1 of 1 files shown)


jadarnel27:
I think this check should use `sys.dm_os_memory_clerks` instead of `sys.dm_os_memory_cache_entries`.

I found a functional repro for bloating that particular cache entry here: [USERSTORE_TOKENPERM is Huge](https://www.sqlservercentral.com/Forums/Topic1489818-2799-1.aspx)

Here's the code:

```
CREATE APPLICATION ROLE [AppRole] WITH DEFAULT_SCHEMA = [dbo], PASSWORD = N'password'

declare @counter int = 1

While(@counter < 20000)
Begin

DECLARE @cookie varbinary(8000); 

EXEC sp_setapprole @rolename = 'AppRole', @password = 'password', @fCreateCookie = true, @cookie = @cookie OUTPUT;

select user_name();

EXEC sp_unsetapprole @cookie;

--GO

SELECT USER_NAME();

SET @counter = @counter + 1

end
```

Tested on SQL Server 2012.  Running the script above a few times got this cache entry up to 2GB (then it mysteriously stopped growing).

When I run the query below, I get lots of rows, but all of the "pages_kb" values are zero:

    select pages_kb from sys.dm_os_memory_cache_entries where [type] = 'USERSTORE_TOKENPERM';

This one, however, keeps going up and up the more I run that looping code from above:

    select pages_kb from sys.dm_os_memory_clerks where [type] = 'USERSTORE_TOKENPERM' and [name] = 'TokenAndPermUserStore';

The check should probably use *that* pages_kb instead.  

As far as the plan cache clearing, I'm not positive how to test to see if that's happening.  The count in `select * from sys.dm_exec_cached_plans` is remaining steady, but I don't have very many rows (200) in my local SQL Server 2012 instance, and there's not much going on with it other than this test 😄

BlitzErik:
Holy cow, awesome. Where'd you even find that? We both looked around for a bit.

jadarnel27:
Haha, I just searched "USERSTORE_TOKENPERM" by itself and the Google gods found favor with me I guess.

PR #1659, opened 2018-06-26
Issue 1610

### LINKED ISSUE #1610: Add sp_rename code for DTA indexes in SpBlitzIndex
One thing that might be nice is if you added the sp_rename code for indexes that might be dups for the _dta_% indexes. These may be good indexes, but they are poorly named. You could start them as IX_col_col_col names, with the user then overriding the name if needed.

Dropping isn't great to provide the code, but the rename (as in lab 1 of the Index Tuning class) is nice.

If you want to, maybe add this for all system generated names, such as poor PKs.

BlitzErik:
Fixes #1610 

 - Script renaming of DTA indexes rather than create scripts

How to test this code:
 - Run on a database with DTA indexes in them

 - SQL Server 2016
  - SQL Server 2017


BrentOzar:
Trying to review - where am I supposed to see it? (I think I'm missing something obvious.)

BlitzErik:
I had just created an index like this to test: `CREATE INDEX _dta_index_Users ON dbo.Users (Age);`

It's close enough in proximity to what I see in our data for the naming convention used. It worked fine for that.

PR #1696, opened 2018-07-26
Fix Arithmetic Overflow from old system view

josh-simar:
Changes code to not use old system view that is erroring when user has an id over smallint.

Fixes # 1695

 -  stop using sysmembers system view in check 86

How to test this code:
 - Have a user with a id over 32,767
 - run sp_Blitz

 - SQL Server 2016


josh-simar:
Closing as I threw this in too fast without enough testing.

MisterZeus:
@josh-simar Could break out that Super Long Line into multiple lines, for readability?

Issue #1679, opened 2018-07-11
Procedure parameter OutputType - "NONE" without output


jeffchulg:
**Do you want to request a *feature* or report a *bug*?**
If you're reporting a bug, include the version number of the script you're using. If it's not the current version, upgrade to the current version and test that before reporting a bug - we fix a lot of stuff in each new build.

feature

**What is the current behavior?**

sp_Blitz has the ability to set a value of `NONE` to its parameter `@OutputType`. This allows a DBA to setup a collection without returning any dataset. It's not the case for instance for `sp_BlitzFirst`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via http://sqlfiddle.com**

**What is the expected behavior?**

My request is to generalize this behavior to all procedures with an `@OutputType` parameter.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**

BrentOzar:
Jeff - did you want to work on this, or do you want us to close it?

jeffchulg:
Hi Brent,

I will take care of the development of this as soon as possible.

PR #1699, opened 2018-08-08
Add sp_Blitz's parameter @OutputType to sp_BlitzIndex and sp_BlitzFirst

jeffchulg:
Fixes #  #1679

 - some tab to space adjustments
 - some reformatting
 - add a parameter @Output parameter or the handling of a "NONE" value for this parameter

How to test this code:
 - same as usual should give the same result as usual
- set @OutputType to NONE alone will make it fail
 - set @OutputType to NONE with @OutputTableName to something will store the results to the output table 

 - SQL Server 2014

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -46,7 +46,7 @@ SET @VersionDate = '20180701';
 
 IF @Help = 1 PRINT '
 sp_BlitzFirst from http://FirstResponderKit.org
-	
+    
 This script gives you a prioritized list of why your SQL Server is slow right now.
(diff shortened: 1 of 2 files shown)


BrentOzar:
Jeff - thanks for the pull request, but let's avoid reformatting requests. Makes it incredibly difficult to see what you've changed - try looking at the diff:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/pull/1699/files

I'm going to close this, but feel free to reopen another one with only the logic changes that you need. Sorry about that!

PR #1702, opened 2018-08-10
#1697 sp_BlitzIndex skip partitions checking

### LINKED ISSUE #1697: sp_BlitzIndex - partitions check
Seems like the script checks for total partition count even when it is being executed for a single table.
Does it seem reasonable to add filter by current table in this check?
Here is an example when script fails for non-partitioned table.
`EXEC master.dbo.sp_BlitzIndex @DatabaseName='bizproj', @SchemaName='dbo', @TableName='userLastEvent'`
```
Starting run. sp_BlitzIndex(TM) v6.7 - July 01, 2018
Create temp tables.
Adding UQ index on #IndexSanity (database_id, object_id, index_id)


bizproj
Inserting data into #IndexColumns for clustered indexes and heaps
Inserting data into #IndexColumns for nonclustered indexes
Inserting data into #IndexSanity
Checking partition count
Failure populating temp tables.
Last @dsql: SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
                        SELECT @RowcountOUT = SUM(1) FROM [bizproj].sys.partitions WHERE partition_number > 1 OPTION    ( RECOMPILE );
Msg 50000, Level 16, State 1, Procedure sp_BlitzIndex, Line 1503 [Batch Start Line 6]
bizproj database failed to process. Setting @SkipPartitions = 1 because > 100 partitions were found. To check them, you must set @BringThePain = 1.
```

BrentOzar:
If TableName is specified. Closes #1697.

  - SQL Server 2017

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -678,7 +678,7 @@ SET @NumDatabases = @@ROWCOUNT;
 /* Running on 50+ databases can take a reaaallly long time, so we want explicit permission to do so (and only after warning about it) */
 
 BEGIN TRY
-        IF @NumDatabases >= 50 AND @BringThePain != 1
+        IF @NumDatabases >= 50 AND @BringThePain != 1 AND @TableName IS NULL
         BEGIN
 
             INSERT    #BlitzIndexResults ( Priority, check_id, findings_group, finding, URL, details, index_definition,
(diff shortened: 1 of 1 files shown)


alekseybochkov:
@BrentOzar thanks!

PR #1751, opened 2018-09-07
Make Mode 0 more useful

### LINKED ISSUE #1739: Missing information when running Mode=4 for all databases - M4 tab in CCC
Bug - Missing information, such as borderline duplicate key indexes when looking at the M4 tab in the CCC file. When I run Mode=4 manually in a particular database, they show up. The M4 tab has a much shorter list than when run manually for each database.

Compare the Sheet1 info to the M4 tab in 20180823_125000. M4 contains info for 2 databases. Sheet1 is for Mode=4 for one database, Sheet2 is Mode=4 for the other database. M4 has 39 rows. Just Sheet1 has 124 rows. So BIG different and haven't even added in Sheet2 yet.

Missing information such as borderline missing indexes

The Mode=4 output for each database should match the output when running Mode=4 and GetAllDatabases=1

BlitzErik:
Fixes #1739

Fixes # .

 - Make mode 0 generally more useful to people

How to test this code:
 - Run in @Mode = 0, see important stuff that was missing before

 - SQL Server 2016
  - SQL Server 2017

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -25,7 +25,7 @@ ALTER PROCEDURE dbo.sp_BlitzIndex
     @GetAllDatabases BIT = 0,
     @BringThePain BIT = 0,
     @ThresholdMB INT = 250 /* Number of megabytes that an object must be before we include it in basic results */,
-	@OutputType         VARCHAR(20)     = 'TABLE' ,
+	@OutputType VARCHAR(20) = 'TABLE' ,
     @OutputServerName NVARCHAR(256) = NULL ,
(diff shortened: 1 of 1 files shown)


BlitzErik:
Stuff that changed:
1. High/Low missing index impact just checks usage, leaves out impact. Impact is frequently a bag of wonk.
2. Adds borderline dupes to output
3. Mode 0 will warn about > 21 indexes on a table, Mode 4 will warn about 7
4. Abnormal psychology is no longer in Mode 0 (xml, spatial, column store, compressed)

PR #1758, opened 2018-09-10
#1757 sp_Blitz - checking for statistics

### LINKED ISSUE #1757: sp_Blitz - add check looking for statistics with no histogram contents
Feature

If an object does have rows, but the associated non-filtered statistic has no rows in the histogram (like if it hasn't had an update stats since upgrading/restoring), tell the user to update stats on that object.

Should only run when CheckUserDatabaseObjects = 1, and only if sys.dm_db_stats_histogram exists.

Query for the web page to list all of the affected objects - don't return all these in sp_Blitz though:

```
SELECT 'UPDATE STATISTICS ' + QUOTENAME(SCHEMA_NAME(o.schema_id)) + '.' + QUOTENAME(o.name) + ' WITH FULLSCAN;'
  FROM sys.all_objects o 
  INNER JOIN sys.stats s ON o.object_id = s.object_id AND s.has_filter = 0
  OUTER APPLY sys.dm_db_stats_histogram(o.object_id, s.stats_id) h
  WHERE o.is_ms_shipped = 0 AND o.type_desc = 'USER_TABLE'
    AND h.object_id IS NULL
    AND 0 < (SELECT SUM(row_count) FROM sys.dm_db_partition_stats ps WHERE ps.object_id = o.object_id)
  ORDER BY o.name;
```

BrentOzar:
If you liked Men Without Hats, you'll love Statistics Without Histograms. Closes #1757.

 - SQL Server 2017

### DIFF
--- a/Documentation/sp_Blitz Checks by Priority.md
+++ b/Documentation/sp_Blitz Checks by Priority.md
@@ -6,8 +6,8 @@ Before adding a new check, make sure to add a Github issue for it first, and hav
 
 If you want to change anything about a check - the priority, finding, URL, or ID - open a Github issue first. The relevant scripts have to be updated too.
 
-CURRENT HIGH CHECKID: 218.  
-If you want to add a new one, start at 219.
+CURRENT HIGH CHECKID: 220.  
+If you want to add a new one, start at 221.
 
 | Priority | FindingsGroup | Finding | URL | CheckID |
 |----------|-----------------------------|---------------------------------------------------------|------------------------------------------------------------------------|----------|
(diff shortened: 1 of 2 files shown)


MisterZeus:
Wow, this check (220) blows up for me on SQL Server 13.0.4466, within a 2008-compat database. :-(

I've ran DBCC CHECKDB against the 1 table, but there is still 1 statistic within that table that gives the following error, once it's fed into dm_db_stats_histogram.

I even tried DROP and CREATE on that 1 statistic, but I still get this when running the latest dev sp_Blitz with @Debug = 2:

Running CheckId [220].
Location:	 d:\b\s3\sources\sql\ntdbms\include\typesystem\_retypesbase.inl:338
Expression:	 PbValGetConst()[CSsVariant::OFF_VER] == CSsVariant::CURR_VER
SPID:		 57
Process ID:	 5904
Msg 3624, Level 20, State 1, Line 2
A system assertion check has failed. Check the SQL Server error log for details. Typically, an assertion failure is caused by a software bug or data corruption. To check for database corruption, consider running DBCC CHECKDB. If you agreed to send dumps to Microsoft during setup, a mini dump will be sent to Microsoft. An update might be available from Microsoft in the latest Service Pack or in a Hotfix from Technical Support.
Msg 596, Level 21, State 1, Line 0
Cannot continue the execution because the session is in the kill state.
Msg 0, Level 20, State 0, Line 0
A severe error occurred on the current command.  The results, if any, should be discarded.

I can try and see if I could code out my table and data, and anonymise them, if anyone's interested?
To be honest, this feels like a SQL Server bug, rather than anything to do with sp_Blitz, since check 220 works fine for other stats within the same table!

PabloInNz:
We are getting this exact same issue on the latest version of Sp_blitzIndex running against 2017 Version 14.0.2037.2 Web Edition. Is there anything that we can do (other than ditching Web Edition that is) ?

PR #1767, opened 2018-09-14
Dupe check fix

### LINKED ISSUE #1712: sp_BlitzIndex - shows duplicate keys alert for indexes in different databases
Bug. Not really big bug, like a West-Nile-infested Miami mosquito, and not even a cockroach. Just one of those insects where you go, "Oh hey, I don't really want you in the house, but I'm going to watch you hang out for a minute." Maybe like a ladybug.

If you have 2 databases where object IDs are the same (like two restored copies of the same database), sp_BlitzIndex reports duplicate keys, but the database names are different. See 20180808_132522 in our client files for an example.

Only show dupes if the index is in the same database.

BlitzErik:
Fixes #1712

 - Filter dupe checks for size in both applicable places

How to test this code:
 - Create two indexes with the same key columns, but different includes, so that one is < 250MB, and one is > 250MB. Run `@GetAllDatabases = 1`, watch one show up and the other not.

 - SQL Server 2016
  - SQL Server 2017

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -2150,11 +2150,23 @@ BEGIN;
         RAISERROR('check_id 1: Duplicate keys', 0,1) WITH NOWAIT;
             WITH    duplicate_indexes
                       AS ( SELECT  [object_id], key_column_names, database_id, [schema_name]
-                           FROM        #IndexSanity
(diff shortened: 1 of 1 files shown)


MisterZeus:
aliased as ip because is is a keyword?

PR #1896, opened 2018-12-21
dbo.SqlServerVersions - little change in object existence check + allow script execution

jeffchulg:
 - object existence includes schema name (with OBJECT_ID built-in function)
 - a "GO" between object creation and DELETE statement
- a "GO" at the end of the script

How to test this code:
- run it as you are used to

 - SQL Server 2014

### DIFF
--- a/SqlServerVersions.sql
+++ b/SqlServerVersions.sql
@@ -1,5 +1,5 @@
 
-IF NOT EXISTS (SELECT NULL FROM sys.tables WHERE [name] = 'SqlServerVersions')
+IF (OBJECT_ID('dbo.SqlServerVersions') IS NULL)
 BEGIN
 
     CREATE TABLE dbo.SqlServerVersions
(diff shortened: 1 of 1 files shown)


BlitzErik:
@jeffchulg thanks for this. Can you open a corresponding issue for it? I'll get it merged in as soon as you do.

BlitzErik:
@jeffchulg doing this here isn't the best idea, because when the script gets regenerated, it'll be overwritten. It'd have to happen [over here](https://github.com/jadarnel27/SqlServerVersionScript). Open an issue for @jadarnel27 in his repo.

jadarnel27:
@jeffchulg yep, I'm happy to make that change if you want to create an issue.  Or if you're comfortable with C#, I'm open to a PR as well!

jeffchulg:
@jadarnel27  yes, it could be great. 
I'm not comfortable with C#

PR #1902, opened 2019-01-03
Issue 1894 validate database parameter

### LINKED ISSUE #1894: sp_DatabaseRestore: validate @Database against FULL backup file
**Version of the script**
Commit from Issue 1185

Imagine one of these calls

EXEC _DBA.dbo.sp_DatabaseRestore
    @Database = 'SomeDB',
    @RestoreDatabaseName = 'SomeDB**Test**'
    @BackupPathFull = '\\blah\SomeDB**Copy**',
    @MoveFiles = 1,
   ...
 or

EXEC _DBA.dbo.sp_DatabaseRestore
    @Database = 'SomeDB',
    @BackupPathFull = '\\blah\SomeDB**Copy**',
   ...

SQL Pattern matching means we can restore SomeDBCopy because it matches SomeDB in filters like
     BackupFile LIKE N'%' + @Database + N'%'
And file moves don't work as expected because of the simple REPLACE

Fixing the LIKE/REPLACE code is possible but we can validate that the full backup files does actually match the @Database parameter.

Pick any database naming convention as above

We restore exactly 
Tested here on SQL 2012+ 2017 so assum all versions have the same issue

ShawnCrocker:
Fixes #1894 

- Validate @Database against backup file
- Add other validation
- Side effects: some reformatting so I could follow logic

How to test this code:
 - Define @Database parameter different to the database backup file

 - SQL Server 2012
 - SQL Server 2016


BrentOzar:
Ooo, man, do we really need to change the line endings? The diff looks like a hot mess, so it's tough to test.

ShawnCrocker:
It's mostly nesting of BEGIN/END blocks. I went through so i could follow the logic.

Although... my local pull had line ending changes before I re-committed.

BrentOzar:
Woohoo! Nice job, sir. Thanks!

PR #1913, opened 2019-01-04
Issue 1910

M-Ron:
Fixes # 1910.

 - Add the same SQL Server version check as in sp_BlitzCache for the @BlitzCacheSortOrder parameter.
 - If it fails the check, don't set @BlitzCacheSortOrder to 'memory grant'.

How to test this code:
 - Kinda tricky to test as you'd need an older SQL Server instance that can trigger CheckID 34.

 - SQL Server 2008
 - SQL Server 2016


BrentOzar:
Nice work!

PR #1939, opened 2019-02-04
Fixes #1938 - sp_BlitzIndex create table script modification 

### LINKED ISSUE #1938: sp_BlitzIndex - Remove tablename from index create scripts
**Version of the script**
current (7.2)

table name is included in the INDEX CREATE scripts. In server tuning Brent mentioned he would prefer if this was removed. Going to also sort out capitalization while i'm there.

RichBenner:
Fixes #1938  .

 - removed table name from create scripts
 - some capitalisation

How to test this code:
 - run blitzindex and check the 'create index' recommendations. shouldn't include tablename and should be nicely capitalized

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -535,13 +535,13 @@ IF OBJECT_ID('tempdb..#FilteredIndexes') IS NOT NULL
                     END + CASE WHEN included_columns IS NOT NULL THEN N'INCLUDES: ' + included_columns + N' '
                         ELSE N''
                     END,
-                [create_tsql] AS N'CREATE INDEX [ix_' + table_name + N'_' 
+                [create_tsql] AS N'CREATE INDEX [IX_ 
                     + REPLACE(REPLACE(REPLACE(REPLACE(
                         ISNULL(equality_columns,N'')+ 
                         CASE WHEN equality_columns IS NOT NULL AND inequality_columns IS NOT NULL THEN N'_' ELSE N'' END
                         + ISNULL(inequality_columns,''),',','')
                         ,'[',''),']',''),' ','_') 
-                    + CASE WHEN included_columns IS NOT NULL THEN N'_includes' ELSE N'' END + N'] ON ' 
+                    + CASE WHEN included_columns IS NOT NULL THEN N'_Includes' ELSE N'' END + N'] ON ' 
                     + [statement] + N' (' + ISNULL(equality_columns,N'')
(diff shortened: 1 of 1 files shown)


RichBenner:
missed a ', i'll resubmit

tgspacelabs:
Please reinstate the TABLE NAME.  This is screwing up all my code as I use the same column names in numerous tables so when I copy the index creation script portion I have to edit each one to replace the missing table names.

BrentOzar:
@tgspacelabs why? What do you need the table name in the index name for?

tgspacelabs:
To keep them from colliding with each other:
Example:
CREATE TABLE [Test1]
    ([ID] INT NOT NULL,
     [Column1] VARCHAR(30) NOT NULL);

CREATE TABLE [Test2]
    ([ID] INT NOT NULL,
     [Column1] VARCHAR(30) NOT NULL);

CREATE NONCLUSTERED INDEX [IX_Column1] ON [dbo].[Test1]([Column1]);

--From inspecting just the index name, which table has the index for Column1?  If we had the table name as before it would be clear which table the index refers to.

Thanks,
Tony Green
Spacelabs Healthcare

BrentOzar:
What do you mean colliding? You can have two indexes with the same name. This works fine:

```
CREATE TABLE [Test1]
    ([ID] INT NOT NULL,
     [Column1] VARCHAR(30) NOT NULL);

CREATE TABLE [Test2]
    ([ID] INT NOT NULL,
     [Column1] VARCHAR(30) NOT NULL);

CREATE NONCLUSTERED INDEX [IX_Column1] ON [dbo].[Test1]([Column1]);
CREATE NONCLUSTERED INDEX [IX_Column1] ON [dbo].[Test2]([Column1]);
```

tgspacelabs:
Yes, but it’s still not clear by inspection which table the index belongs to.

Thanks,
Tony Green
Spacelabs Healthcare

BrentOzar:
On inspection of what? If you're looking at a query plan, the table's name is in front of the index. If you're looking at the create index statement, it's right there in the same line.

tgspacelabs:
If you don’t want to do this, that’s fine.  I’ll just use other tools.

BrentOzar:
No problem - if you want this, just come back and make a clear case for why you need it, and we'll go from there.

tgspacelabs:
What was the rationale for removing it in the first place?

Thanks,
Tony Green
Spacelabs Healthcare

BrentOzar:
Because when you're reviewing query plans, long index names make it hard to see what SQL Server is filtering on. If you've got a long table name, your plans look like this:

Screen Shot 2019-04-30 at 7 32 28 AM

Issue #1948, opened 2019-02-08
sp_BlitzCache - Version Check returning datasets


jeffchulg:
**Version of the script**
Look in the stored procedure, and it'll have a version date & number near the top. Put that in here. If it's not the current version (dated in the last month), then upgrade to the current version and test that before reporting a bug - we fix a lot of stuff in each new build. We'll flat out close bug reports for older builds.

2019.01.01
**What is the current behavior?**
It's not possible to only get back the VersionDate without running the actual code...

**If the current behavior is a bug, please provide the steps to reproduce.**
DECLARE @BlitzCacheVersionDate  DATETIME2;
EXEC sp_BlitzCache @SkipAnalysis = 1 , @VersionDate = @BlitzCacheVersionDate OUTPUT;

**What is the expected behavior?**
If you do following commands, you can get back the version without actually execute the whole procedure logic:
EXEC sp_Blitz @Help = 1, @VersionDate = @BlitzVersionDate OUTPUT ;
EXEC sp_BlitzFirst @Help = 1, @VersionDate = @BlitzFirstVersionDate OUTPUT;

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
all

BrentOzar:
Sure, if you'd like to code that, we'd welcome that as a contribution. Here's the contributing guide to get started:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/CONTRIBUTING.md

If you didn't intend to code it, no worries - we'll close this in a few days if you don't respond. Thanks!

BrentOzar:
Great job! Merged into dev, and it'll be in the February release along with your name in the credits. Love the idea and the execution.

PR #1949, opened 2019-02-08
Add Version Checking  for all SP

jeffchulg:
Fixes # .

 - In sp_BlitzInMemoryOLTP => changed internal variable name from @Version to @MSSQLVersion
 - In all stored procs, added @Version VARCHAR(30) as OUTPUT parameter
 - In all stored procs, added @VersionCheckMode BIT which when set to 1 ends almost directly the SP

How to test this code:
Call each SP as usual
Call each SP with @VersionCheckMode = 1
For instance:

DECLARE @BlitzCacheVersion     VARCHAR(30);
DECLARE @BlitzCacheVersionDate DATETIME2;
EXEC dbo.sp_BlitzCache @VersionCheckMode = 1, @Version = @BlitzCacheVersion OUTPUT, @VersionDate = @BlitzCacheVersionDate OUTPUT ;
SELECT @BlitzCacheVersion as BlitzCacheVersion , @BlitzCacheVersionDate as BlitzCacheVersionDate

 - SQL Server 2016

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -16,7 +16,7 @@ ALTER PROCEDURE dbo.sp_BlitzLock
 	@EventSessionPath VARCHAR(256) = 'system_health*.xel', 
 	@Debug BIT = 0, 
 	@Help BIT = 0,
-	Version     VARCHAR(30) = NULL OUTPUT,
+	@Version     VARCHAR(30) = NULL OUTPUT,
 	@VersionDate DATETIME = NULL OUTPUT,
     @VersionCheckMode BIT = 0
 )
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -16,7 +16,9 @@ ALTER PROCEDURE dbo.sp_BlitzLock
 	@EventSessionPath VARCHAR(256) = 'system_health*.xel', 
 	@Debug BIT = 0, 
 	@Help BIT = 0,
-	@VersionDate DATETIME = NULL OUTPUT
+	Version     VARCHAR(30) = NULL OUTPUT,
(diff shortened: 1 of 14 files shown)


sp_BlitzLock.sql
```diff
@@ -16,7 +16,9 @@ ALTER PROCEDURE dbo.sp_BlitzLock
 	@EventSessionPath VARCHAR(256) = 'system_health*.xel', 
 	@Debug BIT = 0, 
 	@Help BIT = 0,
-	@VersionDate DATETIME = NULL OUTPUT
+	Version     VARCHAR(30) = NULL OUTPUT,
```
BrentOzar: Thanks for the contribution, but can you test your work before doing the pull request?  Just at a glance, this won't compile. Try running all of the submissions. Thanks!

jeffchulg:
Addresses #1948

jeffchulg:
Corrected... sorry...

sp_ineachdb.sql
```diff
@@ -25,15 +25,21 @@ ALTER PROCEDURE dbo.sp_ineachdb
   @is_broker_enabled   bit = NULL,
   @user_access         nvarchar(128)  = NULL, 
   @Help                BIT = 0,
-  @VersionDate DATETIME = NULL OUTPUT
+  @Version             VARCHAR(30)    = NULL OUTPUT,
+  @VersionDate         DATETIME       = NULL OUTPUT,
+  @VersionCheckMode    BIT            = 0
```
BrentOzar: SO CLOSE! This one wasn't working because the proc requires an input parameter for cmd. It fails if you call it with VersionCheckMode only. I'm going to set the command to be a nullable input parameter after I accept your pull request. Great work overall though! Love the idea and the execution. Good job.

PR #1953, opened 2019-02-08
Update sp_BlitzCache.sql

RonMacNeil:
Complete the join between #missing_index_detail and ##bou_BlitzCacheProcs as otherwise on servers running jobs with many steps, each with missing indices, will cause #missing_index_pretty to grow exponentially, causing sp_BlitzCache to run for hours.

Fixes # .

 - SQL Server 2008
 - SQL Server 2012
 - SQL Server 2014
 - SQL Server 2016
 - SQL Server 2017


BrentOzar:
That's so awesome. Nice find, sir. I merged it into dev, and it'll be in the February release along with a credit to you. Thanks!

PR #1962, opened 2019-02-13
Make sp_BlitzQueryStore installable on versions lower than 2016 but failing

### LINKED ISSUE #1961: Make sp_BlitzQueryStore installable on versions lower than 2016 but failing
I would like to be able to install all the stored procedures from this package.
In other FRK scripts, when something is not "as expected", the SP fails, like in sp_BlitzIndex when the script is outdated.

Plus, the FRK would have only one "all-in-one" installation script. 

As the QueryStore feature is not available below 2016, the condition defined before the SP creation could be integrated inside its code, and `DROP IF EXISTS` statement should be put in dynamic statements `EXEC sp_executesql N'DROP IF EXISTS;';`.
I've tested on a SQL Server 2014 and it works like a charm.

Let as is

As much as we'd love to build everything that everyone wants for free, we need your help. Open source is built with your help and code. Are you ready to commit time to this project? Have you got existing code you can help contribute to solve the problem?

Yes, accepted (and see above).

jeffchulg:
Fixes #1961 .

 - Moved SQL Server version checking inside the SP
 - Dynamic Query for `DROP TABLE IF EXISTS`
 - Dynamic Query for index creation on temporary tables
 - Dynamic Query for "database with query store enabled check"

How to test this code:
 - `EXEC dbo.sp_BlitzQueryStore` -- fails before SQL Server 2016
 - `EXEC dbo.sp_BlitzQueryStore @DatabaseName = 'XXX'` works on SQL Server 2016
 - 'EXEC dbo.sp_BlitzQueryStore @VersionCheckMode = 1`  -- works in any case

 - SQL Server 2008 R2
 - SQL Server 2014
 - SQL Server 2016

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -549,7 +549,7 @@ CREATE TABLE #working_plan_text
 	top_three_waits NVARCHAR(MAX)
 ); 
 
-EXEC sp_executesql N'CREATE CLUSTERED INDEX wpt_ix_ids ON #working_plan_text(plan_id, query_id, query_hash);';
+EXEC sp_executesql N'CREATE CLUSTERED INDEX wpt_ix_ids ON #working_plan_text(plan_id, query_id, query_plan_hash);';
 
 /*
 This is where we store warnings that we generate from the XML and metrics
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -8,23 +8,6 @@ SET STATISTICS IO OFF;
 SET STATISTICS TIME OFF;
 GO
 
-DECLARE @msg NVARCHAR(MAX) = N'';
-
-	-- Must be a compatible, on-prem version of SQL (2016+)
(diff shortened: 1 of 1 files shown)


jeffchulg:
The same kind of operation could eventually be done for sp_BlitzInMemoryOLTP...

BrentOzar:
Hi Jeff. Ooo, thanks for the commit, but here's the thing: dynamic SQL is a whole lot harder to write and troubleshoot.

I recognize that this change would make your life a little easier around installation time, but it makes the maintainers' and authors' lives a lot harder. I need to make authoring and testing as easy as practical, which means unless dynamic SQL is really needed desperately for something, then we'd probably rather avoid it.

Going forward, can you make an issue before you want to make a change, and then discuss the change you're looking for in the issue? We talk about this in the contributing guide (linked on the home page of the project.) I hate telling you that, because your code is great and you're a good contributor, but it'll just save you some heartache on changes like this.

jeffchulg:
Hi Brent, no problem... It did not take me a lot of time... and I was not sure it would be accepted anyway !
But it illustrated my request.

Issue #1994, opened 2019-03-06
sp_Blitz - Add Check for all FRK SP version date is the same


jeffchulg:
**Is your feature request related to a problem? Please describe.**

As I added the capability to run all the SP in VersionCheckMode, the next step could be to run a check for "schema consistency" (and I think sp_Blitzis the right candidate to do it).

**Describe the solution you'd like**

In sp_BlitzFirst, create a temporary table which contains the name of each SP, except sp_Blitz.

Run a loop that will:
0) Create a variable Highest
1) Check the procedure exists
1a=> if not exists ==> continue
1b => if exists:
   run the stored procedure in VersionCheckMode (using dynamic SQL) and maybe between TRY...CATCH.
   if the version for at least one SP is older than the one from sp_Blitz, then there is an inconsistent installation
   If the version for at least one SP is higher than the one from sp_Blitz, then there is an insconsistent installation
   

**Describe alternatives you've considered**

Create a new dedicated stored procedure

**Are you ready to build the code for the feature?**
As much as we'd love to build everything that everyone wants for free, we need your help. Open source is built with your help and code. Are you ready to commit time to this project? Have you got existing code you can help contribute to solve the problem?

Why not... it does not seem to be difficult or time-consuming to implement...

BrentOzar:
I love it! That's a great idea. Things to think about:

* One or more of the scripts may not exist (and that's OK, shouldn't flag a warning that there's an inconsistent installation)
* The scripts may be installed in the master database, but you may be in a different database context
* The scripts may be installed in a user database, and you may be in the same database context (example: users install it in a "DBA" database and run them from that same database, without db/schema prefixes)
* The scripts may be installed in a user database, and you may be in a DIFFERENT database context (example: installed in "DBA", but run from tempdb with fully prefixed names, like "EXEC DBA.dbo.sp_Blitz")

jeffchulg:
HI Brent,

the use of DB_NAME() inside sp_Blitz should solve the "the scripts may be installed" sentences, shouldn't it?

BrentOzar:
Yes, as long as you use that to build fully qualified calls to the procs, that will work, I bet.

jeffchulg:
Hi Brent,

could you please review the code attached... I don't want to insert it directly to sp_Blitz before it's ok for you...
[...]

PR #1997, opened 2019-03-18
First responder consistency check

jeffchulg:
Addresses #1994 

Tested on SQL Server 2008 R2 and 2014 with missing components.

I added the schema name in the equation because I'm transferring all the First Responder Kit objects to another schema (FirstResponderKit) with a custom PS script. 

It's possible, for instance for testing purpose, that objects resides in both 'FirstResponderKit' and 'dbo' schemas.

The main idea here was to generate the set of commands to be executed so that the overhead of adding a new object is minimal.

As I'm writing, it would also be easy to add to this check the capability to pinpoint depreciated components (like sp_BlitzTrace), but this would be the subject of another pull request.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -4341,6 +4341,9 @@ IF(
     )
 )
 BEGIN
+
+    IF @Debug IN (1, 2) RAISERROR('Running check with id %d',0,1,2000);
+
     SET @spBlitzFullName                    = QUOTENAME(DB_NAME()) + '.' +QUOTENAME(OBJECT_SCHEMA_NAME(@@PROCID)) + '.' + QUOTENAME(OBJECT_NAME(@@PROCID));
     SET @BlitzIsOutdatedComparedToOthers    = 0;
     SET @tsql                               = NULL;
@@ -4409,7 +4412,7 @@ BEGIN
                 '    UNION ALL ' + @crlf +
                 '    SELECT ''sp_ineachdb'',''P'',1' + @crlf +
                 '    UNION ALL' + @crlf +
-                '    SELECT ''SqlServerVersions'',''U'',0' + @crlf +
+                '    SELECT ''SqlServerVersions'',''U'',1' + @crlf +
                 ')' + @crlf +
                 'INSERT INTO #FRKObjects (' + @crlf +
                 '    DatabaseName,ObjectSchemaName,ObjectName, ObjectType,MandatoryComponent' + @crlf +
@@ -4474,10 +4477,11 @@ BEGIN
     WHERE ObjectType = 'P' 
     AND ObjectSchemaName IS NOT NULL
     ;
-    /* For debug 
-    SELECT * 
-    FROM #StatementsToRun4FRKVersionCheck ORDER BY SubjectName,SubjectFullPath,StatementId  -- in case of schema change  ;
-    */
+    IF(@Debug in (1,2))
+    BEGIN
+        SELECT * 
+        FROM #StatementsToRun4FRKVersionCheck ORDER BY SubjectName,SubjectFullPath,StatementId  -- in case of schema change  ;
+    END;
     
     
     -- loop on queries...
@@ -4505,7 +4509,7 @@ BEGIN
             BREAK;
         END;
 
-        RAISERROR('Statement: %s',0,1,@tsql);
+        IF @Debug IN (1, 2) RAISERROR('    Statement: %s',0,1,@tsql);
 
         -- we start a new component
         IF(@PreviousComponentName IS NULL OR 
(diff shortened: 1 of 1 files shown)


sp_Blitz.sql
```diff
@@ -201,6 +232,47 @@ AS
...
+        -- This one will contain the statements to be executed
+        -- order: 1- Mandatory, 2- VersionCheckMode, 3- VersionCheck
+
+        CREATE TABLE #StatementsToRun4FRKVersionCheck (
+            StatementId                 INT IDENTITY(1,1),
+            CheckName                   VARCHAR(256),
+            SubjectName                 VARCHAR(256),
+            SubjectFullPath             VARCHAR(1024),
+            StatementText               NVARCHAR(MAX),
+            StatementOutputsCounter     BIT, -- currently not used
```
BrentOzar: If it's not used, let's not add it for now.

sp_Blitz.sql
```diff
@@ -4256,6 +4328,511 @@ IF @ProductVersionMajor >= 10
...
+                    
+/*This checks that First Responder Kit is consistent. 
+It assumes that all the objects of the kit resides in the same database, the one in which this SP is stored
+It also is ready to check for installation in another schema.
+*/
+IF(
+    NOT EXISTS ( 
+        SELECT  1
+        FROM    #SkipChecks
+        WHERE   DatabaseName IS NULL AND CheckID = 2000
```
BrentOzar: Let's use the next CheckID from the Documentation folder. There's a file in there with the check IDs, and you can add this new one.

sp_Blitz.sql
```diff
@@ -4256,6 +4328,511 @@ IF @ProductVersionMajor >= 10
...
+                INSERT  INTO #BlitzResults( 
+                    CheckID ,
+                    Priority ,
+                    FindingsGroup ,
+                    Finding ,
+                    URL ,
+                    Details
+                )
+                SELECT 
+                    2000 AS CheckID ,
```
BrentOzar: You'll want to use different CheckIDs for different warnings. (People have built tools on top of sp_Blitz's output, and they handle the different CheckIDs differently - think of them like sys.messages.)

sp_Blitz.sql
```diff
@@ -4256,6 +4328,511 @@ IF @ProductVersionMajor >= 10
...
+                    1 AS Priority ,
```
BrentOzar: For priority, see the list of checks in the documentation folder. Because varying versions of sp_Blitz* aren't really that big of a deal (relative to, say, database corruption or not having backups), I'd put them down in the informational section, like in the 200s.

sp_Blitz.sql
```diff
@@ -4256,6 +4328,511 @@ IF @ProductVersionMajor >= 10
...
+                    'Reliability' AS FindingsGroup ,
+                    'Component ' + @CurrentComponentFullName + ' is not at the minimum version required to run this procedure' AS Finding ,
+                    'https://www.BrentOzar.com/blitz/' AS URL ,
+                    'VersionCheckMode has been introduced in component version date "20190219". This means its version is from before that date.' AS Details;
+                ;
```
BrentOzar: I would just flag the component as out of date and move on - if sp_Blitz has this code, and the other procs aren't at a compatible version, then you know you've got a mismatch. No need for a separate CheckID there, too, nice bonus.

sp_Blitz.sql
```diff
@@ -4256,6 +4328,511 @@ IF @ProductVersionMajor >= 10
...
+                    FindingsGroup ,
+                    Finding ,
+                    URL ,
+                    Details
+                )
+                SELECT 
+                    2000 AS CheckID ,
+                    250 AS Priority ,
+                    'Informational' AS FindingsGroup ,
+                    'First Responder kit component ' + @CurrentComponentFullName + ' is at the expected version' AS Finding ,
```
BrentOzar: If no action needs to be taken, don't output anything - only return rows in sp_Blitz where users need to take an action.

sp_Blitz.sql
```diff
@@ -6672,7 +7249,7 @@ IF @ProductVersionMajor >= 10
 						(@@SERVERNAME IS NOT NULL
 						AND
 						/* not a named instance */
-						CHARINDEX('\',CAST(SERVERPROPERTY('ServerName') AS NVARCHAR(128))) = 0
+						CHARINDEX(CHAR(92),CAST(SERVERPROPERTY('ServerName') AS NVARCHAR(128))) = 0
```
BrentOzar: Is there a reason we made this change?

jeffchulg: It's just that, in my code editor, backslash is interpreted as escaping character. 
I'm used to replace any backslash char by CHAR(92).
I can remove this if it's not a good idea.

Issue #1985, opened 2019-02-23
sp_BlitzCache: use the OS's default decimal separators (commas vs periods) in output


BrentOzar:
**Version of the script**
SET @Version = '7.3';
SET @VersionDate = '20190219';

**What is the current behavior?**
When run on a VM with the default decimal separator as comma, and default digit grouping separator as period:

![screen shot 2019-02-23 at 4 21 28 am](https://user-images.githubusercontent.com/245462/53286389-8fa6cd80-3722-11e9-8da4-085a5c231299.png)

sp_BlitzCache isn't honoring that:

![screen shot 2019-02-23 at 4 19 36 am](https://user-images.githubusercontent.com/245462/53286391-9df4e980-3722-11e9-8914-702a3147b595.png)

**What is the expected behavior?**
The output should match the user's regional settings.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Has probably been this way since the dawn of time. (It's present in other scripts too, like sp_BlitzIndex's text column for object size, and sp_BlitzFirst, but let's tackle this one first.)

NielsSchNL:
This is my first-ever reply to a GitHub project. I'm eager to participate but please forgive me if I'm not following the correct code of conduct. 

I'd be happy to take a swing at this if you'd like. However, I'm not clear on which regional settings should be honored. Are we trying to honor the server OS regional settings or the settings the client OS creating the session (for instance SSMS)?

AFAIK (and please correct me if I'm wrong) the client OS regional settings are not visible in a session. Each session simply defaults to the server/database's settings (as shown in @@LANGUAGE). The server OS regional settings may be found in sys.dm_os_windows_info but that won't cover every scenario. The best way to move forward on this may be to introduce a parameter.

Sidenote: if another locale would've been honored then sp_BlitzCache would break anyway as there's some hard-coded REPLACE going on to strip '.00'-decimals in the output.

BrentOzar:
Hi, and welcome to the project! It'd be the server's default settings - the same thing you see when you run this:

`SELECT CAST(1 AS DECIMAL(10,2)) / 10 AS OneTenth`

Depending on your server's defaults, you may see either comma separators or decimal separators there.

NielsSchNL:
I've created a pull request with my changes ( #2012 ). I've tested it as far as I could. Seems to work just as expected now, based on the session's current language setting.

[...]

PR #2012, opened 2019-03-25
#1985 Changed output to honor session's regional setting

NielsSchNL:
Changed the output to reflect current session's regional settings (regarding numeric format, decimal and grouping separator). 
The original CAST(AS MONEY) was extended to FORMAT(CAST(x AS MONEY), 'N'), making sure the number is formatted according to regional settings. 
The logic to strip zero-valued decimals to integers was corrected to support the now dynamic decimal separator, which in turn is determined by simply testing the output of a FORMAT statement. AFAIK there is no way to pull the current decimal separator from SQL Server's configuration. If there is, that method should be used here. 

Can be tested by changing the current session's language. For instance:
SET LANGUAGE Dutch;
Will set the session's language to Dutch (obviously) and therefore use a comma as decimal separator and period as grouping separator.

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -4800,7 +4800,6 @@ BEGIN
     EXEC sp_executesql @sql, N'@Top INT, @min_duration INT, @min_back INT, @minimumExecutionCount INT', @Top, @DurationFilter_i, @MinutesBack, @MinimumExecutionCount;
 END;
 
-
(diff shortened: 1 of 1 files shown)


BrentOzar:
Been a while since I've requested the changes, so I'm going to close this pull request just to keep the First Responder Kit repo a little cleaner. If you want to tackle it again though, feel free - I love the idea of this!

PR #2040, opened 2019-05-01
Moved SQL header build outside of If block

troyjennings:
To resolve Issue 2036, moved SQL header syntax build outside of if block so that it is constructed when the full backup path is null.


BrentOzar:
Looks good! Thanks sir.

PR #2046, opened 2019-05-09
Update sp_blitzcache to avoid arithmetic overflow errors

ianmanton:
Updated sp_blitzcache to avoid arithmetic overflow errors - table definitions had changed to BIGINT but the actual logic in this section still casted values as money which was breaking the ceiling for this on our setup.  I've pretty much blanket changed everything except averages to be bigint in the section it was tripping over on and have been running for a few days with no failrues since so looks ok and worth someone who knows what they're doing casting an eye over it!

Not sure if I've done this process right.... hopefully!


ianmanton:
That said the powerbi report now crawls through loading anything from spblitz, and the resource intensive queries section does show any results ... not sure if that's as a resutl of my change or not though!

BrentOzar:
This is a little tricky: when you do this, you're losing fidelity on displayed results. The top is the old version, the bottom is after the rounding to ints:

Screen Shot 2019-07-02 at 6 58 21 AM

My first reaction is that I hate to lose fidelity, but the more I look at these results, the more I think the rounding makes sense. Nice work! Merging in, will credit you in this month's release notes.

ianmanton:
I probably just blanket altered anything money to bigint when I should have been more specific in which columns I targeted (as believing all main targets were integers not decimals)

Perhaps worth scaling back and not applying the line changes to the first 6 columns in the screenshot to retain the fidelity where it’s more relevant?  Your call however! 😊

Cheers
Ian

PR #2061, opened 2019-06-04
Update sp_ineachdb.sql

sm8680:
Added 2 Variables 
@SQLVersion: This was implemented so that large organizations with SQL Server 2008 and R2 can use this sproc.  Currently without this change SQL Server versions 2008 and 2008R2 will error with the following.  Msg 208, Level 16, State 1, Procedure sp_ineachdb, Line 176
Invalid object name 'sys.dm_hadr_database_replica_states'.

@ServerName:  This was added to account for VMware SnapShots at times @@SERVERNAME will either come up NULL or with old server name.  CONVERT(sysname, SERVERPROPERTY('ServerName')) is a better option and reduces minor nuisances.  

Feel free to blend in the code if you feel this brings value to the project.


BrentOzar:
Good work! Added to the July release, will credit you in the release notes. Thanks!

PR #2101, opened 2019-08-30
Issue 2063 selects with writes

erikdarlingdata:
This will flag any select queries that have caused writes, with a few boundaries:
- They're not already flagged by the index spool check
- They're not already flagged as a "big spill" in 2016+ where supported

This should catch things discussed [here](https://dba.stackexchange.com/questions/191825/).

Some notes:
- Doc for BlitzCache was updated
- Also adds all the necessary warning-generation code
- This changes the `##BlitzCacheProcs`, so you'll have to update it in the collector
- I added placeholder URL to a dba.se question, but feel free to replace that if you'd like when local content is available


BrentOzar:
Great job! Thanks sir.

Issue #2051, opened 2019-05-14
sp_BlitzCache: add @SortOrder = 'query hash'


BrentOzar:
**Is your feature request related to a problem? Please describe.**
A lot of folks aren't parameterizing their queries, so the plan cache is full of one-offs. They want to solve performance issues, but they're not sure where to start.

**Describe the solution you'd like**
Either of these would be good:

* @SortOrder = 'query hash' - get the top 10 query hashes grouped from sys.dm_exec_query_stats, sorted by, say, duration, and show those in the grid. This might be easier to run it like we run @SortOrder = 'all' - by getting the list of top 10 query hashes, and then running sp_BlitzCache again with just those query hashes - or...
* @GroupByQueryHash = 1 - add a group by query hash to the dynamic SQL. I think this is going to be spectacularly hard though.

BrentOzar:
@erikdarlingdata I was testing it, and it's reeeeally close, but the qhg variable ends up with a leading single quote, so the first hash doesn't go in. To see what I mean:

```
DropIndexes;
GO
CREATE INDEX IX_DisplayName ON dbo.Users(DisplayName);
GO
DBCC FREEPROCCACHE;
GO
SELECT * FROM dbo.Users WHERE UPPER(DisplayName) = 'Brent Ozar';
GO
SELECT * FROM dbo.Users WHERE UPPER(DisplayName) = 'Erik Darling';
GO
SELECT * FROM dbo.Users WHERE LOWER(DisplayName) = 'Brent Ozar';
GO
SELECT * FROM dbo.Users WHERE LOWER(DisplayName) = 'Erik Darling';
GO
sp_BlitzCache @SortOrder = 'query hash'
GO
```

Only the lowers show up. To diagnose it, in sp_BlitzCache, I went to where you're setting qhg, and added a couple of diagnostic lines at the end of this:

```
    SELECT TOP (1)
	         @qhg = STUFF((SELECT DISTINCT N',' + CONVERT(NVARCHAR(MAX), qhg.query_hash, 1) 
    FROM #query_hash_grouped AS qhg 
    WHERE qhg.query_hash <> 0x00
    FOR XML PATH(N''), TYPE).value(N'.[1]', N'NVARCHAR(MAX)'), 1, 1, N'''')
	OPTION(RECOMPILE);
    
    SELECT * FROM #query_hash_grouped;
    SELECT @qhg AS qhg;
```

And that's how I found the leading single quote:

[...]

PR #2102, opened 2019-08-30
Issue 2051 query hash sort

erikdarlingdata:
Grabs top 10 highest cpu consuming plans with highest count of query hashes (multiple plans), and runs BlitsCache to find them.

Some quick notes:
- Doesn't change anything for the collector
- Adds a new sort order! Documentation for `@Help` was updated to reflect
- Adds a GOTO 👎 
- I added a self-imposed query requirement to group by database id, to avoid queries with the same hash that run in multiple databases which may be parameterized, etc.
- The rest is pretty self-explanatory: BlitzCache is re-executed with all passed in parameters, and the list of query hashes gathered
- Sort order is not passed to the calling block to prevent an infinite loop 💯 💯 💯 
- Sort order defaults back to cpu, but I'm fine with that changing (or if you want a `@SecondarySortOrder` parameter)

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -345,7 +345,7 @@ BEGIN
     UNION ALL
     SELECT N'@SortOrder',
            N'VARCHAR(10)',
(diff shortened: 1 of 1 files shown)


BrentOzar:
Woohoo! Awesome, will dig into this this weekend.

sp_BlitzCache.sql
```diff
@@ -6434,6 +6441,70 @@ END;
...
+                    FROM   sys.dm_exec_plan_attributes(qs.plan_handle) AS pa
+                    WHERE  pa.attribute = 'dbid' ) AS ca
+    GROUP BY qs.query_hash, ca.value
+    HAVING COUNT_BIG(*) > 1
+    ORDER BY max_worker_time DESC,
+             records DESC;
+    
+    DECLARE @qhg NVARCHAR(MAX) = N''
+    
+    SELECT TOP (1)
```
BrentOzar: This qhg string ends up with a leading single quote, so the first plan hash doesn't get included. I'll put notes in the issue on how to check it.

BrentOzar:
This pull request is for #2051, seeing if adding that link will help me assign the issue....

PR #2126, opened 2019-09-23
Fix Cannot resolve collation conflict between "Language_A" and "Language_B" in add operator occurring in ORDER BY statement column 3.

chaowlert:
I have problem when I run sp_BlitzIndex on database with different collation to tempdb. And found problem on step that script casts int to nvarchar. Casting will create different collation to temp table, and it causes exception when sorting.

I fixed it by remove casting. int can be concatenated to nvarchar without casting.


BrentOzar:
Great, thanks! I've added it to the dev branch, and it'll be in the October release, and we'll credit you in the release notes. Have a great week!

chaowlert:
Hi sorry, I'm really sorry. Please revert this PR. I created PR without proper test. 

It seemed like `nvarchar` + `int` will produce exception. I don't get exception at the first time because it returns 0 row. I will create new PR soon.

PR #2133, opened 2019-09-25
Fix Cannot resolve collation conflict between "Language_A" and "Language_B" in add operator occurring in ORDER BY statement column 3.

chaowlert:
This is to fix collation problem when running sp_BlitzIndex. Solution in #2126 will produce exception when rows exists.


BrentOzar:
Hahaha, no problem, thanks.

PR #2137, opened 2019-10-01
#2136 Fix BlitzQS DB Name Comparison for Query Store Check.

### LINKED ISSUE #2136: BlitzQueryStore Azure QS Check Failing due to incorrect result from DB_ID
**Version of the script**
Version = '3.8', 
VersionDate = '20190922';

Here is the relevant piece of code:
```/*Does it have Query Store enabled?*/
RAISERROR('Making sure [%s] has Query Store enabled', 0, 1, @DatabaseName) WITH NOWAIT;
IF 	
	((DB_ID(@DatabaseName)) IS NOT NULL AND @DatabaseName <> '')
AND		
	(   SELECT DB_NAME(d.database_id)
		FROM sys.databases AS d
		WHERE d.is_query_store_on = 1
		AND d.user_access_desc='MULTI_USER'
		AND d.state_desc = 'ONLINE'
		AND DB_NAME(d.database_id) = @DatabaseName ) IS NULL
BEGIN
	RAISERROR('The @DatabaseName you specified ([%s]) does not have the Query Store enabled. Please check the name or settings, and try again.', 0, 1, @DatabaseName) WITH	NOWAIT;
	RETURN;
END;
```

Currently when checking if query store is enabled, `sp_blitzquerystore` compares the `database_id` value from `sys.databases` against the result returned from the `DB_NAME` function. The issue with this is that in Azure SQL DB, the database IDs returned using functions do not always match up with those in `sys.databases`. This is documented [here](https://docs.microsoft.com/en-us/sql/t-sql/functions/db-id-transact-sql?view=sql-server-2017).

I specifically encountered this when using a readable secondary as part of an availability group.

Set up a database as part of an Azure Availability group and install `sp_blitzquerystore` then try to run it on the secondary. 

This will likely fail with the error that query store is not enabled. Technically you could argue that this would be accurate, as it's a secondary and QS isn't enabled for the secondary itself, however, if you query sys.databases, as far as it's concerned, QS is enabled, meaning that the error is being returned erroneously. I would also expect that for most people using this it would be more desirable to be able to run the script on the secondary to query the results from the primary, rather than just not be able to run it at all on the secondary.

The reason for the failure can be confirmed by looking at the database IDs returned using `DB_ID` and comparing against those in `sys.databases` - the ID in `sys.databases` won't match, therefore the name being used for the comparison is incorrect.

The database names/IDs being compared should match up. Or in my case, I was hoping to be able to run the script on the secondary server to look at the stats from the primary copy of the database which has query store enabled.

Azure SQL DB.

paulmchenry:
I've added in a new parameter which sets the database ID to what it is in `sys.databases`, rather than using the result from `DB_ID`, as this can differ in Azure SQL DB and the docs advise using `sys.databases` for comparisons. (Documented [here](https://docs.microsoft.com/en-us/sql/t-sql/functions/db-id-transact-sql?view=sql-server-2017))

Comparisons for the check on whether or not query store is enabled are then done just using the DatabaseName and DatabaseID variables and comparing them with the values in `sys.databases`.

There are possibly other places those functions will be causing problems in Azure, but having tested this change I at least managed to get this working correctly and the script will now run on a readable secondary and return the query store results from the primary.

Closes #2136

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -200,7 +200,7 @@ IF  (	SELECT COUNT(*)
 		PRINT @msg;
 		RETURN;
 	END;
-
+									
 /*Making sure your databases are using QDS.*/
 RAISERROR('Checking database validity', 0, 1) WITH NOWAIT;
 
@@ -241,14 +241,14 @@ END;
 /*Does it have Query Store enabled?*/
 RAISERROR('Making sure [%s] has Query Store enabled', 0, 1, @DatabaseName) WITH NOWAIT;
 IF 	
-	((DB_ID(@DatabaseName)) IS NOT NULL AND @DatabaseName <> '')
-AND		
-	(   SELECT DB_NAME(d.database_id)
-		FROM sys.databases AS d
-		WHERE d.is_query_store_on = 1
-		AND d.user_access_desc='MULTI_USER'
-		AND d.state_desc = 'ONLINE'
-		AND DB_NAME(d.database_id) = @DatabaseName ) IS NULL
+
+	( SELECT [d].[name]
+		FROM [sys].[databases] AS d
+		WHERE [d].[is_query_store_on] = 1
+		AND [d].[user_access_desc]='MULTI_USER'
+		AND [d].[state_desc] = 'ONLINE'
+		AND [d].[database_id] = (SELECT database_id FROM sys.databases WHERE name = @DatabaseName)
+	) IS NULL
 BEGIN
 	RAISERROR('The @DatabaseName you specified ([%s]) does not have the Query Store enabled. Please check the name or settings, and try again.', 0, 1, @DatabaseName) WITH	NOWAIT;
 	RETURN;


sp_BlitzQueryStore.sql
```diff
@@ -32,6 +32,7 @@ GO
 ALTER PROCEDURE dbo.sp_BlitzQueryStore
     @Help BIT = 0,
     @DatabaseName NVARCHAR(128) = NULL ,
+		@DatabaseID INT = NULL,
```
BrentOzar: I'm a little confused, so forgive me if I'm missing something obvious. Why is this an input parameter? What's the scenario where the user knows the database ID, but the system doesn't? (Meaning, why couldn't it just be an internal variable, not exposed to the user?)

sp_BlitzQueryStore.sql
```diff
@@ -205,7 +206,14 @@ IF  (	SELECT COUNT(*)
 RAISERROR('Checking database validity', 0, 1) WITH NOWAIT;
 
 IF (@is_azure_db = 1)
+BEGIN
 	SET @DatabaseName = DB_NAME();
+	SET @DatabaseID = ( 
+										 SELECT database_id
+										 FROM sys.databases
+										 WHERE [name] = @DatabaseName
+										);
```
BrentOzar: Why would this only be done for Azure? I thought your issue was about running it on the secondary where the database ID would be different - why not just always set this?

sp_BlitzQueryStore.sql
```diff
@@ -241,14 +249,21 @@ END;
...
-		WHERE d.is_query_store_on = 1
-		AND d.user_access_desc='MULTI_USER'
-		AND d.state_desc = 'ONLINE'
-		AND DB_NAME(d.database_id) = @DatabaseName ) IS NULL
+	( SELECT [d].[name]
+		FROM [sys].[databases] AS d
+		WHERE [d].[is_query_store_on] = 1
+		AND [d].[user_access_desc]='MULTI_USER'
+		AND [d].[state_desc] = 'ONLINE'
+		AND [d].[name] = @DatabaseName
```
BrentOzar: Why do we need to check both the d.name AND d.database_id? If we're setting a variable with the database_id, why not just use the database_id in the filter?

sp_BlitzQueryStore.sql
paulmchenry: There's no reason for this at all. I just searched on the DatabaseName parameter because the 2 were connected in my head and I put it there without thinking. It shouldn't be though, so I'll fix this! 🤦‍♂

sp_BlitzQueryStore.sql
paulmchenry: I don't think there's any reason necessarily to only do it for Azure, but I was just trying to make as little change as possible but fix the issue, so my thinking here was why change it for SQL Server when I only think of the problem as being specific to Azure SQL DB. 

If you think I should just have it set everywhere for consistency though I'm happy to update that.

sp_BlitzQueryStore.sql
paulmchenry: You're right that this is probably unnecessary, and the reason I added it in was that I wanted the code to be as close to what it was originally, and since originally it was passing in the ID value to get the name I thought of this as being the closest thing. 

Plus I was thinking it just serves as a validation that the ID and name match up. I don't know that there's a situation where they wouldn't, but I didn't see any harm in it. I'll alter it to just use the ID though.

sp_BlitzQueryStore.sql
```diff
@@ -238,17 +239,28 @@ BEGIN
...
+
 /*Does it have Query Store enabled?*/
 RAISERROR('Making sure [%s] has Query Store enabled', 0, 1, @DatabaseName) WITH NOWAIT;
 IF 	
-	((DB_ID(@DatabaseName)) IS NOT NULL AND @DatabaseName <> '')
+	(
+		(SELECT database_id
+		 FROM sys.databases
+		 WHERE name = @DatabaseName
+		) IS NOT NULL AND @DatabaseName <> ''
```
BrentOzar: Dumb question - do we still need this query at all if we tweak line 262 to just be a subquery?

sp_BlitzQueryStore.sql
```diff
@@ -238,17 +239,28 @@ BEGIN
...
-		WHERE d.is_query_store_on = 1
-		AND d.user_access_desc='MULTI_USER'
-		AND d.state_desc = 'ONLINE'
-		AND DB_NAME(d.database_id) = @DatabaseName ) IS NULL
+	( SELECT [d].[name]
+		FROM [sys].[databases] AS d
+		WHERE [d].[is_query_store_on] = 1
+		AND [d].[user_access_desc]='MULTI_USER'
+		AND [d].[state_desc] = 'ONLINE'
+		AND [d].[database_id] = @DatabaseID
```
BrentOzar: Dumb question - why do we need a variable at all? Why not just say d.database_id = (SELECT database_id FROM sys.databases WHERE name = @DatabaseName?)

paulmchenry: We don't. Updated.

sp_BlitzQueryStore.sql
paulmchenry: I guess not the `database_id` part so much, but there's still the check for the empty string. Should I remove that?

BrentOzar: That would kinda get fixed by the line 256 check for @DatabaseName, I guess, no need for that either, hahaha. You're kicking butt!

paulmchenry: Ok, updated. That turned into a much simpler change than it seemed originally 😄

paulmchenry:
This probably isn't the ideal place to mention this, but I don't know if you have come across this problem: http://blog.tdwright.co.uk/2018/09/15/making-a-hash-of-things-using-s-in-git-commit-messages/
TLDR; due to GitHub using # to track issues and it also being the character used for comments, it causes problems when doing an interactive rebase. I worked around it using the method in the article and changing the comment character, but it might be worth noting this somewhere, e.g. the contribution guide.

paulmchenry:
No problem, thanks Brent!

Issue #2149, opened 2019-10-17
sp_blitz reports unused heaps as used


rsocol:
sp_blitz version 7.8 (20190922) reports heaps as being used although they have zero user_seeks, user_scans, user_lookups and user_updates. I would expect such heaps to be reported as not used.

BrentOzar:
Looking at the code you sent in - if you head heaps with 0's for user_seeks, user_scans, user_lookups and user_updates, then your fix wouldn't change anything. The NULLIF wouldn't be triggered if the values were zero. Can you tell me more about the user_seeks, user_scans, user_lookups and user_updates that you're seeing in production that led to the problem? Thanks!

rsocol:
I have some rows that have 0, 0, 0, 0 and some rows that have NULL, NULL, NULL, NULL. 
The current code considers that the table is unused only if all values are NULL.
My code changes the 0-s to NULL-s (because I'm using NULLIF, not ISNULL).

BrentOzar:
Ahhhh, I see what you're doing! Great catch. I've merged it into the dev branch, it'll be in the October release, and I'll credit you in the release notes. Thanks!

PR #2151, opened 2019-10-17
corrected fix for issue #2149

rsocol:
Unused heaps were reported as used.
See https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2149

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -6161,7 +6161,7 @@ IF @ProductVersionMajor >= 10
 		  INNER JOIN [?].sys.partitions p ON i.object_id = p.object_id AND i.index_id = p.index_id
 		  INNER JOIN sys.databases sd ON sd.name = N''?''
 		  LEFT OUTER JOIN [?].sys.dm_db_index_usage_stats ius ON i.object_id = ius.object_id AND i.index_id = ius.index_id AND ius.database_id = sd.database_id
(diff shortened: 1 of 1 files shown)


sp_Blitz.sql
```diff
@@ -6101,7 +6101,7 @@ IF @ProductVersionMajor >= 10
 		  INNER JOIN [?].sys.partitions p ON i.object_id = p.object_id AND i.index_id = p.index_id
 		  INNER JOIN sys.databases sd ON sd.name = N''?''
 		  LEFT OUTER JOIN [?].sys.dm_db_index_usage_stats ius ON i.object_id = ius.object_id AND i.index_id = ius.index_id AND ius.database_id = sd.database_id
-		  WHERE i.type_desc = ''HEAP'' AND COALESCE(ius.user_seeks, ius.user_scans, ius.user_lookups, ius.user_updates) IS NOT NULL
+		  WHERE i.type_desc = ''HEAP'' AND COALESCE(NULLIF(ius.user_seeks,0), NULLIF(ius.user_scans,0), NULLIF(ius.user_lookups,0), NULLIF(ius.user_updates,0)) IS NOT NULL
```
BrentOzar: Hmm, I'm a little confused about this. Coalesce takes the first non-null value. If user_seeks was null, the NULLIF would immediately change it to zero, and none of the rest of the values would be reviewed.

Can you tell me a little more about the data you're seeing that led you to implement this fix, and I might be able to understand it better? Thanks!

Issue #2180, opened 2019-10-29
sp_DatabaseRestore : Expand functionality to accept multiple backup paths


CubsRep:
**Is your feature request related to a problem? Please describe.**
Server using multiple drives or Mount Points for backups are being stripped by default, if using Ola Hallengren it matches to the count of drives specified or @NumberOfFiles parameter value. sp_DatabaseRestore supports stripped files if all reside in single location or path notif there are separated drives or mount points 
Eg:
`EXEC dbo.sp_DatabaseRestore 
         @Database     = N'SampleTest'
         ,@BackupPathFull    = N'Y:\BACKUP\FULL\SampleTest,Y:\BACKUP1\FULL\SampleTest'
         ,@BackupPathDiff    = N'Y:\BACKUP\DIFF\SampleTest,Y:\BACKUP1\DIFF\SampleTest'
         ,@BackupPathLog    = N'Y:\BACKUP\LOG\SampleTest,Y:\BACKUP1\LOG\SampleTest'
         ,@RestoreDiff    = 1
         ,@ContinueLogs    = 0
`
**Describe the solution you'd like**
Split Full/Diff/Log Backup paths (only in CSV format). Iterate file loading into stage table thru loop. For `RESTORE HEADERONLY` and `RESTORE FILELISTONLY` statements pick one of the stripped back file as metadata will be similar across set.  FULL and DIFF will use the most recent backup set available that itself serve as basepoint for LOG restores by comparing LSNs which the current proc does

**Describe alternatives you've considered**
Manually copying .bak files into single location and executing the proc

**Are you ready to build the code for the feature?**
i'd love to contribute. modified the proc to add this feature and it's in Testing phase

BrentOzar:
OK, cool, sounds good! Go ahead and hit the Contributing Guide (it's in the Documentation folder in the repo.) A couple of things people have been running into lately: make sure it works on a case-sensitive server, and make sure it works going back to SQL Server 2012. Thanks!

CubsRep:
Opened new PR  #2241 for code review

CubsRep:
New PR #2277 opened resolving rebase conflicts

BrentOzar:
We didn't get anyone testing this despite waiting a couple of months, so I've gone ahead and merged it in. It'll be in the July release along with credit to you in the release notes. Thanks!

PR #2188, opened 2019-11-08
Restore from Multiple backup paths

CubsRep:
- #2180 
- Added `OFFILNE` option to `@ExistingDBAction` parameter. Flag Id = 4


CubsRep:
@BrentOzar - PR opened for review

BrentOzar:
OK, thanks! I don't use sp_DatabaseRestore personally, so can you maybe hop into the Slack room and get other folks to do testing on it as well? There are other folks who use it who may be able to help. Thanks!

reharmsen:
> 
> 
> OK, thanks! I don't use sp_DatabaseRestore personally, so can you maybe hop into the Slack room and get other folks to do testing on it as well? There are other folks who use it who may be able to help. Thanks!

I like the idea of being able to restore striped backups.
Would love to try this in our environment to see whether I would be able to speedup backup and restore procedures.
Unfortunately I am not able to set this up, in the near future... might be February (or later) before I have the ability and time

CubsRep:
I have had sync issues with the current PR.. so opened new #2241 for review and testing purposes

PR #2189, opened 2019-11-12
Remove Square Braces

paschott:
Removed square braces from FRK-specific objects to make it easier to convert this into a series of temporary procs if we can't create them on a target server.  Can now replace "dbo.sp_Blitz" with "##sp_Blitz" and run.

### DIFF
--- a/Install-All-Scripts.sql
+++ b/Install-All-Scripts.sql
@@ -37045,7 +37045,7 @@ GO
 IF OBJECT_ID('dbo.sp_DatabaseRestore') IS NULL
 	EXEC ('CREATE PROCEDURE dbo.sp_DatabaseRestore AS RETURN 0;');
 GO
-ALTER PROCEDURE [dbo].[sp_DatabaseRestore]
+ALTER PROCEDURE dbo.sp_DatabaseRestore
     @Database NVARCHAR(128) = NULL, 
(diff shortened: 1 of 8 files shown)


BrentOzar:
Howdy, and thanks for the pull request. Before you get started working, though, take a few minutes to read the contributing guide:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/master/CONTRIBUTING.md

It explains how to jump in and code. Before you change anything, can you open an issue and discuss the change you're considering making, and get feedback from other folks on the proposed changes?

In this case, the Install% scripts are generated automatically during our build process, so you don't have to edit those.

Also, I'm confused - can't you replace [sp_Blitz] with [##sp_Blitz]? I'm not sure what the point of this change is.

paschott:
I could have sworn that the dbo.sp.... syntax gave me trouble last time when I tried a global replace but will try again. If that doesn't work I'll file an issue and adjust the pull request accordingly. If it does work, I will drop this one.

Sorry about missing the contributor guidelines.

BrentOzar:
OK, cool. I'm going to close this for now - I did a create proc [dbo].[##test] and it worked fine, so I'm guessing it had something to do with your specific testing scenarios. Let's revisit that in the issue once you track down what the issue is. Thanks!

PR #2271, opened 2020-01-28
#2269 sp_DatabaseRestore Remove Ola Dependency

### LINKED ISSUE #2269: Missing @DatabaseContext parameter in sp_DatabaseRestore
The sp_DatabaseRestore stored procedure uses the CommandExecute procedure from the Ola Hallengren solution, but it is no more compatible with the latest version of the Ola Hallengren solution. The latest version of the CommandExecute procedure (2020-01-14) requires new @DatabaseContext input parameter, but this parameter is missing in the current version of the sp_DatabaseRestore procedure and so the procedure sp_DatabaseRestore fails.

I hotfixed this in my local installations by adding this parameter into all calls of CommandExecute procedure inside sp_DatabaseRestore with the constant value of @DatabaseContext = 'master' which fixes the issue for me.

estraffen:
Dependencies for dbo.CommandExecute and dbo.DatabaseIntegrityCheck have been removed. Calls to these stored procedures have been replaced with sp_executesql and DBCC CHECKDB respectively. Closes #2269

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -682,7 +682,7 @@ BEGIN
 			        PRINT @sql;
 		        END;
 		        IF @Debug IN (0, 1) AND @Execute = 'Y'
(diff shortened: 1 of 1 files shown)


BrentOzar:
Looks good! I merged this w/o testing it, which is unusual, but I don't use this proc a lot, and it looks fine at first glance. Nice work! It's in the dev branch now, and it'll be in the February release.

PR #2272, opened 2020-01-29
Update sp_ineachdb.sql

AaronBertrand:
Add a parameter `@exclude_pattern` to allow you to exclude databases based on a pattern.


BrentOzar:
Thanks sir! Will be in the Feb release and I'll credit you in the release notes.

PR #2273, opened 2020-02-02
Renaming Documentation to publish it as a wiki

PierreLetter:
Publishing to a Wiki in Azure DevOps Server doesn't allow for spaces in file names.
We are deploying the Kit as default on all our servers and want to have the doc handy for people to read in our DBTools repo wiki which this solution is now part of (thanks a lot!)
Thought sharing this light change would probably benefit others.


BrentOzar:
Works for me! Thanks, merged into dev, and will be in the February release.

PR #2278, opened 2020-02-07
Issue 2234

erikdarlingdata:
Checks to make sure we don't accuse parallel plans of nearly being parallel.


BrentOzar:
Good work! Merged into dev, will be in the February release.

PR #2283, opened 2020-02-11
Issue 2219

### LINKED ISSUE #2219: sp_BlitzLock: add parameter for outputting victims only
In the Consultant Toolkit, we're showing all of the deadlock participants. Since most of the troubleshooting is done via the deadlock graph, add an output parameter that restricts the top result set to victims only.

Yep!

### LINKED ISSUE #2282: sp_BlitzLock: Object names not populated correctly
**Version of the script**
SELECT @Version = '2.92', @VersionDate = '20200123';

Object names don't populate in the final output

Just run the script, heh

Object names should populate in the final result set

All, it used to work. This is a bug introduced in https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/commit/f00cd13cbd245ebf710e820eb5495c947af637f8 where event_date is converted to a datetime.

Which messes up this:

image

Because these don't match:

image

I'm fixing this in another PR to add Victim Only output.

erikdarlingdata:
Closes #2219 
Closes #2282

This PR:
 * Adds a filter to output victims only
 * Adds some missing RECOMPILE hints
 * Adds an explicit BEGIN/END around the stats update
 * Fixes data type issue when converting to... whatever that thing is
 * Adds better support for identifying when heaps are deadlocked (to make it more like BlitzIndex)
 * Fixes an issue where object names stopped populating because of weird data types. It's back to functional now. I'm not psyched that the most precise matching data type on those columns is `DATE`, but it's not a harmful situation. I'll have to dig more on what exactly isn't matching, and maybe add an `event_date_match` column to the table that doesn't do the whole big convert.


BrentOzar:
Awesome! Thanks, sir. Merged, will be in the February release.

PR #2289, opened 2020-02-16
Issue 2280: Skip 0 read, 0 write indexes

### LINKED ISSUE #2280: sp_BlitzIndex: 0 read, 0 write indexes
I'm thinking this is a good fit for mode 0 (but it might also help with the 10k line cutoff in mode 4, too 👀👀👀). Filter out indexes that have no reads or writes. It'll make the output more succinct, and help people focus on things that will have an impact.

Great taste, less filling

Worse taste, more filling

Yeah, if the nice people who read these things think it'd be useful.

erikdarlingdata:
This pull request SHALL HERETOFORE:
 * Add a parameter to disinclude indexes with no reads OR writes, meaning they are totally inactive
 * De-clutter the select list a little bit so it matches the rest of the statement

Totally inactive indexes will still be listed when:
 * You override the default parameter value
 * You're in mode 1, 2, or 3
 * The `@TableName` parameter is not null

RE The last one: I figure if you're zooming in on a table, you want everything, and who knows? Maybe there's a filtered index on there that hasn't been touched at all?

Closes #2280

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1210,15 +1210,15 @@ BEGIN TRY
                              ELSE N''''
                         END AS filter_definition' ELSE N''''' AS filter_definition' END + N'
                         , ISNULL(us.user_seeks, 0),
-						ISNULL(us.user_scans, 0),
+                        ISNULL(us.user_scans, 0),
                         ISNULL(us.user_lookups, 0),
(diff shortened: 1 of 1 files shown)


BrentOzar:
The motion is seconded and carried. The Senate agrees. This pull request has been paid for by the Campaign for Clearer Index Results.

PR #2291, opened 2020-02-16
Issue 2290: Select with writes bug

### LINKED ISSUE #2290: sp_BlitzCache: Selects w/ Writes may fire incorrectly for stored procedures
**Version of the script**
All of them since I added this check :(

This check matches on the QueryHash, but the way the XML is parsed into temp tables, all plans are associated with all query hashes. So for a proc like this, you'll see that a select did writes even if it didn't:

```
CREATE PROC dbo.Wronged
AS
BEGIN
SET NOCOUNT, XACT_ABORT ON;
    SELECT *
    INTO #Badges
    FROM dbo.Badges AS b
    WHERE b.Name = N'Popular Question'
    
    
    SELECT TOP 1 * 
    FROM dbo.Badges AS b 
    WHERE b.Name = N'Popular Question'
END
GO 

EXEC dbo.Wronged

EXEC sp_BlitzCache @StoredProcName = 'Wronged', @ExpertMode = 1, @Debug = 1;
```

See above

Don't be goofy

All of them

erikdarlingdata:
Gets the query hash from the plan XML, also changes the join semantics to make sure the rows matched have writes associated with them.

Closes #2290

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -3382,7 +3382,10 @@ RAISERROR('Checking for selects that cause non-spill and index spool writes', 0,
 WITH XMLNAMESPACES (
     'http://schemas.microsoft.com/sqlserver/2004/07/showplan' AS p )
 , selects
-AS ( SELECT s.QueryHash
+AS ( SELECT CONVERT(BINARY(8), 
+                RIGHT('0000000000000000' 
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks sir! Merged into dev, will be in the February release.

Issue #2286, opened 2020-02-12
sp_BlitzLock - Parallel Deadlocks not showing in first result set


Adedba:
**Version of the script**
SELECT @Version = '2.92', @VersionDate = '20200123';

**What is the current behavior?**
First result set not populated when the only deadlock present is a parallel deadlock

**If the current behavior is a bug, please provide the steps to reproduce.**
Not sure how to reproduce a parallel deadlock, fortunately the system I am working on gets them all the time it seems. 

Run sp_BlitzLock with default parameters

First result set is not populated but the second resultset is populated with findings.

**What is the expected behavior?**

Expect both result sets to be available with the first showing the deadlock graph xml

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**

Only tested with SQL 2012 SP4

Please don't shoot me, I am not entirely sure if this is a bug or not  :P

Here is the output sp_BlitzLock @Debug = 1;

Result set 1 empty
Result set 2 has findings
#deadlock_process temp table has deadlock information
#deadlock_resource_parallel temp table has information too

![BlitzLock1](https://user-images.githubusercontent.com/33764798/74342622-afa32700-4da1-11ea-9ff9-d525a177dad2.png)
![BlitzLock2](https://user-images.githubusercontent.com/33764798/74342623-b03bbd80-4da1-11ea-80ee-92395ce9938b.png)

Looking at the code I can see: 

```
CROSS APPLY (SELECT TOP 1 * FROM  #deadlock_resource_parallel AS drp WHERE drp.owner_id = dp.id AND drp.wait_type = 'e_waitPipeNewRow' ORDER BY drp.event_date) AS cao
CROSS APPLY (SELECT TOP 1 * FROM  #deadlock_resource_parallel AS drp WHERE drp.owner_id = dp.id AND drp.wait_type = 'e_waitPipeGetRow' ORDER BY drp.event_date) AS caw
```

If I change the CROSS APPLY to OUTER APPLY for both it returns a result with owner_mode 'e_waitPipeNewRow'

From what I can see I think it might be where only one of the wait types ('e_waitPipeNewRow' or 'e_waitPipeGetRow') exist but not both.

Adedba:
Forgot to add the deadlock xml, will grab that tomorrow!

Adedba:
Cannot grab the whole thing as there is too much vendor data in there but here is the resource list: 

`<resource-list>
[...]

PR #2305, opened 2020-02-17
Show all parallel deadlocks

Adedba:
#2286 - Show all parallel deadlocks


BrentOzar:
Looks good, thanks! It'll be in the March release and I'll credit you in the release notes.

PR #2313, opened 2020-02-22
Invalid option for CheckID 1 of sp_Blitz

srutzky:
Changed `SERVERPROPERTY('EngineName')` to be `SERVERPROPERTY('EngineEdition')` on Line 876. As far as I can tell, "EngineName" is an invalid option that always returns `NULL`. This causes the `IF` condition to always use the `ELSE` portion, which should only be used for "Azure Managed Instance"s. According to the documentation for the "[SERVERPROPERTY](https://docs.microsoft.com/en-us/sql/t-sql/functions/serverproperty-transact-sql)" function, it's the "EngineEdition" option that returns "8" for "Azure Managed Instance". Changing this on my system gets this to work as expected.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -873,7 +873,7 @@ AS
 
 						IF @Debug IN (1, 2) RAISERROR('Running CheckId [%d].', 0, 1, 1) WITH NOWAIT;
 
-                        IF SERVERPROPERTY('EngineName') <> 8 /* Azure Managed Instances need a special query */
+                        IF SERVERPROPERTY('EngineEdition') <> 8 /* Azure Managed Instances need a special query */
                             BEGIN
 						    INSERT  INTO #BlitzResults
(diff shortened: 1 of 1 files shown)


BrentOzar:
Great catch, thanks sir! Merged, will be in the March release. Thanks!

BrentOzar:
(Also, should note - looks like the only drawback would have been that we'd have included backups that ran on other servers, which shouldn't have had a big effect for folks. Worst case would have been giving them some false positives that their databases were being backed up when they weren't.)

srutzky:
Just FYI: this bug was introduced in PR #1485 via the following commit (the exact line should be highlighted):

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/commit/5c3588eaaac39f43d0be8fddfe58cfa2831c6f01#diff-45927e55fc4332260bd0ae9708e4ef62R696

Issue #2134, opened 2019-09-27
sp_Blitz and sp_BlitzFirst: add warning about high USERSTORE_TOKENPERM cache size


erikdarlingdata:
Back when I wrote this check in, I apparently tacked it onto the warning about the plan cache being erased recently. I feel like that could easily get missed since the message for it is already pretty long.

If you look [here](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_Blitz.sql#L3459), the code is already written to get the size. 

The question is: is this important enough to be its own check?

If so, is 2 GB the right size to warn at?

RobAshton-IM:
"... is 2 GB the right size to warn at?"

Tough one. If it's 2 GB of 4 GB in total, then I'm interested for sure. If it's 2 GB of 256 GB, then probably not so much.

Is it worth pulling in from SQL OS how much memory is available for all memory caches to fight over to see if USERSTORE_TOKENPERM is using a lot more than its fair share of the available space?

[Looking at this post from Bulent Ozkir](https://blogs.technet.microsoft.com/bulentozkir/2014/01/09/tokenandpermuserstore-related-information-on-sql-server-2012/), and [this thread on SQL Server Central](https://www.sqlservercentral.com/forums/topic/userstore_tokenperm-is-huge) gives me a few ideas of variations on this theme to point out different reasons why USERSTORE_TOKENPERM has blown out of proportion. Perhaps there's a new group of checks here?

erikdarlingdata:
@RobAshton-IM I don't think I'm going to work on this one. If you want to, feel free and let me know. Otherwise, I'll probably close this out.

BrentOzar:
Yeah, I'll tackle it. I've got a client fighting this issue, and I'm pretty sure there's a bug in SQL Server around it. They have a Microsoft support call open, and we're still trying to track it down.

[...]

PR #2319, opened 2020-03-02
Issue 2134

erikdarlingdata:
Add a check for USERSTORE_TOKENPERM if >30% of the plan cache is an hour or less old.
Compare the cache size to the size of the buffer pool.
If it's >10%, add a warning to the end results.


BrentOzar:
Working on #2134

Issue #1947, opened 2019-02-07
sp_BlitzIndex: Stats info running from wrong database


mabrahamsen:
**Version of the script**
sp_BlitzIndex(TM) v7.2 - January 28, 2019: Database [XYZ] as of 2019-02-07 13:15

**What is the current behavior?**
When you execute a BlitzIndex specifying a database name, and a index name (to get the stats view mode), it will pull stats / histogram from an index in the current database (aka the active database in your SSMS query view) - not the database specified.

**If the current behavior is a bug, please provide the steps to reproduce.**
Run BlitzIndex for a table, but make sure your current active database is not the same as the database specified.

EXEC sp_BlitzIndex @DatabaseName = 'XYZ', @SchemaName = 'dbo', @TableName = 'XYZ';

It will pull stats / histogram from an index, in the current database - assuming it finds it. I guess it will load by some identifiers that may or may not be in the active database as well. It will find a match in some active databases, but not in others.

**What is the expected behavior?**
Load stats from the specified database and index, not the current active database.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**

Only tested on SQL 2017. I believe this is a new feature in a very recent update.

mabrahamsen:
Seems there is no way to set the database for:
--
CROSS APPLY sys.dm_db_stats_properties(s.object_id, s.stats_id) AS props  
CROSS APPLY sys.dm_db_stats_histogram(s.[object_id], s.stats_id) AS hist
--
So they would need to be executed in the correct database context. The system tables could be easily fixed by specifying database name. I couldn't get that to work correctly with the functions.

I am not sure what infrastructure sp_blitzindex would have to execute a SQL string in a specific database context - I couldn't find any by doing a very quick code scan.

BrentOzar:
Great catch, and thanks for the work reproducing it!

Yeah, looks like a bug in sys.dm_db_stats_histogram and sys.dm_db_stats_properties.

To work around it, I did two things:

1. Put a check in sp_BlitzIndex to only query statistics if we're checking the same database we're in (just to prevent the false alarm of stats), and
2. Filed a bug in SQL Server: https://feedback.azure.com/forums/908035-sql-server/suggestions/36803269-sys-dm-db-stats-histogram-and-sys-dm-db-stats-prop

PR #2341, opened 2020-04-16
#1947 sp_BlitzIndex stats for current db

AndyPLsql:
For solve #1947 i switch database context using "USE" operator. It's works on SQL Server 2012.


BrentOzar:
The problem is that while the USE works, the result set is still empty.

Here's how to reproduce it:

```
CREATE DATABASE StatsTest;
GO
USE StatsTest;
GO
CREATE TABLE dbo.Folks (PersonName VARCHAR(40));
INSERT INTO dbo.Folks VALUES ('Brent Ozar');
CREATE INDEX PersonName ON dbo.Folks(PersonName);
GO
EXEC sp_BlitzIndex @DatabaseName = 'StatsTest', @TableName = 'Folks';
GO
/* Read the result sets, see the stats on PersonName */

USE master;
GO
EXEC sp_BlitzIndex @DatabaseName = 'StatsTest', @TableName = 'Folks';
GO
/* No stats on PersonName even with your fix */
```

Thanks for taking a shot at it, though.

PR #2354, opened 2020-04-27
fix a typo ("undre") + update year in 2 spots

amtwo:
In the license, fix "undre" because it was bothing tf out of me.

Also noticed two spots that list copyright year as 2019, even though main copyright statement is 2020. Fix that too.


BrentOzar:
Got it, thanks sir! Merged into dev, will be in the April release and will credit you in the release notes. Thanks!

Issue #2359, opened 2020-05-05
sp_BlitzCache: Should we warn about lots of single use plans?


erikdarlingdata:
**Is your feature request related to a problem? Please describe.**
We warn about percentages of plans created in the past X hours, is this also a valuable metric to also warn about? Perhaps for something additional to look at alongside the token perm warning?

**Describe the solution you'd like**
I think there's already some logic in Blitz to warn about single use plans. It might be good to surface that here.

**Describe alternatives you've considered**
Running sp_Blitz every 10 minutes.

**Are you ready to build the code for the feature?**
Yeah, it should be easy to drop in.

erikdarlingdata:
I started looking at how I'd wanna do this today. Sticking this query here so I don't lose it, and in case there's any feedback to incorporate. I haven't figured out thresholds to warn at yet, so if you have OPINIONS, feel free to prolapse them here.

```
WITH many_plans AS 
(
    SELECT SUM(x.duplicate_plan_handles) AS duplicate_plan_handles
    FROM (
        SELECT COUNT_BIG(DISTINCT plan_handle) AS duplicate_plan_handles
        FROM sys.dm_exec_query_stats qs
            CROSS APPLY sys.dm_exec_plan_attributes(qs.plan_handle) pa
        WHERE pa.attribute = N'dbid'
        GROUP BY qs.query_hash, pa.value
        HAVING COUNT_BIG(DISTINCT plan_handle) > 5
    ) AS x
),
     single_use_plans AS 
(
    SELECT COUNT_BIG(*) AS single_use_plan_count
    FROM sys.dm_exec_cached_plans AS cp
    WHERE cp.usecounts = 1
    AND   cp.objtype = N'Adhoc'
    AND   EXISTS ( SELECT 1/0
                   FROM sys.configurations AS c
                   WHERE c.name = N'optimize for ad hoc workloads'
                   AND   c.value_in_use = 0 )
    HAVING COUNT_BIG(*) > 1
),
     total_plans AS 
(
    SELECT COUNT_BIG(*) AS total_plans
    FROM sys.dm_exec_cached_plans AS decp
)
SELECT m.duplicate_plan_handles, 
       CONVERT(DECIMAL(3,2), m.duplicate_plan_handles / (1. * NULLIF(t.total_plans, 0))) * 100. AS percent_duplicate,
       s.single_use_plan_count, 
[...]

PR #2381, opened 2020-05-19
Fix sp_BlitzLock heap identification

### LINKED ISSUE #2360: BlitzLock shows heap in a table with clustered index
**Version of the script**
Version = '2.94', @VersionDate = '20200324'

I'm watching a lock alert for a heap when table has a clustered index

I've created a clustered index with allow pages / rows off, what generated a lock escalation at index level. However, the SP showed that there is a lock at the table's heap!

Lock escalation happening at table's clustered index

Microsoft SQL Server 2016 (SP2-CU10) (KB4524334) - 13.0.5492.2 (X64)   Oct  4 2019 19:14:08   Copyright (c) Microsoft Corporation  Enterprise Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor) 
lock
Tested only in this server with this script version

erikdarlingdata:
Closes #2360 

Fixes tables with clustered indexes being identified as heaps when object level locks are taken.

Also adds more clarity to CheckId 2 about what type of object was involved in the deadlock.


erikdarlingdata:
Okay, I beefed this a little bit. It has the changes for #2360 and also closes #2359. My clever branching scheme didn't work 😫

This update adds two plan cache checks:
- Single use plans
- Queries with multiple plans (> 5)

As a side effect, I had to expand the finding column in ##BlitzCacheResults a little bit. It also adds a new table called #plan_usage into the mix to hold results.

As submitted, these will fire when the plan cache is more than 5% inhabited by either species. Feel free to change this. I'm terrible at deciding baseline numbers.

BrentOzar:
Looks good! I like it. Thanks sir!

PR #2385, opened 2020-05-22
Add LastCompletionTime

### LINKED ISSUE #2377: sp_blitzfirst/sp_blitzcache not showing queries with duration that exceeds @minutesBack setting
Version = '7.95', VersionDate = '20200506'

We have sp_blitzfirst setup via SQL Agent to run every 15 minutes. 
While testing query performance for a really slow query that would run for 15-20 minutes, we went and checked the sp_blitzcache table to get the query plan and performance statistics on it and it wasn't in there, but other, less long running and less expensive queries were there.
The problem appears to be that sp_blitzcache uses @minutestback to look at the last Execution time for the queries in the cache, but the Execution time doesn't get saved until the query completes. So if the query starts before sp_blitzcache is run and completes after,  it doesn't get captured because it's not available in sys.dm_exec_query_stats yet. When it does complete, the next run of sp_blitzcache doesn't include it because the execution time was more then 15 minutes ago, even though it completed within the last 15 minutes. 
Here's an example:
The query stated executing at 5:36 PM
At 5:45 when the sp_Blitzfirst job ran, the query was still running, so it wasn’t captured
The query finished at 5:52 PM
When the sp_Blitzfirst job ran at 6 PM, the query had finished, but the last execution time stored for the query was more than 15 minutes prior (5:36 PM) so sp_blitzcache didn’t capture it.

This seems like a bug to me, or at the least, an unexpected "feature"
I would expect sp_blitzfirst/sp_blitzcache to include queries that completed within the last 15 minutes ( or whatever the @minutesback threshold is). 
I'd especially want it to be capturing queries running longer than 15 minutes as those are probably queries that really need to be looked at, but they are not being collected.
I Took a look at the code, and I think it's a matter of calculating the last completion time and using that to compare against the @minutesback value instead of last execution time
sys.dm_exec_query_stats has the last_elapsed_time in addition to last_execution_time.
So instead of the current filter
    AND x.last_execution_time >= DATEADD(MINUTE, @min_back, GETDATE())
It could be this:
    AND dateadd(ms, (x.last_elapsed_time/1000.), x.last_execution_time) >= DATEADD(MINUTE, @min_back, GETDATE()) 

Along with this change, add a column LastCompletionTime to the ##BlitzCacheProcs so later in the proc, the filter on ##BlitzCacheProcs can be changed to:
	SET @insert_sql += N' AND LastCompletionTime >= DATEADD(MINUTE, @min_back, GETDATE() ) ';
instead of:
	SET @insert_sql += N' AND LastExecutionTime >= DATEADD(MINUTE, @min_back, GETDATE() ) ';

I've already worked up a version of sp_blitzcache with this code change and I've tested it out and it appears to work as intended. I've attached a copy of the modified version for review. 

This was all tested on SQL Server 2017

-Ray Rankins

[sp_blitzcache_fix.txt](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/files/4632022/sp_blitzcache_fix.txt)

erikdarlingdata:
Closes #2377


rrankins:
I tested the changes and they work to return queries based upon their last completion time instead of last execution time, but, if you had already setup a table to capture the results from sp_blitzcache with a prior version, the insert into the table is failing because the table does not have the LastCompletionTime column.  Adding the following at line 6801 before the code that does the insert fixes this error:

/* If the table doesn't have the new LastCompletionTime column, add it. See Github #2385. */
    SET @ObjectFullName = @OutputDatabaseName + N'.' + @OutputSchemaName + N'.' +  @OutputTableName;
    SET @insert_sql = N'IF NOT EXISTS (SELECT * FROM ' + @OutputDatabaseName + N'.sys.all_columns
        WHERE object_id = (OBJECT_ID(''' + @ObjectFullName + N''')) AND name = ''LastCompletionTime'')
        ALTER TABLE ' + @ObjectFullName + N' ADD LastCompletionTime DATETIME;';
    EXEC(@insert_sql);

BrentOzar:
@rrankins we're not doing change detection at this time - that'd be the kind of thing we warn folks about in the release notes and give them an alter table script to run. If you want to automatically manage changes to the BlitzCache table, can you make a separate issue for that and work on the changes as described in the contributing guide? If it's important enough to you to code, then definitely we'd love to have the pull request.

Issue #2420, opened 2020-06-09
sp_BlitzFirst is reporting Queries with "10000x cardinality misestimations" for the sp_BlitzFist query to find cardinality misestimations


gdoddsy:
Not sure if this should be a suggestion or a bug. I won't mind if you just want to close it either.

**Version of the script**
Version: 7.95, Date: 20200506

**What is the current behavior?**
When you run sp_BlitzFirst, one of my top warnings is that I have "Queries with 10000x cardinality misestimations", but the query is the query that finds misestimation-ed queries.

**If the current behavior is a bug, please provide the steps to reproduce.**
First time that I've seen it, all I did was run sp_blitzfirst. Not sure why it came up this time.

**What is the expected behavior?**
Either that the query doesn't misestimate, or the query doesn't report itself as being a problem.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Happening in SQL 2019, I haven't noticed it before.

BrentOzar:
Ouch! I popped open the code and I see what's happening. This is only happening when sp_BlitzFirst is running simultaneously from different sessions. When we populate the @bad_estimate table variable that's used to find these issues, we already have this filter:

`WHERE deqp.session_id <> @@SPID`

So that means you're seeing the sp_BlitzFirst running from another session, like the one you so cleverly set up to run automatically every 15 minutes. I know because I've seen this same thing happen when I've been teaching the Mastering classes.

I would love to exclude that query, but to do it, we (and by we I mean you) would have to examine the contents of the query itself, and at the point where we fill the @bad_estimate table, we haven't joined to the text of the currently running query. If we (and by we I mean you) wanted to do this, we (you get the point) would need to:

* Add a join to something to get the text of the currently running query
* Examine it to see if it's the sp_BlitzFirst query (and to do that you may wanna add an obvious comment to the dynamic SQL that's getting built, like /* From sp_BlitzFirst */, and then look for that)

If you're interested in doing that, holler, but otherwise I'm going to close this because it's more work than I wanna do, because I'm wearing a bathrobe. ;-)

gdoddsy:
![image](https://user-images.githubusercontent.com/25063374/84729985-5b6eb480-afd8-11ea-87ea-8478a04dbf6b.png)

It's not happening at the moment, but I feel like it was picking up the session that I was in. Let me see if I can get it to happen again.

gdoddsy:
[...]

PR #2437, opened 2020-06-17
Removing incorrect logic from @Bad_Estimate check. Issue #2420

urizen-source:
See long explanation in #2420 discussion.

The short version: row_count < estimate_row_count *10000 leads to alot of false positives, as for instance 7 < 5*10000 while 7 vs 5 is a completly okay estimate.

Also, since the DMV is populated on the fly, I do not see fixing the logic to do a cardinality estimate check in this "direction" as an option as row_count will in the beginning of the execution be very low(sometimes 0) which will then always cause an incorrect red flag, just because it has not had time to process yet. 
Row count > estimate_row_count*10000 is always an incorrect warning so that still makes sense.

Removing this section caused me to no longer be able to reproduce the blaming of sp_blitzfirst that the issue is about which I could before(on 2019).

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -1828,8 +1828,6 @@ If one of them is a lead blocker, consider killing that query.'' AS HowToStopit,
                                 deqp.request_id,
                                 CASE WHEN deqp.row_count > ( deqp.estimate_row_count * 10000 )
                                      THEN 1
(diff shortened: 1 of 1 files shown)


BrentOzar:
This fixes #2438. I separated out the lowball estimate issue just in case Greg's self-reporting issue still happens. Great work! It's in the dev branch now and will be in the July release along with credit for you in the release notes.

PR #2459, opened 2020-07-04
2452 Fixes sp_BlitzLock object names with periods

### LINKED ISSUE #2452: sp_BlitzLock Incorrect databasename for Check ID 9 if the database has a dot in the name
**Version of the script**
2.96
2020-06-06 00:00:00.000

If the database name has a DOT in the name eg [Data.Base], only the last part of the name is returned in the database_name.

On [Data.Base] with Deadlock data, run:
`EXEC [dbo].sp_BlitzLock @databasename='Data.Base'`
result:
```
check_id	database_name	object_name	finding_group	finding
9	Base dbo.table More Info - Table	EXEC sp_BlitzIndex @DatabaseName = 'Base', @SchemaName = 'dbo', @TableName = 'table';
```

I expect the full database name.

Tested on 14.0.1000.169

defcon84:
Closes #2452.

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -739,12 +739,14 @@ You need to use an Azure storage account, and the path has to look like this: ht
 
 		/*Get each and every table of all databases*/
 		DECLARE @sysAssObjId AS TABLE (database_id bigint, partition_id bigint, schema_name varchar(255), table_name varchar(255));
-		INSERT into @sysAssObjId EXECUTE sp_ineachdb @COMMAND =
-		  N'SELECT DB_ID() as database_id, p.partition_id, s.name as schema_name, t.name as table_name
+		INSERT into @sysAssObjId EXECUTE sp_MSforeachdb
+		  N'USE [?];
+		    SELECT DB_ID() as database_id, p.partition_id, s.name as schema_name, t.name as table_name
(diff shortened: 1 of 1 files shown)


sp_BlitzLock.sql
```diff
@@ -727,6 +737,15 @@ You need to use an Azure storage account, and the path has to look like this: ht
         AND dp.id = aj.id
 		OPTION ( RECOMPILE );
 
+		/*Get each and every table of all databases*/
+		DECLARE @sysAssObjId AS TABLE (database_id bigint, partition_id bigint, schema_name varchar(255), table_name varchar(255));
+		INSERT into @sysAssObjId EXECUTE sp_ineachdb @COMMAND =
```
BrentOzar, line 742: This is a little bit tricky: sp_ineachdb isn't available in all systems. I know it's in the First Responder Kit, but we can't guarantee that folks have installed ALL of the procs. (For example, the Consultant Toolkit runs each of the First Responder Kit queries as standalone queries, not as stored procedures.) Can you either use sp_MSforeachdb, or else have a graceful fallback if sp_ineachdb isn't installed?

defcon84:
@BrentOzar Sounds the right thing to do, I have changed it.

PR #2464, opened 2020-07-10
Re-adding @ExpertMode check to ensure it returns wait stats

### LINKED ISSUE #2463: sp_BlitzFirst - Wait stats not returned when @ExpertMode = 1
**Version of the script**
Current released version.

When called with @ExpertMode = 1 we do not get wait stats returned any more. This behaviour was altered during the following pull request;

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/commit/625979e76fc6f5d2392b5a22d72fe15e9bd7ebee#diff-552d8a2836417d18a75bac2ab9f8fa82

The modification of line 3964 to remove "@ExpertMode = 0 AND " means we're now hitting that block when @ExpertMode = 1 whereas we want to be hitting the if block after this.

Run the current BlitzFirst with @ExpertMode = 1 and you will not get wait stats returned.

This version (just before that pull request) works fine;

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/db25ecbac09466678f06c072ce7bf8bb47af0111/sp_BlitzFirst.sql

Well, it'd be nice if we could see the wait stats returned please.

RichBenner:
Fixes #2463 

This change returns the functionality for @ExpertMode = 1.

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -3964,7 +3964,7 @@ If one of them is a lead blocker, consider killing that query.'' AS HowToStopit,
                     AND wd2.wait_time_ms-wd1.wait_time_ms > 0
                 ORDER BY [Wait Time (Seconds)] DESC;
         END;
-        ELSE IF @OutputType <> 'NONE' AND @OutputXMLasNVARCHAR = 0 AND @SinceStartup = 0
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks! Good detective work.

PR #2467, opened 2020-07-12
Fix link in Contributing.md

gdoddsy:
Do I need to create an issue in github for this? I was reading some of the documentation and came across this link that doesn't work. Just replaced the %20 with _ and it now works (at least, I think it's going to the right place).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,7 @@ We know that's a pain, but that's the kind of thing we find out in the wild. Of 
 4. Make changes.
 
    Make changes to one or more of the files in the project.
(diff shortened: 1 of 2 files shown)


BrentOzar:
Hi - check the files changed in the pull request. You've also got changes to sp_BlitzIndex in there:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/pull/2467/files

To avoid that, when you work on multiple things at once, you wanna start different branches in your Github repo. Here's how my workflow goes:

* Create an issue
* Create a local branch with that issue number (like 2467_fixing_contributing_link)
* Switch to that branch
* Make my changes
* Do a pull request from that branch

Then, if I wanna work on another issue at the same time, I can step through that same process again, and then switch back and forth between the different branches.

gdoddsy:
Ah - my bad. I'll close this pull request and put this into it's own branch and make a pull request from there.

Issue #2227, opened 2019-12-22
sp_BlitzIndex missing index requests: add datatype to column list


BrentOzar:
**Is your feature request related to a problem? Please describe.**
When I'm looking at the Consultant Toolkit output for missing indexes, I want to see the column datatypes before I take Clippy's word on an index.

For example, when I run this:

```
DropIndexes;
GO
SELECT AboutMe FROM dbo.Users WHERE Location = 'Antarctica';
GO 10
sp_BlitzIndex @TableName = 'Users'
GO
```

I get this in the missing index finding: 

EQUALITY: [Location] INCLUDES: [AboutMe] 

And I'd like it to use the same output style that we use in the Definition column of the first result set. If I create that above index, the Definition column looks like this:

[1 KEY] Location {nvarchar 200} [1 INCLUDE]  AboutMe {nvarchar -1}

So we probably want the missing index output to look like this:

EQUALITY: [Location] {nvarchar 200} INCLUDES: [AboutMe] {nvarchar -1}

**Describe alternatives you've considered**
Outputting all of the column definitions for all of the tables to a separate Consultant Toolkit tab, but that would take up a heck of a lot of space and we probably don't need it, and it'd involve back-and-forth lookups between tabs.

**Are you ready to build the code for the feature?**
Yeah, or maybe see if Erik wants to build it?

BrentOzar:
Other things we may wanna test to make sure it'll work okay:

* SQL Server 2016, but the database is in 2008 compat mode
* Missing index requests with each of these in equality, inequality, and includes:
    * NVARCHAR(100), NVARCHAR(MAX)
    * VARCHAR(100), VARCHAR(MAX)
    * XML, JSON
    * TEXT, NTEXT
    * SMALLINT, INT, BIGINT
    * Dates

BrentOzar:
This isn't quite working yet, but it's close enough for government work and horseshoes, so I'm going to accept it and start working on a tweak to it. Thanks for the pull request. It'll be in the August release with credit to you in the release notes.

gdoddsy:
@BrentOzar - I'm planning to add in the missing commas in the next day or 3, did you want a new issue for that, or can I continue this issue? I don't mind either way, whatever makes your life easier.

BrentOzar:
No worries! Yeah, let's do a new issue for that just because I wanna get the August release out this weekend.

PR #2468, opened 2020-07-12
Add data types to missing index column details

gdoddsy:
Had a go at fixing issue #2227 
It only works for SQL 2016 and newer which isn't ideal, but I think I've got the code right to give the current behavior on older versions. Happy to use something other than string_split and string_agg if you have a standard way in First Responders Kit that makes it compatible with other versions.

I've only tested it out on Stack Overflow database in SQL 2019 in the VM I used for Mastering SQL classes. No idea how it'll work in other environments.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1503,10 +1503,81 @@ BEGIN TRY
 
 
         RAISERROR (N'Inserting data into #MissingIndexes',0,1) WITH NOWAIT;
-        SET @dsql=N'SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks. I'm okay with it only working on newer versions, but there's a problem, though: check the name of the index that gets recommended:

CREATE INDEX [DisplayName__{nvarchar(40)}] ON [StackOverflow2013].[dbo].[Users] ([DisplayName]  {nvarchar(40)})  WITH (FILLFACTOR=100, ONLINE=?, SORT_IN_TEMPDB=?, DATA_COMPRESSION=?);

The datatype is going in there too. I'm going to close this pull request just so you can work on that part if you want. Thanks sir!

PR #2472, opened 2020-07-12
Fix link in Contributing.md

gdoddsy:
Fix link in Contributing.md

Just replaced the %20 with _ and it now works (at least, I think it's going to the right place).


BrentOzar:
Perfect, that works! Thanks sir.

PR #2473, opened 2020-07-12
Issue 2227- sp_BlitzIndex show datatypes on missing indexes

gdoddsy:
Having another go at #2227.
I've added 3 new columns to `#missingindexes` so you can list the columns with data types and columns without.
Also noticed it only works for SQL 2017+ (not 2016 as I previously thought) due to use of `STRING_AGG`.

I wouldn't mind having another crack at getting rid of `STRING_AGG` use `FOR XML PATH`. If it's not too hard, that'll make this compatible with 2016+.

If you have a simple `string_split` replacement then I can add that as well, but everything I've seen says to use a user defined function, which I don't think this script would want to do.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -534,9 +534,12 @@ IF OBJECT_ID('tempdb..#Ignore_Databases') IS NOT NULL
             user_seeks BIGINT NOT NULL,
             user_scans BIGINT NOT NULL,
             unique_compiles BIGINT NULL,
-            equality_columns NVARCHAR(4000), 
+            equality_columns NVARCHAR(4000),
+            equality_columns_with_data_type NVARCHAR(4000),
             inequality_columns NVARCHAR(4000),
+            inequality_columns_with_data_type NVARCHAR(4000),
             included_columns NVARCHAR(4000),
+            included_columns_with_data_type NVARCHAR(4000),
 			is_low BIT,
                 [index_estimated_impact] AS 
                     REPLACE(CONVERT(NVARCHAR(256),CAST(CAST(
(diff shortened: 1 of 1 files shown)


gdoddsy:
I've removed `string_agg` and have set the check for ProductVersion back to SQL 2016.

If you have any suggestions for replacing `string_split`, let me know, otherwise I'll leave this as is.

I've only tested on a handful of databases, all on SQL 2019.

sp_BlitzIndex.sql
```diff
@@ -547,11 +550,11 @@ IF OBJECT_ID('tempdb..#Ignore_Databases') IS NOT NULL
                         + N'%; Avg query cost: '
                         + CAST(avg_total_user_cost AS NVARCHAR(30)),
                 [missing_index_details] AS
-                    CASE WHEN equality_columns IS NOT NULL THEN N'EQUALITY: ' + equality_columns + N' '
+                    CASE WHEN equality_columns IS NOT NULL THEN N'EQUALITY: ' + equality_columns_with_data_type + N' '
```
BrentOzar, line 553: In here, we're checking one column for nulls, but using another. (Greg saw this on the stream, so he's going to dig into it.)

gdoddsy:
I've fixed up based on review, but I'm not getting the same null result that you were getting even before the change when running the repro script. When I run it, I see this:
image
Which is what I was aiming for.
I'm running it against the stack overflow database on SQL 2019 in SQL 2017 and 2019 compatibility level. Any ideas on what different settings we might have?

gdoddsy:
Unrelated thought, if I refer to string_agg in the stored proc, does that prevent it from running on 2008? I'm pretty sure the if logic prevents it from trying to call, but I'm not sure if SQL will still try to compile and look for that function. I don't have a box older than 2016 to test on.

sp_BlitzIndex.sql
gdoddsy: I've fixed the null check, didn't fix the repro step problem.

gdoddsy:
@BrentOzar - I don't know what to do with this one. It works on my machine without getting a null. I'm doing the following:

- Start SQL Server with the StackOverflow database attached used for your Mastering classes
- Run DropIndexes
- Run `SELECT AboutMe FROM dbo.Users WHERE Location = 'Antarctica';` 10 times
- Run `sp_BlitzIndex @TableName = 'Users'`

The result I get is:
image

I'm not sure where to go from here.

I did notice that if I drop compatibility back to 2008 then the call to STRING_SPLIT breaks the code, so you shouldn't take this pull request as it is anyway.

gdoddsy:
I've removed string_split, can now be compiled _and_ returns the updated results on SQL 2008 or newer. I'm going to park trying to recreate the null you were getting (see previous message), I can't work out what I'm doing differently.

If you're planning to do a live code review of this, and you can give me a few hours warning, I'll try and jump on the twitch stream again which might make discussing it easier.

BrentOzar:
This isn't quite working yet, but it's close enough for government work and horseshoes, so I'm going to accept it and start working on a tweak to it.

Issue #2489, opened 2020-07-23
sp_BlitzWho - Parallel blocked processes not showing a blocking_session_id


Adedba:
**Version of the script**
SELECT @Version = '7.97', @VersionDate = '20200712';

**What is the current behavior?**
If a session running a parallel query is blocked it does not show as being blocked.

**If the current behavior is a bug, please provide the steps to reproduce.**

/* Query window 1 , Blocking query*/

BEGIN TRAN 
	UPDATE Users 
	SET Reputation = Reputation +1 
	WHERE Id < 100
--ROLLBACK 

/* Query window 2 , Parallel query blocked by Query window 1 */

SELECT Location,COUNT(*)
FROM Users 
GROUP BY Location

/* Query window 3, run sp_BlitzWho and obeserve the blocking_session_id column */
EXEC sp_BlitzWho;

The blocking_session_id column will show as NULL for Query window two, now if you stop query two and replace it with: 

(Single threaded query on my laptop) 

SELECT TOP 1 * 
FROM Users

run this and then run sp_BlitzWho again and you should see that the blocking_session_id column is now showing correctly.

**What is the expected behavior?**

blocking_session_id should not be NULL for the blocked session.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Tested on SQL2016

Adedba:
Bit of a pain to narrow down this one, so here is what I did: 

Ran sp_BlitzWho with @Debug = 1,

Using the output I found the blocking_session_id column:

				   CASE WHEN r.blocking_session_id <> 0 AND blocked.session_id IS NULL 
							       THEN r.blocking_session_id
							       WHEN r.blocking_session_id <> 0 AND s.session_id <> blocked.blocking_session_id 
							       THEN blocked.blocking_session_id
							       ELSE NULL 
						      END AS blocking_session_id,
							  
Added some new columns alongside for debugging: 
						  
				   CASE WHEN r.blocking_session_id <> 0 AND blocked.session_id IS NULL 
							       THEN 1
							       WHEN r.blocking_session_id <> 0 AND s.session_id <> blocked.blocking_session_id 
							       THEN 2
							       ELSE NULL 
						      END AS blocking_clause,
							  r.blocking_session_id AS r_blocking_session_id,
							  blocked.session_id as blocked_session_id,
							  s.session_id AS s_session_id,
[...]

PR #2500, opened 2020-08-03
2489 parallel blocked processes not showing a blocking session

Adedba:
Version of the script
SELECT @Version = '7.97', @VersionDate = '20200712';

What is the current behavior?
If a session running a parallel query is blocked it does not show as being blocked.

If the current behavior is a bug, please provide the steps to reproduce.

/* Query window 1 , Blocking query*/

BEGIN TRAN
UPDATE Users
SET Reputation = Reputation +1
WHERE Id < 100
--ROLLBACK

/* Query window 2 , Parallel query blocked by Query window 1 */

SELECT Location,COUNT(*)
FROM Users
GROUP BY Location

/* Query window 3, run sp_BlitzWho and obeserve the blocking_session_id column */
EXEC sp_BlitzWho;

The blocking_session_id column will show as NULL for Query window two, now if you stop query two and replace it with:

(Single threaded query on my laptop)

SELECT TOP 1 *
FROM Users

run this and then run sp_BlitzWho again and you should see that the blocking_session_id column is now showing correctly.

What is the expected behavior?

blocking_session_id should not be NULL for the blocked session.

Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?
Tested on SQL2016

### DIFF
--- a/sp_BlitzWho.sql
+++ b/sp_BlitzWho.sql
@@ -335,6 +335,16 @@ BEGIN
 							     THEN blocked.blocking_session_id
 							    ELSE NULL 
 						    END AS blocking_session_id , 
+			       COALESCE(wt.wait_info, RTRIM(blocked.lastwaittype) + '' ('' + CONVERT(VARCHAR(10), 
+						    blocked.waittime) + '')'' ) AS wait_info ,											
+				   CASE WHEN r.blocking_session_id <> 0 AND blocked.session_id IS NULL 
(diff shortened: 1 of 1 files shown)


Adedba:
Forgot to reference #2489 Doh!

BrentOzar:
Woohoo! Thanks, will dig into this this morning and get it into the August release.

Issue #2528, opened 2020-08-13
sp_BlitzCache: Implement OutputServerName 


PierreLetter:
**Is your feature request related to a problem? Please describe.**
Wanting to have central repository for sp_BlitzCache to know what is consuming most ressources in my DataWarehouse platform consisting of multiple SQL Server instances.

**Describe the solution you'd like**
Implementing the same logic as for sp_Blitz (and basically rebuilding the 'ouput to table' logic of sp_blitzCache)
Details on how I am planning to do it below.

**Are you ready to build the code for the feature?**
Took the core 'Output to table logic but encapsulated it in what is done in sp_Blitz to cover all cases:

If OutputTableName is NOT NULL
Go to OutputResultsToTable
=> removed the other checks to allow for ## output tables, like in sp_Blitz. Figured it would be useful to somebody.

IF @OutputServerName IS NOT NULL
=> same logic as sp_Blitz setting ValidOutputServer and ValidOutputLocation variables

IF @ValidOutputLocation = 1
=> create the ouput table if it doesn't exist.
    IF @ValidOutputServer = 1
    doubling quotes for execution at target
Kept the Issue 2377 and Issue 2514 updates in there even if obsolete for a 1st run. In case somebody wants to update an old table.

IF @ValidOutputServer = 1
=> insert at target else execute locally

ELSE IF (SUBSTRING(@OutputTableName, 2, 2) = '##')
=> do it for the ## table

BrentOzar:
Looks good! Merging into the dev branch, will be in the Sept release with credit to you in the release notes.

PR #2529, opened 2020-08-13
sp_BlitzCache remote

PierreLetter:
Implementation of #2528 

Tested on SQL Server 2019 CU6 Enterprise:
local server for regular table in DBATools DB without linked server
local server for regular table in DBATools DB with linked server
local server on ##BlitzCacheResultsTemp table
target server on DBATools DB using linked server

for each case with table existing and not existing.
only tested changing @Top parameter for now to make sure the sp_execute for insertion worked properly with parameters.

One know case where the procedure breaks is if existing table already exists on target and was created locally, and thus having one columns in XML and not NVARCHAR(MAX) format (same as sp_Blitz).

Will continue testing next week to validate all parameter settings, just in case.


BrentOzar:
Looks good! Merging into the dev branch, will be in the Sept release with credit to you in the release notes.

PR #2537, opened 2020-08-21
sp_DatabaseRestore should use the DatabaseName when @RestoreDatabaseN…

### LINKED ISSUE #2536: sp_databaserestore fails when @RestoreDatabaseName is empty string
When suppling an empty string as the RestoreDatabaseName, sp_DatabaseRestore errors out with a cryptic message:

`An object or column name is missing or empty. For SELECT INTO statements, verify each column has a name. For other statements, look for empty alias names. Aliases defined as "" or [] are not allowed. Change the alias to a valid name.
`
EXEC sp_DatabaseRestore @Database = N'Samples'	
						,@RestoreDatabaseName = N''
						,@MoveDataDrive = N'f:\samples\'
						,@MoveLogDrive = N'f:\samples\'
						,@BackupPathFull = N'c:\temp\copy to F drive\Backup\Full'
						,@RunRecovery = 1										
						,@DatabaseOwner = 'sa'		

Ideally, the database will treat the empty string as if no string was specified, or alternatively it will error out with a useful error message.

I will submit a pull request shortly to fix this bug

gdoddsy:
sp_DatabaseRestore should use the DatabaseName when @RestoreDatabaseName is an empty string

Note: If you try to restore the database with the name ' ' (single space) then this will restore the database with the @DatabaseName instead.

Fixes #2536

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -433,7 +433,8 @@ BEGIN
 	IF @Execute = 'Y' OR @Debug = 1 RAISERROR('Fixing @StandbyUndoPath to add a "/"', 0, 1) WITH NOWAIT;
 	SET @StandbyUndoPath += N'/';
 END;
-IF @RestoreDatabaseName IS NULL
+
+IF @RestoreDatabaseName IS NULL OR @RestoreDatabaseName = N''
 BEGIN
 	SET @RestoreDatabaseName = @Database;
 END;


gdoddsy:
I was confused by how N'' = N' ', then I found this: https://www.brentozar.com/archive/2017/04/surprising-behavior-trailing-spaces/

How important is it that you can restore a database with just spaces in it's name?

PR #2544, opened 2020-08-24
Add check (235) for inconsistent query store metadata

### LINKED ISSUE #2525: Add a new check for sys.databases's is_query_store_on vs. sys.database_query_store_options desired/actual state
sys.databases maintains a "cached" bit indicating whether query store is enabled for a given database, however this value can be out of sync with reality. Database mirroring failover seems to disable query store in the database but does not update the bit in sys.databases, perhaps other stuff does too. A naïve scan of sys.databases where is_query_store_on = 1 therefore returns incorrect information. 

Add a check in sp_Blitz for the bit in sys.databases not being in sync with the actual state in database_query_store_options.

No real alternatives to the check, other than obviously not checking. :-) 

Happy to build the fix and provide a PR. Per instructions in the check docs I'm opening the Issue first to discuss general interest in adding the check and what the right check values would be. 

The check # would simply be the next number available. For priority, I'm thinking either 200 or a new 205? 200 is the priority for the current check if query store is enabled at all (check 163). For Findings Group it's probably Performance (like 163), but perhaps Informational?

WaldenL:
Fixes #2525 by adding a check for inconsistent metadata between master's is_query_store_on in sys.databases and the actual state of the query store according to the database-specific QS metadata. 

I was able to test with this hacky-AF approach:
```
CREATE DATABASE test_qs;

ALTER DATABASE test_qs SET OFFLINE;
--copy the mdf/ldf
ALTER DATABASE test_qs SET ONLINE;

ALTER DATABASE test_qs SET QUERY_STORE = ON;

ALTER DATABASE test_qs SET OFFLINE;
--restore the (old) mdf/ldf
ALTER DATABASE test_qs SET ONLINE;

--run sp_blitz and check for finding

DROP DATABASE test_qs;
```

(Also tweaked an outdent on an IF in the same check area)

### DIFF
--- a/Documentation/sp_Blitz_Checks_by_Priority.md
+++ b/Documentation/sp_Blitz_Checks_by_Priority.md
@@ -6,8 +6,8 @@ Before adding a new check, make sure to add a Github issue for it first, and hav
 
 If you want to change anything about a check - the priority, finding, URL, or ID - open a Github issue first. The relevant scripts have to be updated too.
 
-CURRENT HIGH CHECKID: 234.
-If you want to add a new one, start at 235.
+CURRENT HIGH CHECKID: 235.
+If you want to add a new one, start at 236.
 
 | Priority | FindingsGroup | Finding | URL | CheckID |
 |----------|-----------------------------|---------------------------------------------------------|------------------------------------------------------------------------|----------|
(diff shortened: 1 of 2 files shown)


WaldenL:
FWIW, the real-world place I'm seeing this issue is mirroring failover. We're mirroring a bunch of databases -- yes, I know mirroring bad blaa blaa :-) -- and as we flip flop between servers for patching all our databases get set to query store = off. Well, at least they do in the database, but master still says they're enabled.

BrentOzar:
Looks good! Thanks for the pull request. Merged into the dev branch, and will be in the September release with credit to you in the release notes.

Issue #2533, opened 2020-08-17
XML parsing error when running sp_BlitzIndex


KuldarAndares:
**Version of the script**
sp_BlitzIndex(TM) v7.98 - August 08, 2020

**What is the current behavior?**
I get error:
Msg 50000, Level 16, State 1, Procedure sp_BlitzIndex, Line 1980 [Batch Start Line 0]
DB_name database failed to process. XML parsing: line 1, character 4000, unexpected end of input

**If the current behavior is a bug, please provide the steps to reproduce.**
I execute script from master:
EXEC sp_BlitzIndex @DatabaseName=N'DB_name'

From messages tab I can see:
Starting run. sp_BlitzIndex(TM) v7.98 - August 08, 2020
Create temp tables.
Adding UQ index on #IndexSanity (database_id, object_id, index_id)
Number of databases to examine: 1
Checking partition counts to exclude databases with over 100 partitions
Starting loop through databases

DB_name
Inserting data into #IndexColumns for clustered indexes and heaps
Inserting data into #IndexColumns for nonclustered indexes
Inserting data into #IndexSanity
Checking partition count
Preferring non-2012 syntax with LEFT JOIN to sys.dm_db_index_operational_stats
Inserting data into #IndexPartitionSanity
Warning: Null value is eliminated by an aggregate or other SET operation.
Inserting data into #MissingIndexes
Failure populating temp tables.
Last @dsql: SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;WITH ColumnNamesWithDataTypes AS(SELECT id.index_handle,id.object_id,cn.IndexColumnType,STUFF((SELECT cn_inner.ColumnName + ' ' +
					N' {' + CASE	 WHEN ty.name IN ( 'varchar', 'char' ) THEN ty.name + '(' + CASE WHEN co.max_length = -1 THEN 'max' ELSE CAST(co.max_length AS VARCHAR(25)) END + ')'
								WHEN ty.name IN ( 'nvarchar', 'nchar' ) THEN ty.name + '(' + CASE WHEN co.max_length = -1 THEN 'max' ELSE CAST(co.max_length / 2 AS VARCHAR(25)) END + ')'
								WHEN ty.name IN ( 'decimal', 'numeric' ) THEN ty.name + '(' + CAST(co.precision AS VARCHAR(25)) + ', ' + CAST(co.scale AS VARCHAR(25)) + ')'
[...]

PR #2545, opened 2020-08-25
Stop truncating columns

gdoddsy:
Finishing off the fix for #2533 

Existing error in sp_BlitzIndex where #missingindexes was limiting equality/inequality/included columns to nvarchar(4000). As suspected on the live stream, that has been like that since day 1 of open sourced code. I've bumped that up to nvarchar(max).

Also a bunch of CASTs to make sure the columns are not truncated when inserting in there..


BrentOzar:
Nice work! Merging the pull request into dev, will be in the September release with credit to you in the release notes.

Issue #2548, opened 2020-08-26
sp_BlitzFirst - CheckID 44 gets blocked when an index operation is in progress


Adedba:
**Version of the script**
SELECT @Version = '7.97', @VersionDate = '20200712';

**What is the current behavior?**
If an index operation is in progress the statistics updated check gets blocked, if you are logging sp_blitzfirst output to a table then this can leave gaps due to long running collections.

**If the current behavior is a bug, please provide the steps to reproduce.**

**Query window 1:**

Use [StackOverflow2010];

```
/* rebuild a reasonable size index a few times just so we do not miss it during its operation */

	ALTER INDEX [PK_Votes__Id] ON [dbo].[Votes] REBUILD PARTITION = ALL WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON);
	ALTER INDEX [PK_Votes__Id] ON [dbo].[Votes] REBUILD PARTITION = ALL WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON);
	ALTER INDEX [PK_Votes__Id] ON [dbo].[Votes] REBUILD PARTITION = ALL WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON);
	ALTER INDEX [PK_Votes__Id] ON [dbo].[Votes] REBUILD PARTITION = ALL WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON);

```

**Query window 2:**

```
/* Execute sp_BlitzFirst */
EXEC sp_BlitzFirst ;

```

**Query Window 3:** 

```
/* Check for blocking when sp_Blitzfirst reaches CheckID 44 */

EXEC sp_BlitzWho;
```

[...]

PR #2555, opened 2020-08-30
sp_BlitzFirst - CheckID 44 gets blocked when an index operation is in progress

Adedba:
#2548 Within the CheckId 44 code block we now set lock timeout = 0 , in the foreachDB we Try/catch and if we get an error we report the error handling error 1222 (timeout) with a slightly different message and finally set lock timeout back to default of -1. we only show these rows when there are actually updated statistics to show as it is a priority 50 we don't want false alarms.


BrentOzar:
Thanks for the pull request! Merged into the dev branch, will be in the September release with credit to you in the release notes.

PR #2558, opened 2020-09-01
Update sp_blitzcache to remove unqualified join

ianmanton:
Changing unqualified join to a cross apply (as Microsoft data migration wizard complains about it "Starting with database compatibility level 90 and higher, in rare occasions, the 'unqualified join' syntax can cause 'missing join predicate' warnings, leading to long running queries.").  There may be more but there was an awful lot of "from"s to look through in the SP so only spotted this one.


BrentOzar:
Hi Ian, and thanks for the PR. Going forward, please open an issue before doing a pull request. That info is detailed here in our contributing guide: https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/main/CONTRIBUTING.md

Thanks!

ianmanton:
Whoops! Sorry Brent!

PR #2567, opened 2020-09-06
Fix the bug in BlitzWho and adapt BlitzFirst table

AdrianB1:


sp_BlitzFirst.sql
```diff
@@ -1009,22 +1009,7 @@ BEGIN
...
-    BEGIN
-		RAISERROR('Checking for Read intent databases to exclude',0,0) WITH NOWAIT;
-
-        SET @StringToExecute = 'INSERT INTO #ReadableDBs (database_id) SELECT DBs.database_id FROM sys.databases DBs INNER JOIN sys.availability_replicas Replicas ON DBs.replica_id = Replicas.replica_id WHERE replica_server_name NOT IN (SELECT DISTINCT primary_replica FROM sys.dm_hadr_availability_group_states States) AND Replicas.secondary_role_allow_connections_desc = ''READ_ONLY'' AND replica_server_name = @@SERVERNAME;';
-        EXEC(@StringToExecute);
-		
-	END
-
-    DECLARE	@v DECIMAL(6,2),
+	DECLARE	@v DECIMAL(6,2),
```
BrentOzar, line 1012: Why did we take the ReadableDBs population line out of this section? We do need it for sp_BlitzCache I think, based on a quick code review. I'm guessing you wanted to use it down in the sp_BlitzWho section maybe, or maybe you didn't mean to include this file in your pull request? Let me know either way.

sp_BlitzWho.sql
```diff
@@ -522,6 +522,21 @@ IF @OutputDatabaseName IS NOT NULL AND @OutputSchemaName IS NOT NULL AND @Output
 
  END
 
+IF OBJECT_ID('tempdb..#ReadableDBs') IS NOT NULL
```
BrentOzar, line 525: If we're populating a list of ReadableDBs in here, why do we need to make any changes in sp_BlitzFirst?

BrentOzar (changes requested):
Not sure if you meant to check in both sp_BlitzFirst and sp_BlitzWho. Can you give me a quick rundown of why you did what you did, and whether it'll break other things, like how sp_BlitzFirst calls sp_BlitzCache? I think I might not have understood something. Thanks!

AdrianB1:
The implementation of sp_BlitzWho is identical with sp_BlitzCache; a local ReadableDBs is needed in each to be able to run independently when needed.

But with this change an error appears in sp_BlitzFirst:  (Message 50000)  Invalid object name '#ReadableDBs'. Moving the creation of ReadableDB's in sp_BlitzFirst after calling sp_BlitzWho is solving the problem. I did not figure yet why this temp table is reported missing after sp_BlitzWho runs, I will continue to investigate that.

BrentOzar:
Okay, if you change sp_BlitzFirst, make sure sp_BlitzCache still runs appropriately. My guess is that the change you made to sp_BlitzFirst may break sp_BlitzCache's functionality on readable secondaries.

I'm going to go ahead and close this, but give 'er another shot when you're ready. Thanks!

Issue #2573, opened 2020-09-10
expand Github issue #598 to include the 19 new database scoped configurations that can be changed in SQL 2019


mtndew01:
**Is your feature request related to a problem? Please describe.**
Currently sp_BlitzFirst only looks at the database scoped configurations of MAXDOP, LEGACY_CARDINALITY_ESTIMATION, PARAMETER_SNIFFING, and QUERY_OPTIMIZER_HOTFIXES. 

**Describe the solution you'd like**
SQL 2019 introduces the following options for database scoped configurations
IDENTITY_CACHE
INTERLEAVED_EXECUTION_TVF
BATCH_MODE_MEMORY_GRANT_FEEDBACK
BATCH_MODE_ADAPTIVE_JOINS
TSQL_SCALAR_UDF_INLINING
ELEVATE_ONLINE
ELEVATE_RESUMABLE
OPTIMIZE_FOR_AD_HOC_WORKLOADS
XTP_PROCEDURE_EXECUTION_STATISTICS
XTP_QUERY_EXECUTION_STATISTICS
ROW_MODE_MEMORY_GRANT_FEEDBACK
ISOLATE_SECURITY_POLICY_CARDINALITY
BATCH_MODE_ON_ROWSTORE
DEFERRED_COMPILATION_TV
ACCELERATED_PLAN_FORCING
GLOBAL_TEMPORARY_TABLE_AUTO_DROP
LIGHTWEIGHT_QUERY_PROFILING
VERBOSE_TRUNCATION_WARNINGS
LAST_QUERY_PLAN_STATS

**Are you ready to build the code for the feature?**
Lets do this!

BrentOzar:
Sure, sounds good. Just make sure it can be run repeatedly without errors, works on case-sensitive servers, on 2008+, etc. Looking forward to the PR - thanks!

BrentOzar:
Looks good! Thanks for the pull request. Merging into the dev branch, and will be in the October release with credit to you in the release notes.

Small tweak - the new checks also have to be added in the Documentation folder, but I'll knock that out today since it also involves pulling URLs for each of 'em so folks know where to learn more.

PR #2596, opened 2020-09-24
#2573 add new 2019 database scoped configurations

mtndew01:
added in the 19 new database scoped configurations available in 2019.  Tested against a 2016 instance as well as 2019 and verified that the differences are found and returned in the sp_blitz output.


BrentOzar (approved):
Looks good! Thanks for the pull request. Merging into the dev branch, and will be in the October release with credit to you in the release notes.

Small tweak - the new checks also have to be added in the Documentation folder, but I'll knock that out today since it also involves pulling URLs for each of 'em so folks know where to learn more.

PR #2605, opened 2020-10-02
Feature/sort option sp blitz who

DPetrusma:
I've added @SortOrder with a default of 'elapsed_time' which is what it currently sorts by. The ORDER BY appended at the end handles the two values I find useful, plus ELSE defaults to elapsed_time, again the current behaviour.

This procedure doesn't seem to contain the big explanations of all the possible parameter values so I've not included those.

Running on my machine works with no value specified, 'elapsed_time', nonsense text, and 'session_id'


BrentOzar (approved):
Thanks for the pull request, and nice work! I'm merging this into the dev branch, and it'll be in the October release with credit to you in the release notes.

PR #2617, opened 2020-10-07
Extra columns in histogram query

### LINKED ISSUE #2612: sp_BlitzIndex - show modification counter in #1900 stats query
In the `dm_db_stats_properties` view, there's a modification counter column that might be useful to show here, though as I'm writing this I'm half-remembering some weird caveat with it. That gets hit in the query for #1900 that shows the histogram.

Add the `modification_counter` column to the query.

Not knowing how many modifications the lead column in the stats object as had since it was last updated. 

Yes, it's just adding a column to the select list.

erikdarlingdata:
Grabbed some additional columns, also fixed a case sensitivity issue with the `ObjectID` parameter in a couple places that was preventing compilation.

Closes #2612


BrentOzar:
Awww, dagnabbit, I accepted the other pull request before this one. Can you pull the origin and then that'll reset the changes and hopefully make it so I can pull this w/o conflicts? It looks like a no-brainer.

erikdarlingdata:
Yeah, lemme take a look.

Issue #2621, opened 2020-10-08
sp_ineachdb: using @database_list parameter causes error


philosophicles:
`sp_ineachdb`, version 2.99 (date 20200913)
**Suggested fix included below - I can submit a PR if desired.** 

**What is the current behavior?**
Calling this sproc with the `@database_list` parameter seemingly _always_ results in this error: 

```
Invalid length parameter passed to the LEFT or SUBSTRING function.
```

For example (modified example query from [Aaron's blog post](https://www.mssqltips.com/sqlservertip/5694/execute-a-command-in-the-context-of-each-database-in-sql-server--part-2/)): 

```sql
EXEC dbo.sp_ineachdb
@command = N'SELECT db = ''?'', COUNT(*) FROM sys.objects;'
,@database_list = 'master'
;
```

If I'm missing something blindingly obvious and this is user error not a bug, please do scold me appropriately. 

The relevant code snippet is: 

```sql
IF @database_list > N''
  -- comma-separated list of potentially valid/invalid/quoted/unquoted names
  BEGIN
    ;WITH n(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM n WHERE n < 4000),
    names AS
    (
      SELECT name = LTRIM(RTRIM(PARSENAME(SUBSTRING(@database_list, n, 
        CHARINDEX(N',', @database_list + N',', n) - n), 1)))
      FROM n 
      WHERE n <= LEN(@database_list)
        AND SUBSTRING(N',' + @database_list, n, 1) = N','
    ) 
    INSERT #ineachdb(id,name,is_distributor) 
    SELECT d.database_id, d.name, d.is_distributor
      FROM sys.databases AS d
      WHERE EXISTS (SELECT 1 FROM names WHERE name = d.name)
      OPTION (MAXRECURSION 0);
  END
```

I haven't _fully_ understood why this is happening (and really want to!), but I can say that changing the `WHERE EXISTS` line to an `INNER JOIN` between `sys.databases` and `names` prevents it - this is what the next block of code for the `@exclude_list` parameter does. However, they're not truly equivalent of course, and with the include list, could lead to the same database being listed twice (e.g. `@database_list = 'master,master'`). The correlated subquery seems to be what breaks it, therefore: `WHERE d.name in (SELECT name FROM names)` also fails.

I believe a more elegant fix would be to modify the numbers CTE one-liner to this: 

```sql
[...]

PR #2630, opened 2020-10-12
Debug evaluation of @database_list parameter #2621

philosophicles:
Change in database_list parameter block prevents runtime error `Invalid length parameter...`. 
Change repeated in exclude_list block for consistency; keep numbers CTE as short as it can be.

Change tested using: 

```
EXEC dbo.sp_ineachdb
@command = N'SELECT db = ''?'', COUNT(*) FROM sys.objects;'
,@database_list = 'master,[master],mister,SCRATCH,foobar,msdb,tempdb'
,@exclude_list = 'ping,pong,msdb'
;
```

(I have a SCRATCH database in my test environment but mister, foobar, ping, pong do not exist.)

### DIFF
--- a/sp_ineachdb.sql
+++ b/sp_ineachdb.sql
@@ -100,22 +100,22 @@ BEGIN
           @cmd    nvarchar(max),
           @thisdb sysname,
           @cr     char(2) = CHAR(13) + CHAR(10),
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks! Merged into dev, will be in the November release with credit to you in the release notes.

Issue #2482, opened 2020-07-17
sp_BlitzCache - Query hash sort order check skipped on case sensitive instances if query hash is not lower case


Adedba:
**Version of the script**
SELECT @Version = '7.97', @VersionDate = '20200712';

**What is the current behavior?**
@OnlyQueryHashes list population skipped on case sensitive instances if @SortOrder incorrectly cased

If I execute sp_BlitzCache on a case sensitive instance and incorrectly case @SortOrder = 'query hash' only one result set is returned , no query details are returned just a subset of findings.

**If the current behavior is a bug, please provide the steps to reproduce.**
Connect to a case sensitive instance and execute EXEC sp_BlitzCache @SortOrder = 'Query hash';

**What is the expected behavior?**
To ignore my casing on @SortOrder and return the results consistently

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
So far I have tested on SQL2016

Should be a simple fix here I think, just move the existing SET @SortOrder = LOWER(@SortOrder); further up in the script just after the Validate version check. I will get this committed later.

BrentOzar:
Good catch! Yeah, looks like we set the @SortOrder to lower case all the way down in line 1091.

BrentOzar:
Great job! Merged into dev, will be in the August release with credit to you in the release notes.

PR #2632, opened 2020-10-15
Main

MaikelvanGorkom:
Based upon https://youtu.be/2e1fKTzGQtE?t=3374 (#2482 ) did some cleaning up: removed unnecessary LOWER(@SortOrde)


MaikelvanGorkom:
Did something wrong. My bad, will have a look on how to work with GITHUB when i've got more time.

PR #2637, opened 2020-10-18
Re-fix parallel deadlock stuff

### LINKED ISSUE #2636: sp_BlitzLock and parallel deadlocks
**Version of the script**
`SELECT @Version = '2.999', @VersionDate = '20201011';`

Complicated!

In newer versions of SQL Server, parallel deadlocks don't register unless they error out, with the message

> Msg 1205, Level 13, State 18, Line 3
> Transaction (Process ID 52) was deadlocked on lock | communication buffer resources with another process and has been chosen as the deadlock victim. Rerun the transaction.

This is a change that started with 2017 CU10, and also applies to SQL Server 2016 SP2, and SQL Server 2019 (to my knowledge):

[FIX: Many xml_deadlock_report events are reported for one single intra-query deadlock occurrence in SQL Server 2016 and 2017](https://support.microsoft.com/en-us/help/4338715/many-xml-deadlock-report-events-are-reported-for-one-intra-query)

The way that the code was written in #1499 addressed normal parallel exchange spills (deadlocks), but not the errored out ones. In fact, a lot of the code around adding that detection in was defensive, because, like the KB article said, it would result in [a lot of XML being generated](https://www.youtube.com/watch?v=pzaQjS1JstY).

The deadlocks that produce errors have slightly different characteristics, that we don't pick up on.
image

This messes up both the way that BlitzLock figures out if deadlocks are parallel, and the way that it grabs information about them. It also came to my attention while reviewing this that we don't have a roll-up warning about parallel deadlocks.

There's a repro for a normal exchange spill in #1499, the type that no longer produces a deadlock graph. A query that will produce one is like so:

```
DROP TABLE IF EXISTS #ohno

SELECT c.Id, c.UserId, c.CreationDate, c.PostId, c.Score 
INTO #ohno 
FROM dbo.Comments AS c WITH (TABLOCKX)
```

I think the best thing to do would be to add an additional check to the initial XML grab to figure out if XML fragments specific to parallel deadlocks exist, and use that to filter on. Otherwise, the query to filter regular vs parallel deadlocks becomes quite convoluted. And also add a count of parallel deadlocks to the roll-up. I know, I know, it's two different things, but they're closely related.

Code is just about ready to go.

Very all so many

erikdarlingdata:
This PR does the following:
 - Adds a check to the initial XML grab to detect if the deadlock is from intra-query parallelism
 - Adds a total count of parallel deadlocks to the roll-up
 - Corrects missing logic from the inserts
 - Attempts to grab object names for parallel deadlocks (we weren't before for some reason)
 - Fixes the filtering for the results queries
 - Fixes the outer applies to account for a `wait_type` that wasn't there before

Closes #2636

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -320,12 +320,12 @@ You need to use an Azure storage account, and the path has to look like this: ht
 		ORDER BY xml.deadlock_xml.value('(/event/@timestamp)[1]', 'datetime') DESC
 		OPTION ( RECOMPILE );
 
-
 		/*Parse process and input buffer XML*/
         SET @d = CONVERT(VARCHAR(40), GETDATE(), 109);
         RAISERROR('Parse process and input buffer XML %s', 0, 1, @d) WITH NOWAIT;
         SELECT      q.event_date,
                     q.victim_id,
+					CONVERT(BIT, q.is_parallel) AS is_parallel,
                     q.deadlock_graph,
                     q.id,
                     q.database_id,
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks sir! Merged into the dev branch, will be in the November release with credit to you in the release notes. I didn't even attempt to test this one, hahaha.

PR #2643, opened 2020-10-21
Update sp_BlitzIndex.sql

excelosaurus:
Corrected CASE x WHEN NULL THEN statement, which will never be true; syntax in this case is CASE WHEN x IS NULL THEN.


BrentOzar:
Thanks for the pull request, merged into the dev branch and will be in the November release with credit to you in the release notes.

PR #2649, opened 2020-10-29
Update sp_BlitzWho.sql

### LINKED ISSUE #2491: sp_BlitzWho throws an error when HA readable replica exists
**Version of the script**
7.97 20200712

Executed manually several times, everything is fine, but when ran as a SQL job it failed every time "The target database (xxx) is in an availability group and is currently accessible for connections when the application intent is set to read only. For more information about application intent, see SQL Server Books Online.". 

Removing @OutputTableNameBlitzWho = 'BlitzWho' fixed the job. 

Run:
sp_blitzfirst @OutputDatabaseName = 'DBATools',
@OutputSchemaName = 'dbo',
@OutputTableName = 'BlitzFirst',
@OutputTableNameFileStats = 'BlitzFirst_FileStats',
@OutputTableNamePerfmonStats = 'BlitzFirst_PerfmonStats',
@OutputTableNameWaitStats = 'BlitzFirst_WaitStats',
@OutputTableNameBlitzCache = 'BlitzCache',
@OutputTableNameBlitzWho = 'BlitzWho'
it will fail sometimes in a HA configuration with read only replicas.

Commenting out @OutputTableNameBlitzWho = 'BlitzWho', it will run fine. I think it fails only when there are open connections to one of the read only replicas, if there is none then it runs fine.

No errors.

SQL 2016 CU10 13.0.5492.2 on Windows 2016

AdrianB1:
Fixes #2491.

Had to use a different name for the temp table in sp_BlitzWho, because if I keep ReadableDBs and I delete it and recreate it in sp_BlitzWho (to work fine if the sp is called alone) then after sp_BlitzFirst is calling sp_BlitzWho the ReadableDBs will not exist in sp_BlitzFirst; this way BlitzFirst will keep #ReadableDBs and BlitzWho will have WhoReadableDBs. 

There was also the option to not delete ReadableDBs in BlitzWho if exists, but then extra checks were needed in BlitzWho to make sure the table is right, not just the name. I skipped the option as it's more complex.

### DIFF
--- a/sp_BlitzWho.sql
+++ b/sp_BlitzWho.sql
@@ -525,6 +525,21 @@ IF @OutputDatabaseName IS NOT NULL AND @OutputSchemaName IS NOT NULL AND @Output
 
  END
 
+ IF OBJECT_ID('tempdb..#WhoReadableDBs') IS NOT NULL 
+	DROP TABLE #WhoReadableDBs;
+
+CREATE TABLE #WhoReadableDBs 
+(
+database_id INT
+);
+
+IF EXISTS (SELECT * FROM sys.all_objects o WHERE o.name = 'dm_hadr_database_replica_states')
+BEGIN
+	RAISERROR('Checking for Read intent databases to exclude',0,0) WITH NOWAIT;
+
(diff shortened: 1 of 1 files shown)


AdrianB1:
https://youtu.be/8djcDJrph88?t=1924 - why I moved the ReadableDBs later in sp_BlitzFirst - to avoid the situation where the table was created there, then calling sp_BlitzWho would delete it an recreate it, on cpmpletion of BlitzWho the table is removed and on return to BlitzFirst it is no longer existing. Moving it after BlitzWho call avoided completely this case. In the fix above I used a different workaround, used a different name in BlitzWho so the original table in BlitzFirst is not touched.

BrentOzar (approved):
THAT IS GENIUS. Totally makes sense, and I appreciate your explanation. Good catch. Merged into the dev branch, and it'll be in the November release with credit to you in the release notes.

BrentOzar:
THAT IS GENIUS. Totally makes sense, and I appreciate your explanation. Good catch. Merged into the dev branch, and it'll be in the November release with credit to you in the release notes.

Issue #2644, opened 2020-10-21
sp_DatabaseRestore should be able to query MSDB to find the last restored log file and continue restoring from there


gdoddsy:
I'm trying to use sp_DatabaseRestore to apply log files to a warm standby server. I know a chunk of work went into sp_AllNightLog...but I don't want all that overhead. I just want to restore all of the log files after the last log file that was restored, and I'll chuck this into an agent job myself.

We have a log retention period of 90 days, so I don't want to have to scan 90 days worth of files to restore the last 15 minutes of logs. I need a way to set OnlyLogsAfter based on something that can be calculated.

**Describe the solution you'd like**
A flag I can set on sp_DatabaseRestore that will query MSDB to find the last restored log file, and then start restoring from there. It can assume that all it is doing is continuing logs, not restoring fulls/diffs, and can also assume that nobody else is apply diff/full backups over the top.

**Describe alternatives you've considered**
I could setup sp_AllNightLog...but it's a lot of overhead when I only want to do this on 1 database
I could use built in log shipping, but I don't want to change the jobs we have setup to run Ola's jobs as we like how it's all running currently.
I could do something similar to ap_AllNightLog and keep track of the last file that was uploaded, and then use the OnlyLogsAfter parameter that already exists. But then we need tables in a database somewhere, and we don't really have anywhere (trying to avoid using System databases, and seems a waste to create a database just to store this one field).

**Is there another obvious alternative that I'm missing?** Maybe this is already implemented and I'm just not finding the right combination of parameters?

**Are you ready to build the code for the feature?**
Yep, I've already got the basics of it in my mind. I'll hopefully have it going in the next week or so.

BrentOzar:
@rducom thanks for the pull request! Looks good. Merging into the dev branch, will be in the November release with credit to you in the release notes.

PR #2663, opened 2020-11-06
Adds optional @IgnoreAlreadyPresentInMsdb BIT setting

rducom:
We are experiencing serious slow down while restoring backups from distant datacenter, when having a lot of TRN files, because of `RESTORE HEADERONLY FROM DISK` calls on outdated TRN files.

Proposal for https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2644

This approach is naive, by filtering out TRN with names > to the last one registered in msdb.dbo.backupmediafamily.


BrentOzar (approved):
Thanks for the pull request! Looks good. Merging into the dev branch, will be in the November release with credit to you in the release notes.

PR #2675, opened 2020-11-13
Fix multiple plan query

### LINKED ISSUE #2653: sp_BlitzCache: Parameterized queries in different databases are flagged as multiple plans
**Version of the script**
`@Version = '7.999', @VersionDate = '20201011'`

If you have a database per tenant model where the same queries run, they're flagged as multiples even when they're parameterized.

```
DBCC FREEPROCCACHE;
GO 

DECLARE @r INT = 1;
DECLARE @sql NVARCHAR(MAX) = N'SELECT COUNT(*) AS records FROM dbo.Users AS u WHERE u.Reputation = @r;'

EXEC StackOverflow2013.sys.sp_executesql @sql, N'@r INT', @r;
EXEC StackOverflow.sys.sp_executesql @sql, N'@r INT', @r;

EXEC sp_BlitzCache @OnlySqlHandles = '0x02000000E569613767CC36A863977FFE3AFBB63C09CD37060000000000000000000000000000000000000000'
```
image

I addressed this in the query that looks at this to sort by query hash (#2102), and I think I can do the same here, by also grouping by database name:

image

SO MANY!

erikdarlingdata:
Add DatabaseName to group by and join.

Also discovered that this particular update was running twice, so remove the second one. This marks the first time BlitzCache has ever gotten shorter.

Sorry if that screws up any metrics for you 😃

Closes #2653

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -2810,12 +2810,15 @@ SET     NumberOfDistinctPlans = distinct_plan_count,
 FROM (
         SELECT  COUNT(DISTINCT QueryHash) AS distinct_plan_count,
                 COUNT(QueryHash) AS number_of_plans,
-                QueryHash
+                QueryHash,
+				DatabaseName
         FROM    ##BlitzCacheProcs
 		WHERE SPID = @@SPID
-        GROUP BY QueryHash
+        GROUP BY QueryHash,
+		         DatabaseName
 ) AS x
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Looks good! Perfect timing too, doing the release today. Will be in the Nov release with credit to you in the release notes. Thanks sir.

BrentOzar:
Looks good! Perfect timing too, doing the release today. Will be in the Nov release with credit to you in the release notes. Thanks sir.

Issue #2719, opened 2020-12-14
sp_ineachdb - broken on latest release when using @database_list


RichBenner:
**Version of the script**
Current - @Version = '2.99999', @VersionDate = '20201211';

**What is the current behavior?**

(this was reported in the #firstresponderkit slack channel)

When passing @database_list we're not populating #ineachdb because the block of code that does this was commented out in the latest user submitted pull request.

**If the current behavior is a bug, please provide the steps to reproduce.**

run this with your own CSV list of databases, it won't do anything.

EXEC sp_ineachdb @database_list = N'a,b,c', @command = N'whatever command'

**What is the expected behavior?**

It would be nice if it actually ran the command against these databases.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**

It's the script, not the SQL Version and/or the OS

RichBenner:
In fact, it looks simpler than that. On the previous pull request it appears we should be doing an insert into #ineachdb at around line 200, but we're only doing a select. 

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/pull/2704/files

BrentOzar:
Good catch! Merged into the dev branch, and will be in the January release with credit to you in the release notes.

PR #2720, opened 2020-12-14
Update sp_ineachdb - we forgot an insert

RichBenner:
There's an insert missing from the latest version of this proc so we never actually populate #ineachdb


RichBenner:
fixes #2719

BrentOzar:
Good catch! Merged into the dev branch, and will be in the January release with credit to you in the release notes.

Issue #2721, opened 2020-12-14
sp_BlizWho: Ordering is heckin weird now


erikdarlingdata:
**Version of the script**
December 2020

**What is the current behavior?**
It looks like something went amok in #2652 and not all queries get a `00`, which causes odd sorting. The one at the bottom should be third in the list.

![image](https://user-images.githubusercontent.com/2136037/102126169-bdd62f80-3e18-11eb-8682-90d769251633.png)

**If the current behavior is a bug, please provide the steps to reproduce.**
From what I can tell, just running sp_BlitzWho with several queries running

**What is the expected behavior?**
Ordering should, by default, be query execution time (unless I missed some other change).

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Many and more

BrentOzar:
Ouch. That's not good. I'll dig into that when I get back home, and worst case scenario, I'll just back out last month's changes.

erikdarlingdata:
You may want to just back them out for now. It's really screwy.

```
SELECT 
    RIGHT('00' + CONVERT(VARCHAR(20), (ABS(r.total_elapsed_time) / 1000) / 86400), 2),
    CONVERT(VARCHAR(20), (ABS(r.total_elapsed_time) / 1000) / 86400)
FROM (SELECT 2000000000) AS r (total_elapsed_time);

SELECT 
    RIGHT('00' + CONVERT(VARCHAR(20), (ABS(r.total_elapsed_time) / 1000) / 86400), 2),
    CONVERT(VARCHAR(20), (ABS(r.total_elapsed_time) / 1000) / 86400)
FROM (SELECT 3000000000) AS r (total_elapsed_time);
```

![image](https://user-images.githubusercontent.com/2136037/102218247-af852380-3eab-11eb-8434-7b73a06c87f2.png)

It looks fine when there's a whole number, but when there are decimal places it grabs the very end of the decimals.

alihacks:
@erikdarlingdata I was not able to tie your example to reproducing the issue as the value returned by the DMV is INT not float, I only get int results.
I was however able to figure out one discrepancy - the 00 padding was added to the first piece in COALESCE (when total_elapsed_time is available) but not 2nd part (falling back to using last_request_start_time). I will line up a PR to address this shortly.

BrentOzar:
Thanks! Merging into the dev branch, will be in the January release with credit to you in the release notes.

PR #2728, opened 2020-12-16
sp_BlitzWho ordering

alihacks:
Looking at #2721 it seems that the issue is caused by the DECIMAL type returned by ABS, adding the cast to INT seems to fix this issue. I'm running some long term.

I am running a longer test to confirm for sure.


alihacks:
Upon further testing, this doesn't really fix anyhting :)

PR #2730, opened 2020-12-28
Fix issue reported in #2721 - 2nd case in COALESCE

alihacks:
using last_request_start_time was missing 00 padding
I believe this fixes the issue in #2721 while still sorting long sessions as expected:

image


BrentOzar (approved):
Thanks! Merging into the dev branch, will be in the January release with credit to you in the release notes.

PR #2735, opened 2020-12-31
Encode column names for XML

snckirkmarken:
Encode column names for XML to prevent column names with characters that should be converted to entities from breaking the procedure.


snckirkmarken:
Fix is courtesy of help from @alihacks - thanks! 👍

BrentOzar (approved):
Nice work! And my condolences for having to deal with a database structure like that, hahaha. Merging into the dev branch, will be in the January release with credit to you in the release notes.

Issue #2739, opened 2021-01-07
sp_BlitzFirst - Add "Running CheckID" messages when @Debug = 1


Adedba:
**Is your feature request related to a problem? Please describe.**
Depending on the error, troubleshooting sp_BlitzFirst can be a little tricky as there is no output to advise which CheckID is currently running.

**Describe the solution you'd like**
Add RAISERROR WITH NOWAIT informational messages to advise which CheckID is currently running but only when @Debug = 1 

**Describe alternatives you've considered**
Cannot think of an alternative for this one.

**Are you ready to build the code for the feature?**
yep!

BrentOzar:
Great idea! Love it. Thanks for the code. Merging into the dev branch, will be in the January release with credit to you in the release notes.

PR #2740, opened 2021-01-07
2739 sp blitz first   add running check id messages when @debug = 1

Adedba:
#2739 Added Running Check ID messages for Debug = 1.
Added BEGIN and END where missing for various checks where I added RAISERROR commands.

Forced execution of check ID 42 and 43 as the are executed within sp_executesql , tested ok.


BrentOzar (approved):
Great idea! Love it. Thanks for the code. Merging into the dev branch, will be in the January release with credit to you in the release notes.

PR #2747, opened 2021-01-10
Update sp_ineachdb.sql

sm8680:
Line 194 changed 3 occurrences of variable @nospaces to @NoSpaces


BrentOzar:
Awesome, thanks! Just accepting these as-is - I don't have a good testbed for sp_ineachdb. Thanks!

PR #2751, opened 2021-01-14
sp_BlitzLock: Add error check for update stats on temp table, skip if fails

chadbaldwin:
When user runs without sysadmin or db_owner permissions sp_BlitzLock fails reporting '#t' cannot be found. Added try/catch
to skip line if fails for 1088 error. If any other error is returned, it throws as normal.

Tested on SQL 2014/2019.

Assumed OK to use THROW since proc is locked to 2012+ anyway.


BrentOzar (approved):
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the January release with credit to you in the release notes.

BrentOzar:
Ah, dang it. The throw is okay, but ... it breaks the installation scripts on SQL Server 2008. I'm going to leave it in, but just update the documentation. Doh!

PR #2761, opened 2021-01-18
Todd Chittenden

ToddChitt:
Looking through the sp_BlitzIndex result finding, in particular the Duplicate Index group, results are ORDERED BY object name? Why?
From a work-flow perspective, if someone wanted to go after the low-hanging fruit of duplicate indexes, those on larger tables would yield the best bang for the buck. (Yes, I know that's not always 100% true and there might be better metrics to determine the *BEST* candidates for removal, but total_rows or  total_reserved_MB are readily available.)

If you agree, that's great. If not, well, I tried :)
If you REALLY agree and think other results sub-sets in Blitz Index should be sorted differently as well, let me know and I would be glad to help.
Blitz Index Sorted

Thanks again for a great set of tools!


BrentOzar:
That's a great idea! I love it.

The pull request doesn't work because a lot of files have been changed by other folks in the meantime. I'm going to do a separate pull request and take your changed line of code to make it work, but going forward, here's how to make it work well:

First, create an issue in this repo describing what you want to do (so other people can give their feedback, because sometimes they have great ideas too.)

After there's a discussion, and before you start coding, sync your local copy of the First Responder Kit with the upstream remote branch - ours. Read the "pulling to your local branch from the remote" section here: https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/syncing-your-branch It sounds hard, but it's really easy, just one button.

If there are conflicts, that means both you and other people have changed the same lines. That's why it's important to pull the remote changes before you start making any edits - that way, your code will be able to sync in nicely.

Then make your changes and do a pull request. It's totally okay if some time lapses between when you pull the remote changes and when you submit yours, but the longer you wait, the more likely it is that someone else has changed the same lines of code. It's super rare on a project like the First Responder Kit, though.

Thanks again - your change will be in the February release, and I'll credit you in the release notes.

Issue #1049, opened 2017-07-25
sp_Blitz -- Missing BEGIN and END in memory dumps check


BlitzErik:
Missing block after the `IF` to check if there are 5 or more memory dumps.

Issue #100, opened 2015-01-09
BlitzCache export to table produces incorrect results


peschkaj:
From Gail Jurey Gail.Jurey@hobartservice.com

```
Exec sp_blitzcache
@top = 10
--,@sortorder = ‘Duration’
```

Results: returns 10 records (in duration order if I uncomment the @sortorder line)

```
Exec sp_blitzcache 
@top = 10
----,@sortorder = ‘Duration’
,@output_database_name = ‘ABC’
,@output_schema_name = ‘dbo’
,@output_table_name = ‘XYZ
```

PR #2765, opened 2021-01-21
Update sp_Blitz.sql

ToddChitt:
Added "Best Practices" for remote admin connection config option when the [value_in_use] = 1
Check ID #100:
Priority = 250 (vice 170)
Findings Group = "Best Practices" (vice "Reliability")

Thanks for the guidance. 

Side note: I think you can remove CheckID #1049 now as this new logic will look at both cases of the config value. Also, it might be sending mixed signals to state from 1049 that you have a "Non-Default config value" then here state that you have followed best practices by doing so. Hence my statement of "DELETE FROM #BlitzResults WHERE CheckID = 1049;"

IMHO, "Non-Default config values" findings should be reserved for those cases where changing such a default is usually a bad idea.

Blitz Best Practice

Thanks again for this awesome tool!

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -3251,13 +3251,16 @@ AS
 										AND j.category_id <> 100; /* Exclude SSRS category */
 					END;
 
+				/* Dedicated Admin Connection */
 				IF EXISTS ( SELECT  1
 							FROM    sys.configurations
 							WHERE   name = 'remote admin connections'
-									AND value_in_use = 0 )
+									--AND value_in_use = 0 /* No filter, we need to consider both value cases for this option. */
+									)
 					AND NOT EXISTS ( SELECT 1
 									 FROM   #SkipChecks
 									 WHERE  DatabaseName IS NULL AND CheckID = 100 )
+				
 					BEGIN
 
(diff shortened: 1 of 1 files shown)


BrentOzar:
OK, Todd, last time I want to have this discussion.

Before you write code, I need you to start an issue explaining what you want to do, and get consensus from others first. I don't want to break your heart by denying another pull request.

This pull request has a great heart, but...part of the aim of sp_Blitz is that we don't want to return any information if there's no action that needs to be taken on something. Imagine getting a daily email that tells you, "The sky is blue today. The ocean is deep today. Lions are dangerous." If you got that same email every day, you'd start setting up an email filter to put it into a folder and discard it.

In sp_Blitz, we only want to return a line if there's an action that the user needs to take, or needs to be aware of because something has been changed away from its defaults. If we give them a report full of things saying "no action is required here," then they'll get less value out of sp_Blitz.

That's why I asked you to look at the warnings for Cost Threshold and MAXDOP before designing anything. In the case of MAXDOP, for example, if the server only has a few cores, we don't bother telling them anything because there's no action that they need to take.

Here in this case, if you're alerting them that remote access to the DAC has been enabled and that it's a best practice...just don't alert them at all. They're fine. Move on.

I would totally take a pull request like that, but I just want to avoid alerting them when there's no action to be taken.

PR #2777, opened 2021-01-27
Check for >20% runnable queries

### LINKED ISSUE #2776: sp_BlitzFirst: Warn about a high percentage of runnable sessions
Queries can be running, runnable, or suspended.

Running means chugging happily along, suspended means they're waiting on _something else_, and runnable means they're waiting on a CPU. 

If a lot of queries are runnable (waiting on CPU), it might be a sign that you encountered CPU pressure during a sampled run, using the `@Seconds` parameter.

A warning in the headline news section if either the first or second passes encounter a high number of runnable queries compared to active requests in `sys.dm_exec_requests`. 

High is debatable, I guess, but I'd probably start at 20-25% of sessions waiting on a CPU to fire the warning. It can be handy when working on systems that have many active queries. 

Confronting an alligator.

Yes, I'll also be adding this to [one of my own things](https://github.com/erikdarlingdata/DarlingData/tree/master/sp_PressureDetector), too. 

Something like this should be a good enough start:

```
SELECT 
    1 AS pass,
    x.*,
    CONVERT(decimal(5,2),
        (
            x.runnable / 
                (1. * NULLIF(x.total, 0))
        )
    ) * 100. AS runnable_pct
FROM 
(
    SELECT 
        COUNT_BIG(*) AS total, 
        SUM(CASE WHEN status = 'runnable' THEN 1 ELSE 0 END) AS runnable
    FROM sys.dm_exec_requests
    WHERE session_id > 50
) AS x;
```

Sample output:
image

erikdarlingdata:
Updated documentation and sp_BlitzFirst. Blog post is scheduled for February 4th, so that link won't work yet.

Closes #2776

### DIFF
--- a/Documentation/sp_BlitzFirst_Checks_by_Priority.md
+++ b/Documentation/sp_BlitzFirst_Checks_by_Priority.md
@@ -6,8 +6,8 @@ Before adding a new check, make sure to add a Github issue for it first, and hav
 
 If you want to change anything about a check - the priority, finding, URL, or ID - open a Github issue first. The relevant scripts have to be updated too.
 
-CURRENT HIGH CHECKID: 46
-If you want to add a new check, start at 47
+CURRENT HIGH CHECKID: 47
+If you want to add a new check, start at 48
 
 | Priority | FindingsGroup | Finding | URL | CheckID |
 |----------|---------------------------------|---------------------------------------|-------------------------------------------------|----------|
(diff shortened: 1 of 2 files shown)


BrentOzar (approved):
Ooo, I love how you did it on both the first and second pass, very sharp.

BrentOzar:
Nice work! Love it. Merging into the dev branch, will be in the Feb release with credit to you in the release notes.

PR #2788, opened 2021-02-13
added comments around xp_cmdshell since some firewalls block this string

tlovie:
Some firewalls specifically look for the string "xp_cmdshell" and terminate the connection.  Adding comments directly beside like "/**/xp_cmdshell/**/" seems to be enough to fool this behavior.


BrentOzar (approved):
Laughing. If this is what you need, then let's do it. Merging into the dev branch, will be in the Feb release with credit to you in the release notes.

Issue #2787, opened 2021-02-12
sp_Blitz* - Treat "DBATools" like a system database


erikdarlingdata:
**Is your feature request related to a problem? Please describe.**
In a few scripts, we filter out system databases. This can help avoid seeing/processing irrelevant information.

**Describe the solution you'd like**
Starting with "DBATools", I'd like to go through scripts and add common admin database names (DBAdmin would be another candidate) to the screen out list. 

**Describe alternatives you've considered**
A caboose but backwards and made of jelly.

**Are you ready to build the code for the feature?**
Yes, this will be a FUN WEEKEND ACTIVITY.

For the children.

BrentOzar:
Love it.

BrentOzar:
@Iomckee - Woohoo! Looks good, thanks for making those changes. Merging into the dev branch, will be in the March release with credit to you in the release notes. Nice catch on the documentation, too.

Other readers - this was only implemented on sp_BlitzCache, just FYI.

jeffchulg:
Hi, would it be possible to add a custom comma-separated list as an input to these procedures, in case one database is not listed in the changes above? This could be used as an exclusion list at the same time...

erikdarlingdata:
@jeffchulg there's an `@IgnoreDatabases` parameter to `sp_BlitzIndex` if you want to use that as a template for other procs.

jeffchulg:
> 
> 
> @jeffchulg there's an `@IgnoreDatabases` parameter to `sp_BlitzIndex` if you want to use that as a template for other procs.

Sounds great! :)

BrentOzar:
Thanks for the pull request! Merging into the dev branch, will be in the March release with a shout-out to you in the release notes.

PR #2806, opened 2021-03-02
Blitzcache: IgnoreSystemDBs ignore common admin dbs

lomckee:
As per #2787 , excludes common admin database names when invoked with `@IgnoreSystemDBs = 1`.

Database names included (actually all lower case, if they have case sensitive database it won't get picked up in the check):
- DBMaintenance
- DBATools
- DBAdmin

I've updated BlitzCache as well as the install scripts, not sure if they're compiled or you just update both in parallel so :shrug:
Tested against database named `DBMaintenance` with BlitzCache and Install-All scripts.

Please see screenshots of command being invoked after update to script.
BlitzCache_IgnoreSystemDbsExplicitExcludesMaintenance
BlitzCache_IgnoreSystemDBsImplicitOffExcludesMaintenance
BlitzCache_IgnoreSystemDBsOffIncludesMaintenance

### DIFF
--- a/Install-All-Scripts.sql
+++ b/Install-All-Scripts.sql
@@ -14191,7 +14191,7 @@ IF @Help = 1
 	UNION ALL
 	SELECT N'@IgnoreSystemDBs',
 			N'BIT',
-			N'Ignores plans found in the system databases (master, model, msdb, tempdb, dbmaintenance, dbadmin, dbatools, and resourcedb)'
+			N'Ignores plans found in the system databases (master, model, msdb, tempdb, and resourcedb)'
 
 	UNION ALL
 	SELECT N'@OnlyQueryHashes',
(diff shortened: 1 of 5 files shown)


BrentOzar (changes requested):
I like where you were going with this, but one problem: on servers with case-sensitive databases, the database names are case-sensitive, so you'll want to use a LOWER on the filter in order to make it work. (This is easier with master/model/msdb/tempdb because the names are all lower case.) Also, you don't have to edit the Install% scripts - those are auto-generated at release time. Thanks!

lomckee:
Hi @BrentOzar,

Thank you for your feedback. I have made the recommended changes.

I also applied LOWER to `CAST(pa.value AS sysname)`. Truth be told I'm not sure what values this can be but I figured because it was in the same COALESCE function as the DB Name it seemed right.

I have modified a user database in my SQL Server to be case sensitive and preformed the above tests. I'm not sure if that's sufficient coverage for you? Let me know if not.

BrentOzar (approved):
Woohoo! Looks good, thanks for making those changes. Merging into the dev branch, will be in the March release with credit to you in the release notes.

BrentOzar:
Woohoo! Looks good, thanks for making those changes. Merging into the dev branch, will be in the March release with credit to you in the release notes. Nice catch on the documentation, too.

Issue #2803, opened 2021-02-28
Need to rationalize 'brentozar.com'


JohnDBallentineIII:
**Version of the script**
8.01

**What is the current behavior?**
Seen in the code directory - 
'http://www.brentozar.com' - 36 times
'https://www.brentozar.com' - 492 times
'https://brentozar.com' - 12 times
'http://brentozar.com' - 189

**If the current behavior is a bug, please provide the steps to reproduce.**
Stands out when looking at a column of results - sometimes as www.brentozar,com , sometimes as just brentozar.com .

**What is the expected behavior?**
Should be standardized.  https://www.brentozar.com seems to be the most frequently used, and what the web site goes to.  Others should be changed to this standard.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
N/A

BrentOzar:
Thanks for the issue. Sure, if you want to work on this, read through the contributing guide here:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/main/CONTRIBUTING.md

BrentOzar:
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the March release with credit to you in the release notes.

PR #2808, opened 2021-03-03
Rationalized brentozar.com

JohnDBallentineIII:
Fix for issue 'Need to rationalize 'brentozar.com' #2803'


BrentOzar (changes requested):
Hi! Thanks for the pull request, and I love where you're going with this. However, don't modify the Install% scripts - those are auto-generated during the release process. I would just accept this pull request, but I know that's going to break the release process - just edit the sp_Blitz% type scripts and documentation files, but avoid the Install% scripts. Thanks!

BrentOzar (approved):
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the March release with credit to you in the release notes.

Issue #2738, opened 2021-01-05
sp_BlitzFirst - statistics update insert fails, possibly related to special characters in database schema


jeffchulg:
**Version of the script**
Look in the stored procedure, and it'll have a version date & number near the top. Put that in here. If it's not the current version (dated in the last month), then upgrade to the current version and test that before reporting a bug - we fix a lot of stuff in each new build. We'll flat out close bug reports for older builds.
20201211 

**What is the current behavior?**

On SQL Server 2016+, I get following errors:
```

Running Check #044
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
The select list for the INSERT statement contains fewer items than the insert list. The number of SELECT values must match the number of INSERT columns.
Msg 120, Level 15, State 1, Line 34
[...]

PR #2811, opened 2021-03-05
Updates sp_BlitzFirst - Solves #2738

jeffchulg:
Resolution for issue #2738

Change log:

splitted command `sp_MSforeachdb <string>` to string creation then call to `sp_MSforeachdb` with variable


BrentOzar:
Thanks for the pull request! Merged into the dev branch, will be in the March release with credit to you.

Issue #2805, opened 2021-03-01
sp_DatabaseRestore - Full Text Catalog can not be moved to another location


fvanderhaegen:
**Version of the script**
Version: 8.01
VersionDate: 20210222

**What is the current behavior?**
When you want to move the files to another location during the restore the Full Text Catalog is ommitted from the restore command.

**If the current behavior is a bug, please provide the steps to reproduce.**
Restore a backupfile containing Full Text Catalog (backup type=F) and move the files to another location:

```
EXEC sp_DatabaseRestore @database=N'test', 
@backuppathfull=N'Full_Backup_Location',
@moveFiles=1,
@movedatadrive=N'new_location';
```

**What is the expected behavior?**
All files in the backupfile should be restored

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Version SQL Server: SQL Server 2019 Standard
OS: Windows 2019

I don't know if this worked in previous versions

BrentOzar:
Sorry about that, and thanks for the issue. I don't personally use this proc - was this a bug you plan to fix yourself, or are you hoping someone else can code it for you? If you're looking to have someone else code it, you may want to pop into the Slack room to see if you can find someone else willing to help out. (Just setting expectations.) Thanks!

fvanderhaegen:
I've adjusted the code and it's ready to be tested

BrentOzar:
OK, when you're ready, do a pull request. Here's a great tutorial by Rob Sewell: https://sqldbawithabeard.com/2019/11/29/how-to-fork-a-github-repository-and-contribute-to-an-open-source-project/

fvanderhaegen:
The pull request has been created.

BrentOzar:
Thanks for the pull request! Merged into the dev branch, will be in the March release with credit to you.

PR #2815, opened 2021-03-11
#2805 sp_DataBaseRestore FullText Catalog restore

fvanderhaegen:
Added option to move FullText Catalog files during restore


BrentOzar:
Thanks for the pull request! Merged into the dev branch, will be in the March release with credit to you.

PR #2832, opened 2021-03-21
Closes #2831

### LINKED ISSUE #2831: sp_BlitzLock: error "invalid column name dn"
**Version of the script**
SELECT @Version = '8.02', @VersionDate = '20210321';

Executing on a server with deadlocks:

```
Msg 207, Level 16, State 1, Procedure sp_BlitzLock, Line 1501 [Batch Start Line 0]
Invalid column name 'dn'.
```

It worked before this pull request: https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/pull/2825

And it's because we took out the rownumber column aliased as dn, but then we refer to it in the where clause:

Screen Shot 2021-03-21 at 7 52 38 AM

We either need to put the rownumber back in, or remove the dn filter from the where clause, researching.

erikdarlingdata:
dn is back!

You uh, don't need to list that I fixed the bug I introduced in the liner notes. I'm fine with this one going uncredited 😃


BrentOzar:
HAHAHA, deal.

Issue #2823, opened 2021-03-15
sp_BlitzFirst: add @ExportToExcel parameter like sp_BlitzCache has


tmvtmv:
**Version of the script**
sp_BlitzIndex(TM) v8.01 - February 22, 2021

**What is the current behavior?**
The Window with the Findings and Priority doesn't show nicely as columns when copy/pasted to Excel.
![Screenshot 2021-03-14 at 12 15 59](https://user-images.githubusercontent.com/28952111/111121998-b20c1d80-856d-11eb-977e-3ddb4d83dd76.png)

**If the current behavior is a bug, please provide the steps to reproduce.**
* Run sp_BlitzFirst
* Copy with headers Window 2 (with headers Priority - Finding - URL -etc)

**What is the expected behavior?**
* Show up in columns as the other windows do.

Thanks!

BrentOzar:
Correct, Excel doesn't handle copy/pasting of XML columns.

If you'd like to get a copy/paste friendly version, that would be a new feature. Here's the Contributing Guide on how to get started proposing new features and coding them:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/main/CONTRIBUTING.md

In sp_BlitzCache, the way the contributor built it was to add an @ExportToExcel input parameter, defaulted to 0. When set to 1, it returns two sets of results: one set without the XML columns, and a second result set with the XML columns. That way you can choose which ones you'd like to copy/paste.

I'm going to close this issue, but if you want to tackle this by coding it yourself and doing a pull request, that'd be great! Just leave a note here and I'll assign it to you and reopen it.

tmvtmv:
Challenge accepted! :-)

BrentOzar:
Hahaha, ok, cool! If you have any questions, holler.

BrentOzar:
In case others are reading this, here was my note from the code review: Wait - so if all you need to do is set @OutputXMLasNVARCHAR = 1, then why do we need a new parameter and changes to the proc? Why can't you just run it with @OutputXMLasNVARCHAR = 1 and get what you need? (Just trying to minimize change and risk.)

tmvtmv:
Well ... there's that other catch: Excel only allows 32KB characters per column; I've came across output for the "HowToStopIt"-column up to 64KB which makes Excel divide the output over multiple rows = messy.

tmvtmv:
[...]

PR #2835, opened 2021-03-23
See Github #2823 - Added ReportToExcel feature

tmvtmv:
I hope this meets the quality standards ;-)
Input always welcome!


BrentOzar (commented):
Wait - so if all you need to do is set @OutputXMLasNVARCHAR = 1, then why do we need a new parameter and changes to the proc? Why can't you just run it with @OutputXMLasNVARCHAR = 1 and get what you need? (Just trying to minimize change and risk.)

PR #2853, opened 2021-04-01
Issue 2851:  sp_BlitzIndex - Only show Columnstore Index Visualisation details.

Djheath2000:
Issue 2851:  Amended sp_BlitzIndex to only show the columnstore index visualisation details of a table. 

- Added new parameter @ShowColumnStoreOnly bit, default value of 0.
- Setting parameter value to 1 will skip the majority of the "Step 3 - diagnose the patient" section of the procedure and only execute the Columnstore Index checks.
- Parameter is only checked if the @TableName is not null. Otherwise it's never used.
- If parameter is set to 1 and table does not have a columnstore index, then the existing no columnstore index information is shown, as normal, instead of the columnstore visualisation data.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -23,6 +23,7 @@ ALTER PROCEDURE dbo.sp_BlitzIndex
     @SkipPartitions BIT	= 0,
     @SkipStatistics BIT	= 1,
     @GetAllDatabases BIT = 0,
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the April release with credit to you in the release notes.

PR #2866, opened 2021-04-15
Fix #2865

### LINKED ISSUE #2685: sp_AllNightLog Install message when dbo.DatabaseBackup is in custom database for DBAAdminobjects
**Version of the script**
3.999
I get this message:
The module 'sp_AllNightLog' depends on the missing object 'master.dbo.DatabaseBackup'. The module will still be created; however, it cannot run successfully until the object exists

Install Ola SQL Server Maintenance Solution in different database than master.
https://ola.hallengren.com/frequently-asked-questions.html

"Do the objects need to be created in the master database?
The objects can be created in any database: master, msdb, or a utility database for database administrative scripts. You just need to change this line of code in MaintenanceSolution.sql:

USE [master]

to point to the database in which the objects will be created. The jobs will automatically be configured to run against this database."

Should look in the database that is being installed first, then give the error.  Checks start at line 167 and are hardcoded to Master
SELECT COUNT(*) FROM master.sys.procedures WHERE name IN('CommandExecute', 'DatabaseBackup')
NA

### LINKED ISSUE #2865: sp_Blitz: string or binary data truncation on #driveinfo table
**Version of the script**
2021-03-22

The `#driveinfo` table and queries are borked, because:
* The logical volume name column is defined as nvarchar(32) because that's the limit, but
* The query that updates the table adds two parenthesis to the volume name, and
* The first parenthesis has a space in front of it, so
* If a logical volume name is 32 characters long, it throws a string/binary truncation error

That, up there ☝👆

Work correctly?

All of them

erikdarlingdata:
We don't need no string truncation

Closes #2865

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -634,8 +634,8 @@ AS
 			DROP TABLE #driveInfo;
 		CREATE TABLE #driveInfo
 			(
-			  drive NVARCHAR,
-              logical_volume_name NVARCHAR(32), --Limit is 32 for NTFS, 11 for FAT
+			  drive NVARCHAR(2),
+              logical_volume_name NVARCHAR(36), --Limit is 32 for NTFS, 11 for FAT
 			  available_MB DECIMAL(18, 0),
               total_MB DECIMAL(18, 0),
               used_percent DECIMAL(18, 2)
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Ooo, nice catch!

Issue #2870, opened 2021-04-20
sp_BlitzWho - Failing on case sensitive instances


Adedba:
**Version of the script**
SELECT @Version = '8.03', @VersionDate = '20210420'

**What is the current behavior?**
Error when running sp_BlitzWho or logging to a table 

Checking for Read intent databases to exclude
Msg 207, Level 16, State 1, Line 161
Invalid column name 'session_id'.

**If the current behavior is a bug, please provide the steps to reproduce.**
Using a case sensitive SQL instance run:

`EXEC sp_BlitzWho `

or

`EXEC sp_BlitzWho 
	@OutputDatabaseName = 'DBA',
	@OutputSchemaName= 'dbo',
	@OutputTableName = 'BlitzWho';`

**What is the expected behavior?**
No errors

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Tested with SQL 2016 only at this stage.
Worked fine with the previous version

I am happy to pick this up.

BrentOzar:
Good catch! Merging into dev, will do another release with this just to head off problems.

Adedba:
Man that was fast! Thanks Brent!

BrentOzar:
Thank YOU, sir!

PR #2871, opened 2021-04-20
#2870 sp_BlitzWho - Failing on case sensitive instances

Adedba:
#2870 Corrected casing for columns:
Query_Plan
Session_Id
cached_parameter_info


BrentOzar (approved):
Good catch! Merging into dev, will do another release with this just to head off problems.

PR #2884, opened 2021-05-04
Update sp_BlitzIndex.sql

### LINKED ISSUE #2883: sp_BlitzIndex: missing indexes Sample Query Plan is for the wrong query
**Version of the script**
8.03 (20210420)

When running `EXEC sp_BlitzIndex @TableName='MyTable'` I get missing indexes reported with the new Sample Query Plan column, however the plan loaded in there doesn't reference 'MyTable' at all. 

The Sample Query Plan shown does have 1 missing index request, but it's for a different table.

I haven't been able to repro this in an environment where I can share screenshots from. Hopefully the above makes sense.

The Sample Query Plan loaded in the column would have the missing index request.

SQL 2019 CU9. I notice that there's some comments around SQL 2019 in the commit history, maybe it's just not supported yet?

gdoddsy:
Filter the list of missing index query plans to just the missing index for this record.

Tries to fix #2883 
**This has about 1 minute of testing on the server that I'm on. It fixes the problem this time, but I'd suggest it needs a fair bit more testing. I'm not going to have time to do that properly anytime soon.**

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1674,6 +1674,7 @@ BEGIN TRY
 				  	  ORDER BY (q.user_seeks + q.user_scans) DESC, s.total_logical_reads DESC
 			      ) q2
 				  CROSS APPLY sys.dm_exec_query_plan(q2.plan_handle) p
+                  WHERE ig.index_group_handle = gs.group_handle
 			  ) '
 		END
         


BrentOzar:
Awesome, thanks sir. I'll do a round of testing on that this coming week.

BrentOzar:
I've confirmed that this one is working for me much better than the old one was. I think there are probably still some edge cases where it's not going to work, but in the interest of keeping the thing moving forward in each version, I'll roll with this as-is. Thanks sir!

PR #2888, opened 2021-05-06
Update sp_Blitz.sql with minor fixes/enhancements for Amazon RDS

rrankins:
Some changes/enhancements to sp_Blitz that I found helped provide more information when running against Amazon RDS SQL Server, as well as an enhancement to check if Instant File Initialization is enabled for all versions of SQL Server. 
Changes include:
– added comments to RDS skipped checks to explain what they are and why they are skipped.
- modified code to run rdsadmin.dbo.rds_read_error_log instead of xp_readerrorlog when it’s running on an RDS server 
- commented out Insert to skip check 193 since the above code change allows it to run
- added additional query against sys.dm_server_services to confirm if instant_file_initialization is enabled if it can’t find it in the errorlog (possibly due to cycling of the errorlog)


BrentOzar (approved):
Looks good! Thanks for the pull request, merging into the dev branch. Will be in the May release with credit to you in the release notes.

PR #2894, opened 2021-05-10
Uninstall.sql - Treat the table SqlServerVersions like the procedures

### LINKED ISSUE #2893: Uninstall.sql: Drops table dbo.SqlServerVersions: Only in current database / Regardless of @printOnly / Not part of PRINT
**Version of the script**
Current dev-branch at GitHub

The table dbo.SqlServerVersions is only dropped in current database
The table dbo.SqlServerVersions is dropped regardless of @printOnly
The drop statement is not included in the PRINT output

Just have a look at the code:
https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/0e76863536bcd42c8f7f2292711099c43f80cae7/Uninstall.sql#L76-L77

In the first case (only current database) I am not sure, because it may be risky to delete all tables dbo.SqlServerVersions in all databases.
In the second case (regardless of @printOnly) I would expect that @printOnly does not delete any objects, so does not drop the table.
In the third case (PRINT output) I would expect the DROP TABLE to be included in the PRINT. This is how I found the bug: I have missed the statement and so I had a look at the code.

If you want me to code that and open a pull request, just give me a hint.

andreasjordan:
Do not drop the table in any case but treat the table like the procedures.
Closes #2893

### DIFF
--- a/Uninstall.sql
+++ b/Uninstall.sql
@@ -39,6 +39,10 @@ BEGIN
     FROM sys.procedures P
     JOIN #ToDelete D ON D.ProcedureName = P.name;
 
+    SELECT @SQL += N'DROP TABLE dbo.SqlServerVersions;' + CHAR(10)
+    FROM sys.tables 
+    WHERE schema_id = 1 AND name = 'SqlServerVersions';
+
 END
 ELSE
 BEGIN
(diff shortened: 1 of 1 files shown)


BrentOzar (changes requested):
Thanks, but this doesn't work in Azure SQL DB because you're not allowed to use the USE command. We're trying not to add things that break Azure SQL DB if we can avoid it. Can you check to see if you're in Azure SQL DB, and only use the USE command if you're not?

andreasjordan:
I don't see the point yet - because my new "USE" is only used when there is also a "USE" for the DROP PROCEDURE just a few line above in the same code block.
But I will have a closer look and test on Azure this evening after work, the todays client is waiting...

BrentOzar:
OH! Great catch, I didn't see that. I apologize! If the existing script doesn't work in Azure, then it's fine. I'm only worried about breaking scripts that currently exist in Azure. I'll merge this. Thanks!

PR #2896, opened 2021-05-11
Add recent CUs

BrentOzar:


Ant-Green:
@BrentOzar , the add for CU17 for 2016 has the wrong build number, it should be 5888, do you want to revert the commit and re-push or would you like me to fix and put in a new PR for it?

BrentOzar:
Thanks for the catch! You can put in a new PR. Thanks!

PR #2899, opened 2021-05-14
Update sp_databaseRestore to correctly get Diff File to skip log files correctly

### LINKED ISSUE #2898: @StopAt doesn't work when also specifying @OnlyLogsAfter
**Version of the script**
@Version = '8.03', @VersionDate = '20210420';

When specifying both the @StopAt and the @OnlyLogsAfter parameters, the @StopAt doesn't work. We want to use both as we keep 90 days of log backups and don't want to have to check each one to see if it needs to be restored.

```
EXEC sp_DatabaseRestore @Database = N'StackOverflow'																	-- nvarchar(128)
						,@RestoreDatabaseName = N'Redebiz_20210502'																	-- nvarchar(128)
						,@BackupPathFull = N'c:\backups\FULL\'
						,@BackupPathDiff = N'c:\backups\DIFF\'
						,@BackupPathLog = N'c:\backups\Log\'
						,@RunCheckDB = 0		
						,@RestoreDiff = 1		
						,@RunRecovery = 1	
						,@StopAt = N'20210502061500'
						,@OnlyLogsAfter = N'20210501181415'
						,@DatabaseOwner = 'sa'
						,@Execute='N'
						,@Debug=1
```

Note that it checks every single log file, not just those between 1st May 2021 6:15pm and 2nd May 2021 6:15am

That only the logs I want are checked

SQL 2019. It broke when #2180 was merged in and the cursor was removed

I'll have a pull request shortly to deal with this. I think I can simplify the bit of code that is breaking as well.

gdoddsy:
Update sp_databaseRestore to correctly get Diff File to restore and to skip log files that need to be skipped

Fixes #2898 

I think I've cleaned up the logic to make it simpler in the future. As always, it works on my machine, but if someone could check over to make sure I haven't broken the functionality, that'd be great.

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -998,7 +998,11 @@ BEGIN
 	END
     /*End folder sanity check*/
     -- Find latest diff backup 
-    SELECT @LastDiffBackup = MAX(BackupFile)
+	SELECT @LastDiffBackup = MAX(CASE
+		WHEN @StopAt IS NULL OR REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 16 ), '_', '' ) <= @StopAt THEN
+			BackupFile
+		ELSE ''
+		END)
     FROM @FileList
     WHERE BackupFile LIKE N'%.bak'
         AND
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks for the pull request! Merged into the dev branch, will be in the May release with credit to you in the release notes.

PR #2901, opened 2021-05-15
Fixes #2900

### LINKED ISSUE #2900: sp_BlitzCache: plan count bug (reported on dba.se)
**Version of the script**
The latest, according to this [very helpful user with phenomenal code formatting](https://dba.stackexchange.com/a/291612/32281)

We count the wrong thing

Click the link above for more details. It's free.

Count the right thing

Any and all.

erikdarlingdata:
Counting plans should count plan hashes and not query hashes. We should also count items from the base view and not `##BlitzCacheProcs`

Closes #2900


BrentOzar (approved):
Using the tweaked version on query_hash for now, but I'm open with further changes like I mentioned in the ticket - I think we're gonna end up wanting distinct plans too.

Issue #2880, opened 2021-04-30
sp_BlitzIndex: Inflated columns sizes for included NVARCHAR columns


holidasa:
**Version of the script**
`SELECT @Version = '8.03', @VersionDate = '20210420';`

**What is the current behavior?**
This is similar to #2285 except for the column(s) in the INCLUDE definition of an index.
NCHAR and NVARCHAR columns in the INCLUDE display the storage size.

**If the current behavior is a bug, please provide the steps to reproduce.**
Using the 2018 StackOverflow database, execute:
`EXECUTE sp_BlitzIndex @DatabaseName = N'StackOverflow', @TableName = N'Users';`
DisplayName is defined as `NVARCHAR(40)` but when it appears in the INCLUDE, the Definition column displays `nvarchar 80`
https://imgur.com/g1QMpEq

**What is the expected behavior?**
Consistency among key and include column display.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
SQL Server 2019 on Windows

This...
`
                        /* fix start */
                        + N' {' + system_type_name +
                        CASE max_length WHEN -1 THEN N' (max)' ELSE
                            CASE
                                WHEN system_type_name IN (N'char', N'varchar', N'binary', N'varbinary') THEN N' (' + CAST(max_length AS NVARCHAR(20)) + N')'
                                WHEN system_type_name IN (N'nchar', N'nvarchar') THEN N' (' + CAST(max_length / 2 AS NVARCHAR(20)) + N')'
                            ELSE ''
                            END
                        END
                        +  N'}'
                        /* fix end */
`
should replace this
`+ N' {' + system_type_name + N' ' + CAST(max_length AS NVARCHAR(50)) +  N'}'`

around line 2210 as of April 2021

I'm just saying that the same case statement can be applied to the D3 alias.  Contact me if you have any questions.

I've read the T-SQL and know exactly what to change.  I'm now reading about GIT and to save you time.

BrentOzar:
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the May release with credit to you in the release notes.

PR #2906, opened 2021-05-17
Update sp_BlitzIndex for #2880

holidasa:
This is for bug #2880 for sp_BlitzIndex.

Modified the char-based INCLUDED column definition to match how char-based key columns are displayed (i.e. {nvarchar 80} versus {nvarchar (40)}).

Added max size for non-char-based columns (i.e. {bigint 8}) for key, include, and secret columns.


BrentOzar (approved):
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the May release with credit to you in the release notes.

Issue #2913, opened 2021-05-22
sp_BlitzFirst - Server Uptime vs Query Runtime


erikdarlingdata:
**Is your feature request related to a problem? Please describe.**
When looking at wait stats, the relationship between total wait time and server uptime can be useful to highlight when contention is constant. 

However, for servers that aren't constantly slammed or sometimes look bored, it can be harder to justify certain things like adding hardware. For that, you really need to catch contention in the act.

**Describe the solution you'd like**
I'd like to take a stab at figuring how much query runtime there is since the server has been started.. To do this, we need to sum `wait_time` from `dm_os_wait_stats` and `total_cpu_usage` from `dm_os_schedulers` for visible online cores only. I also need to rule out a number of waits, and that list can come from the waits temp table we already populate.

A starting point for this query [is here](https://gist.github.com/erikdarlingdata/c96f6093a3eb8bbb3cc3042f0d3599ae), though I'm on the fence about `SLEEP_TASK` being ignored. It can crop up during hash spills and other unpleasantness.

**Describe alternatives you've considered**
Not day drinking

**Are you ready to build the code for the feature?**
Obviously, if people out there think this is a feasible and useful metric for comparison.

Thanks to Sean Gallardy and Paul White for the sanity checks on the starting point query, of course.

erikdarlingdata:
I've made a couple updates to the script after testing to
* Using a view for portability/testing
* Remove signal wait time, since it's already included in wait time (thanks Paul)
* Format results a little more nicely

I've been testing this with various queries, and the results seem to line up accurately. I was thrown off a bit at first by testing with parallel queries, but since "waits in parallel plans accumulate for each thread, it makes sense to compare with total thread time"

BrentOzar:
This is a brilliant, brilliant idea.

erikdarlingdata:
I have this working for the `@SinceStartup` and `@Seconds` parameters, but I haven't implemented it in the logging portions because I don't want to introduce breaking changes until if/when things are finalized. There's a pull request available if anyone wants to test/tinker.

![image](https://user-images.githubusercontent.com/2136037/120495727-f8159800-c38a-11eb-90ed-9b4a5666cb6c.png)

erikdarlingdata:
@BrentOzar no rush on this, but I bumped the ignore list I had up against the ignoreables in BlitzFirst, and found these differences:

[...]

PR #2924, opened 2021-06-02
Add thread time to wait stats views

erikdarlingdata:
Initial attempt!

Reference #2913


erikdarlingdata:
@BrentOzar Tweaked and tested on 2012 as well to make sure we don't do anything goofy and insert an informational message. This should be good to go.

BrentOzar (approved):
I've banged around on this for about an hour on SQL Server 2019, and I can't get it to break. Looks good! Merging into the dev branch now, and going to use it in this week's Mastering Server Tuning class to see if I spot issues during that.

Issue #2933, opened 2021-06-08
sp_BlitzIndex - Benefit per day on Azure


DavidWiseman:
**Version of the script**
8.04

**What is the current behavior?**
Estimated benefit per day numbers are very high when running on Azure DB.  @DaysUptime is calculated from the create date of tempdb - this doesn't work for azure and we end up defaulting to 0.01 days.

**If the current behavior is a bug, please provide the steps to reproduce.**
I'd consider this a minor bug. Run sp_BlitzIndex on an AzureDB - the uptime will report as 0.01 and the estimated benefit per day values will be inflated.

**What is the expected behavior?**
Return correct uptime / benefit per day (even if it's just a magic number)
Using the create_date of tempdb has long been used to determine instance uptime, but we now have this available in sys.dm_os_sys_info.  This is available since 2008 (All supported versions of SQL Server) and works for Azure DB.  Additionally, the documentation states to use sys.dm_os_sys_info to determine uptime.
https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-db-index-usage-stats-transact-sql?view=sql-server-ver15
This can always be coded to fall back on create_date from tempdb for older (now unsupported) SQL Instances.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
AzureDB

Note: PR for this coming soon.

BrentOzar:
Thanks for the pull request! Looks good, merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #2934, opened 2021-06-08
Fix @DaysUptime in sp_BlitzIndex for AzureDB

DavidWiseman:
Days uptime is now calculated from sys.dm_os_sys_info which works for AzureDB.  For older instances (<2008) we can fallback to create_date of tempdb.
#2933


BrentOzar (approved):

Issue #2941, opened 2021-07-01
sp_BlitzFirst - CheckID 1 (running backups) can show 2 backups running instead of 1


Adedba:
**Version of the script**
SELECT @Version = '8.04', @VersionDate = '20210530';

**What is the current behavior?**
If you have your backup routine code stored in a user database rather than master or you execute a backup command from a user database, when you take a backup you will get two warnings from sp_Blitzfirst stating you have two backups running. rather than 1.

Multiple entries for your spid will exist in the lookup against sys.dm_tran_locks, one for the database id of the database being backed up and one for the database id of the database context where the backup code is being executed from.

The above behaviour cannot be replicated in the master database.

**If the current behavior is a bug, please provide the steps to reproduce.**

Query window one: 
Use any user database and start taking a full backup of a reasonable size which will exceed the 5 minute threshold for sp_BlitzFirst

USE [DBA]; 
BACKUP DATABASE  [StackOverflow2013] 
TO DISK = '\\BACKUPSHARE\backups\StackOverflow2013.bak' 
WITH COMPRESSION,CHECKSUM;

Then wait 5 mins.

Query window two:
EXEC sp_BlitzFirst;

Observe the output for checkid 1 you will notice two rows for running backups, one for Stackoverflow2013 and one for the database context of where you executed the backup command from. Here is my example: 

![image](https://user-images.githubusercontent.com/33764798/124117290-d8e05800-da67-11eb-889f-88e4eaa2442d.png)

I see a potential fix here (it's not the prettiest): 

Around line 1453 ADD a cross apply to sys.dm_exec_sql_text and add an AND clause with a full wildcard on the sql text looking for the database name somewhere in the backup command., (assuming it's there and never shows as a parameter). 
 
```
 	CROSS APPLY sys.dm_exec_query_plan(r.plan_handle) pl
		CROSS APPLY sys.dm_exec_sql_text(r.sql_handle) st 
		WHERE r.command LIKE 'BACKUP%'
		AND  st.text LIKE N'%'+DB_NAME(resource_database_id)+'%'
		AND r.start_time <= DATEADD(minute, -5, GETDATE())
```
		

alternative would be to remove the join below but this is used for some useful information around database size so not favourable:

```
		INNER JOIN (
		SELECT DISTINCT request_session_id, resource_database_id
		FROM    sys.dm_tran_locks
		WHERE resource_type = N'DATABASE'
		AND     request_mode = N'S'
[...]

PR #2942, opened 2021-07-01
#2941 - sp_BlitzFirst Addresses incorrect number of backups running for CheckID 1

Adedba:
#2941 - Added additional join clause to Check 1 so that the correct amount of backups running is shown when the database context of the backup command is ran from a user database.


BrentOzar:
Thanks for the pull request! Good catch. Merging into the dev branch, will be in the July release with credit to you in the release notes.

PR #2945, opened 2021-07-09
add target recovery interval check

### LINKED ISSUE #2938: sp_Blitz: Warn when databases have <> 60 for Target Recovery Interval
On servers where there are a lot of flushcache messages in the error log, changing the [target recovery internal from 0 to 60](https://sqlperformance.com/2020/05/system-configuration/0-to-60-switching-to-indirect-checkpoints) (the default in databases created on 2016+) can really help alleviate stress on the server in several ways.

Right now, having a non-zero value gets called out as a non-default setting. I think that should be taken out, and we should warn when people use 0 instead. Over the last couple years, I've had people bump that setting and have a lot of oddball problems disappear. 

Cataloging fetishes

Yes, but this may need some tweaking for cloud places where the setting may not be available for change. I have not done research in this area.

erikdarlingdata:
Adds a check when available (2012+) to see if target recovery interval is <> 60, and suggest setting it there.

Sure, it could be 61 or something, but I'm not going down that path.

I quoted out the current check for target recovery interval that looks to see if it's not 0 or 60, to not give seemingly conflicting advice.

Right now, the URL is to Aaron's post on the subject. @BrentOzar will need to go code it for uniformity.

I've also updated the documentation to reflect the new high check id, etc.

Closes #2938


BrentOzar (approved):

PR #2949, opened 2021-07-14
Fixed SkipChecks bug caused by collation mismatch

DavidSchanzer:
In the INSERT INTO #SkipChecks dynamic SQL statement, the WHERE clause includes:
    ServerName = SERVERPROPERTY('ServerName')

This is false when the SkipChecks table is in a (linked) server with a different collation to the server on which sp_Blitz is being executed, because ServerName returns data in one collation (eg. Latin1_General_CI_AS) whereas SERVERPROPERTY('ServerName') returns data in (I believe) the current server or database collation (eg. SQL_Latin1_General_CP1_CI_AS).

Therefore, there is a need to include a COLLATE clause so that equality condition works, ie:
ServerName COLLATE DATABASE_DEFAULT = SERVERPROPERTY('ServerName')


BrentOzar (approved):
Thanks for the pull request! Just did a quick visual exam and looks good. (I don't have a testbed setup for linked servers.) Merging into the dev branch. Will be in the next release with credit to you in the release notes.

PR #2953, opened 2021-07-20
sp_DatabaseRestore: fix trailing slash for root drives

### LINKED ISSUE #2952: sp_DatabaseRestore's 'trailing slash adder' doesn't work for root drives when moving data files
**Version of the script**
SELECT @Version = '8.04', @VersionDate = '20210530';

Passing a root drive as a `@MoveDataDrive` (or any other @MovexxDrive param) without a trailing slash does not trigger the 'trailing slash adder' functionality since there is no slash to identify if a back or forward slash should be used:

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/60352a57bd4b82512a2b0c9302d10c009acc30ab/sp_DatabaseRestore.sql#L388-L397

Passing a naked drive, i.e. `F:` as the `@MoveDataDrive` or similar parameter.

The slash adder should automatically assume a backslash needs to be added if no slashes are present in the supplied drive.

SQL 2019 - OS irrelevant. I introduced this bug with https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2067

LowlyDBA:
Fixes #2952

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -344,40 +344,40 @@ CREATE TABLE #Headers
 );
 
 /*
-Correct paths in case people forget a final "\" 
+Correct paths in case people forget a final "\" or "/"
 */
 /*Full*/
-IF (SELECT RIGHT(@BackupPathFull, 1)) <> '\' AND CHARINDEX('\', @BackupPathFull) > 0 --Has to end in a '\'
-BEGIN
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks for the pull request! Just did a quick visual exam and looks good. (I don't have a testbed setup for sp_DatabaseRestore at the moment.) Merging into the dev branch. Will be in the next release with credit to you in the release notes.

PR #2955, opened 2021-07-21
fix for issue 2954

### LINKED ISSUE #2954: sp_BlitzIndex - does not show stats when running against tables in DBs outside session's scope
**Version of the script**
@Version = '8.04', @VersionDate = '20210530'

When running a table-specific sp_BlitzIndex, for a table in a database outside your session's currently-connected database, you get no stats.
I think this is because the query that grabs the table's stats uses some system DMVs like the below which are locked to the session's database and cannot peek into other databases' DMVs, even if you run them as EXEC OtherDB.sys.dm_db_stats_properties (... , ...)

[sys-dm-db-stats-properties-transact-sql](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-db-stats-properties-transact-sql)

- Install sp_Blitz into any db.
- EXEC dbo.sp_BlitzIndex @DatabaseName='NotTheDbYourSessionIsConnectedTo', @SchemaName='OtherDbsSchema', @TableName='OtherDbsTable';
- the table stats comes back with an empty set, even though we have defined the correct target @DatabaseName above

- table's stats should show up for table specific runs of sp_BlitzIndex no matter which database the sproc is installed in or what DB scope the session is running in

- tested in SQL 2016, I think this has always been a bug.

MisterZeus:
fixes #2954

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -2725,17 +2725,18 @@ BEGIN
     /* Show histograms for all stats on this table. More info: https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/1900 */
     IF EXISTS (SELECT * FROM sys.all_objects WHERE name = 'dm_db_stats_histogram')
     BEGIN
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

PR #2965, opened 2021-08-01
Add unique constraint awareness

### LINKED ISSUE #2950: sp_BlitzIndex - Unique constraints are scripted as index create/drops
**Version of the script**
Latest and greatest

Unique constraints get scripted as index create/drop statements. Drops will fail if you're quickly trying to declutter indexes.

```
CREATE TABLE dbo.whatever(id int NOT NULL);
CREATE UNIQUE INDEX i ON dbo.whatever(id);
ALTER TABLE dbo.whatever ADD CONSTRAINT u UNIQUE (id);

EXEC sp_BlitzIndex 
    @TableName = 'whatever';
```

image

These should be scripted as alter table statements, like we do with primary keys.

Well...

erikdarlingdata:
This adds `is_unique_constraint` from `sys.indexes` to a couple of the operating modes.

Note that this also adds the new column etc. to the remote server code for Mode 2 as well, and should be announced as a breaking change (of removed, whatever). I don't have a setup to test this part, but the code looks accurate.

This has been tested in modes 0-4, and I've verified that this column exists going back to at least SQL Server 2008.

Closes  #2950

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -263,7 +263,8 @@ IF OBJECT_ID('tempdb..#Ignore_Databases') IS NOT NULL
               is_indexed_view BIT NOT NULL ,
               is_unique BIT NOT NULL ,
               is_primary_key BIT NOT NULL ,
-              is_XML BIT NOT NULL,
+			  is_unique_constraint BIT NOT NULL ,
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

PR #2968, opened 2021-08-02
Add check for unindexed foreign keys

### LINKED ISSUE #2964: sp_BlitzIndex - Identify unindexed foreign keys
While working with a client recently, there were a bunch of perf/blocking issues because foreign keys weren't indexed. I was shocked that BlitzIndex didn't already have a check in there for that.

Add a #temp table that tracks foreign keys where the columns aren't in the key of an index in the correct order. 

This will have some false positives if the fk is on columns a, b and there's an index on b, a but I'm willing to live with that.

Dump a list of unindexed foreign keys into a new #temp table and report on it in findings.

Yes, I have a demo and query for it

erikdarlingdata:
Check is just a slight variation on the current check for FKs with cascading actions.

Closes #2964


BrentOzar (approved):

Issue #2970, opened 2021-08-04
sp_BlitzWho - Add dm_exec_requests.wait_resource to report and output table.


davidhooey:
**Is your feature request related to a problem? Please describe.**

I often use sp_BlitzWho to track blocking situations it would be nice to be able to see the type of resource being blocked such as `KEY: 5:72057594044350464 (557abd962af2)`. It would indicate if row level, page level, or object level lock is in place.

**Describe the solution you'd like**

Add this column to the output table and report.

**Are you ready to build the code for the feature?**

I can work on this, however I do not have all versions of SQL Server to test.

BrentOzar:
This would definitely be something you would need to code if it’s important to you. Do you want to tackle it, or should I close the issue?

---
Tiny glass keyboard
Typos flowing like rivers
As winter snow thaws

davidhooey:
I will give it a go. 😊 From: Brent Ozar ***@***.***>Date: Wednesday, August 4, 2021 at 11:45 AMTo: BrentOzarULTD/SQL-Server-First-Responder-Kit ***@***.***>Cc: davidhooey ***@***.***>, Author ***@***.***>Subject: Re: [BrentOzarULTD/SQL-Server-First-Responder-Kit] sp_BlitzWho - Add dm_exec_requests.wait_resource to report and output table. (#2970)This would definitely be something you would need to code if it’s important to you. Do you want to tackle it, or should I close the issue?---Tiny glass keyboardTypos flowing like riversAs winter snow thaws> On Aug 4, 2021, at 3:22 PM, davidhooey ***@***.***> wrote:> > ﻿> Is your feature request related to a problem? Please describe.> > I often use sp_BlitzWho to track blocking situations it would be nice to be able to see the type of resource being blocked such as KEY: 5:72057594044350464 (557abd962af2). It would indicate if row level, page level, or object level lock is in place.> > Describe the solution you'd like> > Add this column to the output table and report.> > Are you ready to build the code for the feature?> > I can work on this, however I do not have all versions of SQL Server to test.> > —> You are receiving this because you are subscribed to this thread.> Reply to this email directly, view it on GitHub, or unsubscribe.> Triage notifications on the go with GitHub Mobile for iOS or Android.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android.

BrentOzar:
[...]

PR #2972, opened 2021-08-04
Issue #2970 sp_BlitzWho - Add dm_exec_requests.wait_resource to report and output table

davidhooey:
These tests were done for the following stored procedures.

#### BlitzWho, Install-All-Scripts, Install-Core-Blitz-No-Query-Store, Install-Core-Blitz-With-Query-Store

1. Should create procedure.
2. Should show new wait_resource column in both expert and non-expert modes.
3. Should add new wait_resource column and data to output table.
4. Should select new wait_resource column using BlitzWho_Deltas view.

#### BlitzFirst

1. Should successfully run BlitzFirst which executes BlitzWho to...
    a. show new column.
    b. add new column to the output table.

Are there other tests I should do to verify this change?

Tests were done on:

```
Microsoft SQL Server 2016 (SP1-GDR) (KB4458842) - 13.0.4224.16 (X64) Enterprise Edition (64-bit) on Windows Server 2016 Standard 6.3 <X64> (Build 14393: )
```

All versions of SQL Server should be good as the `wait_resource` column exists in SQL Server 2008 R2.

### DIFF
--- a/Install-All-Scripts.sql
+++ b/Install-All-Scripts.sql
@@ -13095,6 +13095,7 @@ SELECT [ServerName]
       ,[query_cost]
       ,[status]
       ,[wait_info]
+      ,[wait_resource]
       ,[top_session_waits]
       ,[blocking_session_id]
       ,[open_transaction_count]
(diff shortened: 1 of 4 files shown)


erikdarlingdata:
Just a heads up, you don't need to also change the installer scripts in the future. Those gets made each release using [this process](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/Documentation/Development/Merge%20Blitz.ps1).

davidhooey:
Oh awesome. I was wondering if there was a build process but couldn’t find it. Should I revert to just the one file change?

________________________________

erikdarlingdata:
You can if you want, but it doesn't matter since the files all get created from scratch anyway. Just trying to save you some work in the future, really.

davidhooey:
Thanks! I will let this stand for now.

BrentOzar (approved):

PR #2973, opened 2021-08-05
Correct the minor version number for SQL 2017 CU25

DavidSchanzer:
Correct minor version number for SQL 2017 CU25 from 3410 to 3401 as indicated in https://support.microsoft.com/en-us/help/5003830


BrentOzar:
Good catch! Fixed, thanks.

PR #2975, opened 2021-08-07
Remove USE

### LINKED ISSUE #2971: sp_BlitzFirst - USE statement causes failure in Azure SQLDB, et al
**Version of the script**
In CheckId 29, there's an unnecessary `USE` statement in the dynamic SQL, because the DMV in use is already prefixed with `tempdb`:

```
FROM tempdb.sys.dm_db_index_operational_stats(DB_ID(''tempdb''), NULL, NULL, NULL) os
    LEFT OUTER JOIN #TempdbOperationalStats os_prior ON os.object_id = os_prior.object_id
```

`USE` statement causes an error.

Run proc in Cloud Nonsense™

Remove the `USE` statement at the beginning

Azureness

erikdarlingdata:
Make it cloudy

Closes #2971


BrentOzar (approved):
Looks good! Thanks for the pull request. Merged into the dev branch, and will be in the next release along with credit to the Grand Trunk.

PR #2978, opened 2021-08-09
Fix #2962

### LINKED ISSUE #2962: sp_BlitzFirst - Check Id #8 fails on SQL Server 2008R2 due to unavailable column (open_transaction_count) in sys.dm_exec_sessions
Version: 8.05 (20210725)

Description:
CheckID 8 fails in SQL Server 10.50.6560.0 in Lines 1765 and 1776 since column **open_transaction_count** is not available for such a build in sys.dm_exec_sessions

Steps to reproduce:
Run sp_BlitzFirst.sql in a SQL Server instance with version 10.50.6560.0

Result
image

Workaround:
Comment the block for Check 8

Expected behaviour
I was not able to find a way to fix the issue, but it seems to me that if open_transaction_count is required and not available in SQL2008 R2, this check should be only valid for SQL Server 2012+

EitanBlumin:
Replace usage of `sys.dm_exec_sessions` with `sys.sysprocesses` to add support for older SQL versions

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -1755,30 +1755,30 @@ BEGIN
 		    'Query Problems' AS FindingGroup,
 		    'Sleeping Query with Open Transactions' AS Finding,
 		    'https://www.brentozar.com/askbrent/sleeping-query-with-open-transactions/' AS URL,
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

PR #2984, opened 2021-09-02
Prevent DB non-default values msg for offline DBs

DavidSchanzer:
I was getting "This database setting is not the default" being reported for "Delayed Durability Enabled" on databases that were offline. On checking the code, I noticed that it was reporting any databases whose sys.databases.state was <> 1 (Restoring), and the state for an offline database is 6. Rather than change it specifically for offline databases (d.state NOT IN (1,6)), I thought that the code should only be reporting non-default values when the database is online (state = 0) - I hope you agree!


BrentOzar (approved):
Thanks for the pull request! Looks good, totally makes sense, merging into dev, will be in the September release with credit to you in the release notes.

PR #2990, opened 2021-09-11
Use sysprocesses instead

### LINKED ISSUE #2961: sp_BlitzFirst - Check Id #1 fails on SQL Server 2008R2 due to unavailable column (database_id) in sys.dm_exec_sessions
Version: 8.05 (20210725)

**Description:**
CheckID 1 fails in SQL Server 10.50.6560.0 in Line 1538 since column database_id is not available for such a build in sys.dm_exec_sessions

image

**Steps to reproduce:**
Run sp_BlitzFirst.sql in a SQL Server instance with version 10.50.6560.0

Result
image

**Workaround:**
Comment the block for Check 1

**Expected behavior**
I was not able to find a way to fix the issue, but it seems to me that if database_id is required and not available in SQL2008 R2, this check should be only valid for SQL Server 2012+

erikdarlingdata:
Closes #2961


BrentOzar (approved):

PR #3000, opened 2021-09-20
sp_BlitzCache - Fix output to table if called from sp_BlitzFirst

### LINKED ISSUE #2998: sp_BlitzFirst - Table for @OutputTableNameBlitzCache not filled if @OutputType = 'NONE'
**Version of the script**

Current dev-branch downloaded and installed with `Install-DbaFirstResponderKit -SqlInstance sql01\prod -Branch dev -OnlyScript Install-Core-Blitz-No-Query-Store.sql -Force`
Source code says: `SELECT @Version = '8.06', @VersionDate = '20210914';`

I use this command:
```
master.dbo.sp_BlitzFirst 
  @OutputType = 'NONE'
, @OutputDatabaseName = 'BlitzMonitoring'
, @OutputSchemaName = 'dbo'
, @OutputTableName = 'BlitzFirst'
, @OutputTableNameFileStats = 'BlitzFirst_FileStats'
, @OutputTableNamePerfmonStats = 'BlitzFirst_PerfmonStats'
, @OutputTableNameWaitStats = 'BlitzFirst_WaitStats'
, @OutputTableNameBlitzCache = 'BlitzCache'
, @OutputTableNameBlitzWho = 'BlitzWho'
```
I use `@OutputType = 'NONE'` because I don't need any output and just want the tables to be filled. But the table BlitzCache is not filled.

This command fills the table, but also outputs two resultsets, one with the output of sp_BlitzCache and one with the "classic" output from sp_BlitzFirst (sp_BlitzFirst 2021-09-14 00:00:00.0000000 +00:00, Server Info, Thanks!):

```
master.dbo.sp_BlitzFirst 
  @OutputDatabaseName = 'BlitzMonitoring'
, @OutputSchemaName = 'dbo'
, @OutputTableName = 'BlitzFirst'
, @OutputTableNameFileStats = 'BlitzFirst_FileStats'
, @OutputTableNamePerfmonStats = 'BlitzFirst_PerfmonStats'
, @OutputTableNameWaitStats = 'BlitzFirst_WaitStats'
, @OutputTableNameBlitzCache = 'BlitzCache'
, @OutputTableNameBlitzWho = 'BlitzWho'
```

I would expect that the table for @OutputTableNameBlitzCache gets filled like all other tables do, even when using @OutputType = 'NONE'

This SQL Server 2017 on Windows Server 2016 Core and I don't know if this worked in previous versions of the procedure.

andreasjordan:
Fixes #2998.

If we look at sp_BlitzFirst calling sp_BlitzWho, then the output type is not a parameter, but sp_BlitzWho only uses the Output* parameters to know what to do: Either save the results to a table or send them to the user.
Why is this not implemented the same way in sp_BlitzCache? Is there a need to having both: Saving results to a table and also displaying them to the user? If so, this pull request will fix the issue but still do both.

But this will only display the results for sp_BlitzCache and the "normal" output from sp_BlitzFirst, as all other output is only saved to the output tables:
```
EXEC sp_BlitzFirst
  @OutputDatabaseName = 'BlitzMonitoring'
, @OutputSchemaName = 'dbo'
, @OutputTableName = 'BlitzFirst'
, @OutputTableNameFileStats = 'BlitzFirst_FileStats'
, @OutputTableNamePerfmonStats = 'BlitzFirst_PerfmonStats'
, @OutputTableNameWaitStats = 'BlitzFirst_WaitStats'
, @OutputTableNameBlitzCache = 'BlitzCache'
, @OutputTableNameBlitzWho = 'BlitzWho'
```
Is this the intended behaviour?

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -6715,15 +6715,17 @@ SET @AllSortSql += N'
 						END; 
 
 				    END;
-					SET @AllSortSql += N' SELECT DatabaseName, Cost, QueryText, QueryType, Warnings, QueryPlan, missing_indexes, implicit_conversion_info, cached_execution_parameters,ExecutionCount, ExecutionsPerMinute, ExecutionWeight, 
-										  TotalCPU, AverageCPU, CPUWeight, TotalDuration, AverageDuration, DurationWeight, TotalReads, AverageReads, 
-										  ReadWeight, TotalWrites, AverageWrites, WriteWeight, AverageReturnedRows, MinGrantKB, MaxGrantKB, MinUsedGrantKB, 
-										  MaxUsedGrantKB, AvgMaxMemoryGrant, MinSpills, MaxSpills, TotalSpills, AvgSpills, PlanCreationTime, LastExecutionTime, LastCompletionTime, PlanHandle, SqlHandle, SetOptions, QueryHash, PlanGenerationNum, RemovePlanHandleFromCache, Pattern 
-										  FROM #bou_allsort 
-										  ORDER BY Id 
-										  OPTION(RECOMPILE);  ';
 
-				
+					IF(@OutputType <> 'NONE')
+					BEGIN
+						SET @AllSortSql += N' SELECT DatabaseName, Cost, QueryText, QueryType, Warnings, QueryPlan, missing_indexes, implicit_conversion_info, cached_execution_parameters,ExecutionCount, ExecutionsPerMinute, ExecutionWeight, 
+											  TotalCPU, AverageCPU, CPUWeight, TotalDuration, AverageDuration, DurationWeight, TotalReads, AverageReads, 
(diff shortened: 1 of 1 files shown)


jeffchulg (changes requested):
Hi @andreasjordan 

To me, there is one more step needed to be complete this change : sp_BlitzCache calls itself in some places in the code. The `OutputType` parameter is not taken into account for these calls. 

You should add it and the code would never fire untrapped `SELECT` again..

Best Regards.

andreasjordan:
Hi @jeffchulg 

I thought about this. But I would have to change a lot of lines in the AllSorts: section and the `SELECT ... FROM #bou_allsort` would still be fired resulting in an empty resultset presented to the user. So this would just be an "add-on" to not fill the `#bou_allsort`.

Before investing the time to change all the calls to sp_BlitzCache in this section, I would like to have @BrentOzar say what he thinks about this.

BrentOzar:
I'm with @andreasjordan on this one - it's a good enough fix for the problem. But @jeffchulg, if you want to code the more ambitious fix for the SortOrder = 'all', you can go for that too. Thanks!

PR #3027, opened 2021-10-29
sp_Blitz - Fix issues with markdown output

### LINKED ISSUE #3018: Rundate value missing in markdown output
"Two pictures say more than two thousand words":

image

image

If that's totally ok and you don't want to change the output of the markdown - stop reading here and close this issue.

I see two ways to include the actual date in the output:
* Change the ckeck to include the date in the Details column as well.
* Change the creation of the output. There are multiple ways to do this - don't know yet which would be the easiers one. 

And as always: I sure can provide a pull request if we agree on the path we want to take.

### LINKED ISSUE #3021: sp_Blitz - Missing documentation details for @OutputType = 'markdown'
With `@OutputType = 'markdown'`, not only the format of the output is changed to an XML column containing information in markdown syntax, but also the information itself is different compaired to other output types.

Relevant code:
https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/99c3073b26d255fde928ef0e3076e7600e687596/sp_Blitz.sql#L831-L833

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/99c3073b26d255fde928ef0e3076e7600e687596/sp_Blitz.sql#L835-L837

https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/99c3073b26d255fde928ef0e3076e7600e687596/sp_Blitz.sql#L9380

The first one is the default for that parameter, so I would not mention this. But the other two will typically change the output. Especially the last one filters out some information without beeing mentioned in the documentation.

The [official documentation](https://www.brentozar.com/blitz/documentation/) does not mention the markdown output type at all, but the source code does:
https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/ae3423c88581f85489a48030bee8899ea6aac7f8/sp_Blitz.sql#L85

Should we change `''MARKDOWN''=bulleted list` to `''MARKDOWN''=bulleted list (including server info, excluding security findings)`?

### LINKED ISSUE #3022: sp_Blitz - Wrong indentation in markdown output
This output:
image

Evaluates in Typora to:
image

Reason is that there is only one space at the beginning of the line, but markdown needs at least two.
Here are the two strings that would need an additional space:
image

After changing this, it looks like this:
image
image

andreasjordan:
Reverts #3020 (rundate in markdown) because it is not needed as I found the root cause of the missing value.
Fixes #3018 (rundate in markdown) by adding COALESCE in SQL, which was the root cause.
Fixes #3022 (wrong indentation in markdown output) by adding an additional space on the second level.
Fixes #3021 (missing documentation details) by extending the parameter explanation.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -9383,7 +9383,7 @@ IF @ProductVersionMajor >= 10 AND  NOT EXISTS ( SELECT  1
 									WHEN r.Priority <> COALESCE(rPrior.Priority, 0) OR r.FindingsGroup <> rPrior.FindingsGroup  THEN @crlf + N'**Priority ' + CAST(COALESCE(r.Priority,N'') AS NVARCHAR(5)) + N': ' + COALESCE(r.FindingsGroup,N'') + N'**:' + @crlf + @crlf
 									ELSE N''
 								END
-								+ CASE WHEN r.Finding <> COALESCE(rPrior.Finding,N'') AND r.Finding <> rNext.Finding THEN N'- ' + COALESCE(r.Finding,N'') + N' ' + COALESCE(r.DatabaseName, N'') + N' - ' + COALESCE(r.Details,N'') + @crlf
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good, merging into the dev branch.

Issue #2976, opened 2021-08-07
Get Cluster Information - Stored Procedure


ragatilao:
**Is your feature request related to a problem? Please describe.**
Need a stored procedure that returns cluster information of the database server.
**Describe the solution you'd like**
Create a stored procedure that will show if the database server is:
1. Clustered or Standalone
2. If clustered, display all the servers that are part of the cluster
3. If clustered, what is primary and secondary node/s.
3. If clustered, is it using Availability Group, Failover Cluster Instance or both

**Describe alternatives you've considered**
I currently have a tool that is retrieving the cluster information using powershell with universal dashboard as application layer.
It is still easier to get the information using stored procedure if you already have SSMS open.
Our team also uses first responder kit, it will be cool if this stored procedure will be part of the other sp_blitz.

**Are you ready to build the code for the feature?**
As much as we'd love to build everything that everyone wants for free, we need your help. Open source is built with your help and code. Are you ready to commit time to this project? Have you got existing code you can help contribute to solve the problem?
Yes.

AdrianB1:
If this feature is interesting for enough people to decide to include it in sp_Blitz, I can contribute code for AG, I have a few of these to work with.

ragatilao:
> If this feature is interesting for enough people to decide to include it in sp_Blitz, I can contribute code for AG, I have a few of these to work with.

Hi Adrian, I have opened a pull request for this. I just dont know how to link the issue and pull request. 
Get Cluster Information - Stored Procedure #2977

AdrianB1:
Hi, I am currently on vacation in a remote area with Internet only on my
phone, I will be back to civilization in 2 weeks. I think you modified the
installers, this is not how it should be done, the installers are generated
by Brent when doing a new release. Please check with Brent if the new code
should be a standalone script or included in sp_Blitz and modify only that
file.

Adrian

ragatilao:
@nick-fotopoulos, @AdrianB1, can you please check the SQL Statement I temporarily uploaded in my github.
this will be the change that I will put on "53 AS CheckID"

https://github.com/ragatilao/sp_blitz_clusterinfo_addon

if everything looks good, I can close this pull request and do separate one to add the sql statement I made on sp_blitz as you guys have suggested.

BrentOzar:
[...]

PR #3033, opened 2021-11-09
Adding code in sp_blitz to show cluster node information.

ragatilao:
Added cluster info in sp_blitz


ragatilao:
#2976

BrentOzar (approved):
Thanks for the pull request. Looks good - doggone, it's nice to have only 2012+ supported right now, so we can hit the AG DMVs without framing them in dynamic SQL. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3049, opened 2021-12-01
Changed the way the @StopAt param is used.

daniel70:
1. we now check that it is a valid datetime
2. add extra log file to restore (that contains transactions up to @StopAt)
3. add STOPAT parameter to the RESTORE LOG command

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -1271,19 +1271,6 @@ BEGIN
 	
 END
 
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request. Looks good - I haven't tested it since I don't have a testbed for sp_AllNightLog, but conceptually it makes sense. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3051, opened 2021-12-08
Add contributors badge and fix code formating

ktaranov:


BrentOzar (approved):
Thanks for the improvements!

PR #3062, opened 2022-01-19
sp_DatabaseRestore: Fix backup file selection for @StopAt and non-split diff backups

### LINKED ISSUE #3061: sp_DatabaseRestore: @StopAt not respected for non-split diff backups
**Version of the script**
8.08 / 20220108

When specifying @RestoreDiff = 1 and @StopAt, @LastDiffBackup is found correctly but then overridden shortly after. This is ok for split backups (fixed by #2899) but for non-split this causes the latest diff backup to be restored even if it is after the value of @StopAt

As above

The latest diff backup on or before @StopAt should be restored, or the latest if @StopAt not specified

SQL Server 2019 (probably all). Has not worked at least since #2180

rob-hague:
Fixes #3061 

Unrelated to #3049

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -1020,17 +1020,14 @@ BEGIN
 	END
     /*End folder sanity check*/
     -- Find latest diff backup 
-	SELECT @LastDiffBackup = MAX(CASE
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Merging into the dev branch, will be in the next release with credit to you in the release notes. I don't have a testbed for sp_DatabaseRestore, but conceptually the changes make sense.

PR #3084, opened 2022-04-08
2022-04-08 Release

BrentOzar:


Install-All-Scripts.sql
```diff
@@ -28244,7 +28246,6 @@ END;
 	   I took a long look at this one, and:
 		1) Trying to account for all the weird places these could crop up is a losing effort. 
 		2) Replace is slow af on lots of XML.
-	- Your mom.
```
bojanderson, line 28247: I see a potential bug. I assume this line of code was probably critical

Issue #3090, opened 2022-04-25
Check 181 Alerts on false positives


josh-simar:
**Version of the script**
8.08

**What is the current behavior?**
Check 181 checks if maintenance plans have Rebuild/Reorg or Rebuild/Update in their plan but does not check if any of these steps are disabled.

**If the current behavior is a bug, please provide the steps to reproduce.**
Create maintenance plan with a Rebuild/Reorg Disable those steps. It still reports that repetitive tasks are occuring.

**What is the expected behavior?**
If the task is disabled the check should not fire that repetitive tasks are occuring.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
2016 SP1
To be clear, this is working just not reporting correctly and I don't think it ever has.

BrentOzar:
This has actually come up in the past. Just because a step is disabled doesn't mean it's not occurring - you can start a job at any step, and you can automate that process as well. I'm a huge believer that if you don't need a harmful step, you shouldn't disable it - you should remove it.

PR #3091, opened 2022-04-25
Add check for disabled steps in 181

josh-simar:
Check 181 checks maintenance plans for repetitive steps but does not check if those steps are disabled. This adds that check.
#3090


josh-simar:
If you want to cancel this and just change the verbiage on the check I'd be okay with that because I'm not sure if the best thing to do would be to add this check, or really to get say, don't use maintenance plans and here's why.

PR #3095, opened 2022-04-26
sp_DatabaseRestore - add check for CommandExec in current database fix

bwiggins10:
Needed to change the check, as you can execute the blitz scripts from the context of another database, but the previous check would only check if the stored procedure existed in the database where the DatabaseRestore SP resided, and wouldn't check if it existed in the database you were currently using at execution (So it would fail even if the database you were using had it, as it was only checking if it existed where the toolkit was deployed).

Not sure if there was a better way to do this than dynamic SQL... But I've tested this and it resolves the issue I found at least. If there's a better method of doing this, would love to know though.

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -212,11 +212,15 @@ BEGIN
     RETURN;
 END;
 
-IF NOT EXISTS (SELECT name FROM sys.objects WHERE type = 'P' AND name = 'CommandExecute')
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good. I don't know of a better way to do this than dynamic SQL - makes perfect sense here. Merging into the dev branch, and will be in the next release with credit to you in the release notes.

PR #3102, opened 2022-05-24
Adding spid and wait_resource to query results and to output table.

davidhooey:
Tested against Microsoft SQL Server 2016 (SP3-OD) (KB5006943) - 13.0.6404.1 (X64)  Enterprise Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor)

Verified the following functionality:

* Query output includes the new `spid` and `wait_resource` columns.
* `@ExportToExcel` includes the new `spid` and `wait_resource` columns.
* Output table includes two new columns, `spid` and `wait_resource`. 
* Procedure dumps the spid and wait_resource data into the two new columns in the output table.


davidhooey:
That should be cleaner now.

PR #3113, opened 2022-06-16
sp_DatabaseRestore: problem with StopAt when backup file is in different time zone from restoring server

edmundoriviera:
I perhaps haven't named the variables particularly well. In my local copy I've changed @StopAt to be @StopAtUtc for clarity.

This was causing me a problem as the file list would be limited in UTC but the @StopAt passed to restore was being treated as Local time. In my case my server backups are in BST and the files are in UTC.

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -25,6 +25,7 @@ ALTER PROCEDURE [dbo].[sp_DatabaseRestore]
     @ForceSimpleRecovery BIT = 0,
     @ExistingDBAction tinyint = 0,
     @StopAt NVARCHAR(14) = NULL,
+	@BackupServerTimezoneOffsetToUtc INT = 0,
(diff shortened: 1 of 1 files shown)


BrentOzar:
Howdy! When you want to make a change in the First Responder Kit, can you create an issue first describing the change you'd like to make? Then, other folks can chime in with whether this proposed change works for them, and if there's anything we need to think about in the design process. It's documented in the readme here: https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/tree/main#how-to-get-support

I don't actually use sp_DatabaseRestore at this time, so I'm leaving this pull request open to see if anyone else can test it and confirm that it has the desired effect. Another thing you could do is do a walkthrough of what changes you made, and why, and then other folks might be able to give it a double-check.

BrentOzar:
Because we didn't get testers or an issue, closing this one. Thanks anyway!

PR #3115, opened 2022-06-30
Fixes a collection error when server collation is not set to Latin1_General_CI_AI

BjornNordblom:
Fixes a collection error when server collation is not set to Latin1_General_CI_AI.


BrentOzar (approved):

BrentOzar (changes requested):
Ooo, sorry - this is actually in InstallAllScripts. That's automatically generated at release, based on the individual scripts. If you have changes you want to make, please make 'em to the relevant script, not the installer. Thanks!

PR #3116, opened 2022-07-01
sp_Blitz - Add "used" to percentage to make information clearer

andreasjordan:
Since the message starts with "GB free", but the percentage refers to "% used", I suggest the addition of "used".


BrentOzar (approved):

PR #3130, opened 2022-07-22
Fixes #3129 - sp_BlitzFirst in Azure SQL Managed Instances

### LINKED ISSUE #3129: sp_BlitzFirst Check 44 in Azure Managed Instance fails with error
**Version of the script**
Version '8.10', VersionDate '20220718'

Runs through until check 44 then reports "Database '?' does not exist. Make sure the name is entered correctly."

execute sp_BlitzFirst @expertMode = 1 in an Azure MI

No error

New user, so no previous versions tested. I believe Azure MI is in constant state of revision but I think I have a fix:

Unless I am misunderstanding the purpose of the IF...ELSE checks for 'SQL Azure' prior to the population of @StringToExecute and similarly the use of sp_MSForeachdb, shouldn't line 2478 be updated to not include 'USE [?]; '? as it's being appended if required first?

Thank you:-)

dmonlineuk:
Retains checks for Azure SQL DB and tested to work there, but now enables it to work for Azure SQL Managed Instances as (I believe is) intended - also tested.

TLDR - looked for SERVERPROPERTY('Edition') =/<> 'Azure SQL' and updated to SERVERPROPERTY('EngineEdition') =/<> 5

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -2466,12 +2466,12 @@ If one of them is a lead blocker, consider killing that query.'' AS HowToStopit,
 			/* We don't want to hang around to obtain locks */
 			SET LOCK_TIMEOUT 0;
 
-			IF SERVERPROPERTY('Edition') <> 'SQL Azure'
+			IF SERVERPROPERTY('EngineEdition') <> 5
 				SET @StringToExecute = N'USE [?];' + @LineFeed;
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good, merging into the dev branch. Will be in the next release with credit to you in the release notes.

PR #3146, opened 2022-09-30
sp_Blitz and sp_BlitzLock Change so they work on RDS Instances with ServerName Changed

### LINKED ISSUE #3140: sp_Blitz not working on AWS RDS Instances where Server Name is changed
**Version of the script**
Version = 8.10 and Version Date = 20220718

If the sp_Blitz is ran on a AWS SQL Server RDS Instance where the Server Name has been renamed then an error message appears saying:

> User does not have permissions to perform this action
> The server principal **_"whatever_account_you_are_using"_** is not able to access the database "model" under the security context

However if ran on an RDS Instance where the Server Name has not been changed then it works perfectly fine

1. Create an RDS SQL Server Instance through the AWS Console
2. Run `SELECT @@SERVERNAME` to get current Server Name
3. Run
```sql
EXEC sp_dropserver '<old_name>';  
GO  
EXEC sp_addserver '<new_name>', local;  
GO  
```
4. Restart instance for change to take effect
5. Install sp_Blitz and then run `EXEC sp_Blitz`

The expected behavior is that regardless if the Server Name has been changed the script runs the same for all RDS instances.

SQL Server version was 14.0.3381.3 (aka 2017) and because it's an RDS instance I do not have access to the underlying Operating System

### A couple questions I'm adding at the end

**Who will fix this?**

I volunteer as tribute

**Any other information or discussions about this?**

Here's the discussion that was going on in Slack about this topic. Brent agrees we can make the change, but also admits what I'm doing which causes this issue is unorthodox (and I don't deny).

SlackConvo

bojanderson:
This change will close #3140 if accepted and merged.

I made a very minor change to sp_Blitz and sp_BlitzLock and removed the check for ServerName because the ServerName can be changed on an RDS instance.

I tested this on 4 version of RDS and confirmed they it works after this change and confirmed on a regular EC2 instance version of SQL 2019 that it works.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -425,7 +425,6 @@ AS
 		/* If the server is Amazon RDS, skip checks that it doesn't allow */
 		IF LEFT(CAST(SERVERPROPERTY('ComputerNamePhysicalNetBIOS') AS VARCHAR(8000)), 8) = 'EC2AMAZ-'
 		   AND LEFT(CAST(SERVERPROPERTY('MachineName') AS VARCHAR(8000)), 8) = 'EC2AMAZ-'
-		   AND LEFT(CAST(SERVERPROPERTY('ServerName') AS VARCHAR(8000)), 8) = 'EC2AMAZ-'
 		   AND db_id('rdsadmin') IS NOT NULL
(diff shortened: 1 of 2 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good, merging into the dev branch. Will be in the next release with credit to you in the release notes.

PR #3157, opened 2022-10-20
Use DATETIME2 in filter when checking dbi_LastLogBackupTime

carlsixsmith-moj:
Running sp_blitz raises an error doing the date filter on dbi_LastLogBackupTime. This address that issue.

SELECT CONVERT(DATETIME, '2021-06-14 09:58:08.793')     --> this one fails to convert to datetime
SELECT CONVERT(DATETIME2, '2021-06-14 09:58:08.793')   --> this one converts ok.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -1090,7 +1090,7 @@ AS
 												)
 												OR
 												(
-													Convert(datetime,ll.Value) < DATEADD(dd,-7, GETDATE())
+													Convert(datetime2, ll.Value) < DATEADD(dd,-7, GETDATE())
 												)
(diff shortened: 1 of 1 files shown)


BrentOzar:
Hmm, can you elaborate on what you mean? Both of those convert OK on my machine.
Screenshot 2022-12-11 at 10 39 01 AM

BrentOzar:
@carlsixsmith ah! That makes sense. Can you test with the current development branch version first? Looks like there's a merge conflict because it was fixed a different way in another user's pull request with this: Convert(datetime,ll.Value,21)

PR #3175, opened 2022-11-22
sp_BlitzLock combined changes

### LINKED ISSUE #3166: sp_BlitzLock needs a facelift
Since we're coming up on the 6th anniversary of this thing getting written on the plane ride to PASS, there are a number of issues, shortcomings, and enhancements I'd like to knock out for this at once.

* Be able to read from an event file or the ring buffer
* Improve performance generally: some of the XML stuff has obvious issues
* Fix several data duplication bugs
* Clean XML to address occasional character parsing errors
* Resolve [Database Id = 5 Object Id = 98489874] entries

You get the idea. It has gone largely untouched in important ways for a long time.

I would like to work on this to implement everything in a single issue, with as few passes as possible. I just need to get in there and plug through everything.

Continue to wonder why I don't just fix this stuff.

My soul is prepared! How's yours?

erikdarlingdata:
This pull request is a mommy makeover for sp_BlitzLock. 

At a high level, it:
* fixes many bugs
* makes it compatible with azure
* allows for using ring buffers or event files as targets for user defined sessions
* fixes many performance issues
* makes the code more readable without using a single CTE
* Adds more feedback mechanisms
* Includes some breaking changes

Specific bugs it fixes
* dynamic SQL truncation issues when building strings missing N prefixes
* string truncation errors on agent job update
* several flaws in the "findings" table where results were inconsistent across checks for similar patterns
* change XML resolution to use _TRY_CATCH_ instead

Azure compatibility (validated):
* hits the "database" XE views when necessary
* works with ring buffers

Azure compatibility (not validated):
* when hitting event files, it might not work
* i don't have an instance available to test if it gets the right http blob storage URL

Target selection:
* parameter added to specify if you're hitting ring buffers or an event file.
* if NULL, will use DMVs to figure it out automagickally 
* for the system health session, it always goes to the file; I can't get deadlocks to go to the ring buffer target for it

Performance issues:
* an attempt at making the system health files less painful (needs more testing)
* uses the _exists_ method to filters parameters in XML rather than _values_
* does a better job of trying to get parallel queries for CPU intensive XML shredding
* adds TABLOCK and RECOMPILE hints where applicable
* fixes implicit conversion issues in many places
* pulls out unnecessary xml traversals and other unnecessary work 
* change the "get all tables in all databases" dynamic SQL to only run for databases involved in deadlocks
* added database name column to a couple tables to do less work resolving them later

Formatting:
* I did my best to correct issues with current code to hopefully make it more readable to humans
* centralized object creation and various checks for version, edition, location, etc.
* centralized final queries so there aren't two copies to maintain for insert or return to SSMS
* You may not like it, but this is what peak performance looks like

Feedback mechanisms:
* More query start/finish logging
* Known poor performers can return query plans under debug
* More code commenting

Breaking changes:
* Removed the @TOP parameter; it was generally useless
* Added "better" defaults  for start and end date parameters

To do:
* at some point i would like to get some additional safety built in for columns that are returned as XML to remove weird characters. i've started to do it in my scripts, but i haven't figured out a good way to apply it here yet

Closes #3166


erikdarlingdata:
Something is going weird with GitHub and this PR. Gonna close it out and push a new one.

BrentOzar:
Thanks for the pull request! I'm not even gonna pretend I reviewed the code in this, hahaha - you da man. Merging into the dev branch, will be in the December to Remember release with credit to you in the release notes.

PR #3177, opened 2022-11-28
Update sp_DatabaseRestore.sql

wfvdijk:
fixed ticket 3156 sp_databaserestore don't see backups which consists of more than 10 backupfiles when using @stopAt


BrentOzar (approved):
Thanks for the pull request! Merging into the dev branch for testing - I know, I should test it before merging, but I made other SQL 2022 compat fixes for this proc, and the only testbed I have right now is 2022, so it's easier if I merge it then work on any additional necessary changes.

PR #3180, opened 2022-12-01
Prevent lock timeout thrown by sp_msforeachdb

codykonior:

When setting up sp_BlitzFirst in an agent job on a few dozen SQL 2019 with AGs, I was seeing very random periodic job failures caused by lock timeouts. I traced it down to the section where you SET LOCK_TIMEOUT 0 and then look for modified statistics.

You catch the lock timeout error inside the statistics section, but it turns out that sp_msforeachdb is also susceptible to exiting with a lock timeout. By wrapping that in a further try/catch block, I'm able to prevent the job step from failing unnecessarily.

This is nicer than having the job fail randomly, or having to set the job step to exit with success on failure.

I couldn't force the behaviour to happen, even wrapping sp_msforeachdb in a GO 10000 loop and failing over servers everywhere. It's something else. For reference on ~20 servers and running every 15 minutes, one execution would fail every few hours.

Not throwing that error.

SQL 2019 Windows 2016. Unsure if it worked previously.

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -2516,7 +2516,16 @@ If one of them is a lead blocker, consider killing that query.'' AS HowToStopit,
                                     ;
 
 			IF SERVERPROPERTY('EngineEdition') <> 5 /*SERVERPROPERTY('Edition') <> 'SQL Azure'*/
-	            EXEC sp_MSforeachdb @StringToExecute;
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good. Merging into the dev branch, will be in the December release with credit to you in the release notes.

Issue #3163, opened 2022-10-29
sp_BlitzWho - shows procedure definition not the current statement in query_text if proc is executing for the first time


Adedba:
**Version of the script**
SELECT @Version = '8.11', @VersionDate = '20221013';

**What is the current behavior?**
If you have a stored procedure executing for the first time (not in the plan cache) the query_text column will not show you the current statement being executed, it will show you the stored procedure definition instead.

**If the current behavior is a bug, please provide the steps to reproduce.**

Clear your cache (DBCC FREEPROCCACHE) 

create a stored procedure: 

```
USE [StackOverflow];
GO

CREATE PROCEDURE usp_GetTopVotesByYearAndMonth
AS 
BEGIN 
	
	SELECT /* Query 1 of 2 */ TOP (10) 
		DATENAME(YEAR,[CreationDate]) AS VoteYear,
		COUNT(CreationDate) AS TotalVotes
	FROM [dbo].[Votes] v
	INNER JOIN [dbo].[VoteTypes] vt ON v.[VoteTypeId] = vt.[Id]
	GROUP BY 
		DATENAME(YEAR,[CreationDate])
	ORDER BY 
		[TotalVotes] DESC;	
	
	SELECT /* Query 2 of 2 */ TOP (10) 
		DATENAME(MONTH,[CreationDate]) AS VoteMonth,
		COUNT(CreationDate) AS TotalVotes
	FROM [dbo].[Votes] v
	INNER JOIN [dbo].[VoteTypes] vt ON v.[VoteTypeId] = vt.[Id]
	GROUP BY 
		DATENAME(MONTH,[CreationDate])
	ORDER BY 
		[TotalVotes] DESC;		
END 
go
```

Query window 1: Execute stored procedure:
`EXEC usp_GetTopVotesByYearAndMonth;`

Query window 2: Execute sp_BlitzWho:
`EXEC sp_BlitzWho
`
Observer the query_text column, this will contain the stored procedure definition. 

Now execute your stored procedure again and the execute sp_BlitzWho whilst running and you will notice that this time the statement being executed will now show in the query_text column. 

**What is the expected behavior?**
For the current statement in a stored procedure to show in the query_text column

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
Tested on SQL 2016

BrentOzar:
[...]

PR #3197, opened 2022-12-13
3163 - sp_BlitzWho - shows procedure definition not the current state…

Adedba:
#3163 - sp_BlitzWho - shows procedure definition not the current statement.

Ameneded query_text column to get statement offsets from sys.dm_exec_requests instead of sys.dm_exec_query_stats


Adedba:
That's odd, i have a conflict - I will sort this out and re submit

Issue #3203, opened 2022-12-16
sp_AllNightLog - Error Handling Issues


DavidWiseman:
sp_AllNightLog - Error Handling

**Is your feature request related to a problem? Please describe.**

I was helping a colleague setup sp_AllNightLog where the restores were not working. This had a simple solution - the CommandLog table was missing, but troubleshooting the issue wasn't straightforward due to the error handling in sp_AllNightLog.

If you run sp_DatabaseRestore and the CommandLog table is missing you get a nice error message:

```
Msg 50000, Level 16, State 1, Procedure dbo.CommandExecute, Line 157 [Batch Start Line 7]
The table CommandLog is missing. Download https://ola.hallengren.com/scripts/CommandLog.sql.
```

If you run `EXEC sp_AllNightLog @Restore = 1,@Debug=1`, the error is caught.  The CATCH block doesn't actually do anything.  We get `Error number IS NULL` (normal/expected) and `Updating backup_worker for database DB1 for UNsuccessful backup`. These are normal print statements and we don't get any additional information about what caused the error.  We only get the UNsuccessful backup message as there is a check outside the CATCH block to see if we have a DB in a restoring state.

```
Checking restore path
Beginning Restores
Checking variables
Begin tran to grab a database to restore
Updating restore_worker for database DB1
Restoring full backup for DB1
Path for FULL backups for DB1 is \\LAB2022\Backup\LAB2019NODE1\DB1\FULL\
Path for LOG backups for DB1 is \\LAB2022\Backup\LAB2019NODE1\DB1\LOG\
Starting first Full restore from: 
\\LAB2022\Backup\LAB2019NODE1\DB1\FULL\
Getting default data drive for @MoveDataDrive
Getting default log drive for @MoveLogDrive
Setting default data drive for @MoveFilestreamDrive
Setting default data drive for @MoveFullTextCatalogDrive
DIR /b "\\LAB2022\Backup\LAB2019NODE1\DB1\FULL\"
INSERT INTO #FileListParameters WITH (TABLOCK)
       (LogicalName, PhysicalName, Type, FileGroupName, Size, MaxSize, FileID, CreateLSN, DropLSN
       ,UniqueID, ReadOnlyLSN, ReadWriteLSN, BackupSizeInBytes, SourceBlockSize, FileGroupID, LogGroupGUID
[...]

PR #3207, opened 2023-01-06
sp_AllNightLog - Error handling improvements

DavidWiseman:
## Summary 

* Fix error handling
* Add check for CommandLog table
* Change Processing order to prevent DBs with an error condition from monopolizing the queue 
* Moved job enabled check to end of loop - faster & more reliable shutdown. 
* Removed @error_number IS NOT NULL check.  We are in the catch block so an error has occurred and an error number is expected.  

#3203

## Restore Example:
Add this to the top of sp_DatabaseRestore:
```sql
RAISERROR('Fatal error',16,1)
```

New version shows the error details for easy troubleshooting.  We also exit after we process the first DB from the queue with the jobs disabled - rather than wait for the queue to empty.  With the error now been caught the DB ends up back in the queue so we would never shutdown if there is an error condition without also including this change.  

With the old version it's not clear that an error occurred.  We also process each item in the queue until the queue is empty so we are looping more times (potentially resulting in slow/unreliable shutdown).  

As the error handling is now working, an error condition causes the DB to be put back in the queue for immediate re-processing.  The ordering has been changed so it won't always be the first DB to be pulled off the queue (preventing other DBs from processing if there is an error condition on 1 DB).

image

## Backup example:
Add this to dbo.DatabaseBackup:
RAISERROR('fatal error 123',16,1)

Similar to the restore fix, we can now see the error where there was no error indication with the old version.
image

If the CommandLog table is missing we now bail out early with a friendly error message:
image

### DIFF
--- a/sp_AllNightLog.sql
+++ b/sp_AllNightLog.sql
@@ -181,6 +181,12 @@ IF NOT EXISTS (SELECT * FROM sys.procedures WHERE name = 'DatabaseBackup')
 			
 			RETURN;
 		END 
+IF NOT EXISTS (SELECT * FROM sys.tables WHERE name = 'CommandLog')
+		BEGIN 		
+			RAISERROR('Ola Hallengren''s CommandLog table must be installed in the same database (%s) as SQL Server First Responder Kit. More info: http://ola.hallengren.com', 0, 1, @CurrentDatabaseContext) WITH NOWAIT;
+			
+			RETURN;
+		END
 
 /*
 Make sure sp_DatabaseRestore is installed in same database
@@ -797,7 +803,10 @@ LogShamer:
 
 			
 				BEGIN
-	
+					/* Reset error at start of loop*/
+					SELECT 	@error_number = NULL, 
+							@error_severity = NULL, 
+							@error_state = NULL;
 					BEGIN TRY
 							
 							BEGIN TRAN;
@@ -834,7 +843,14 @@ LogShamer:
                                                 AND (bw.error_number IS NULL OR bw.error_number > 0) /* negative numbers indicate human attention required */
 												AND bw.ignore_database = 0
 											  )
-										ORDER BY bw.last_log_backup_start_time ASC, bw.last_log_backup_finish_time ASC, bw.database_name ASC;
+										/* 
+											Process DBs with oldest log backup or the oldest error date if this is more recent than last backup time.
+											If an error occurs the DB is back in the queue but priority is given to DBs with the oldest 'attempted' date.
+											This will prevent a single DB with an error condition from monopolizing the queue
+										*/
+										ORDER BY CASE WHEN bw.last_error_date> bw.last_log_backup_start_time THEN bw.last_error_date ELSE bw.last_log_backup_start_time END ASC, 
+												 bw.last_log_backup_finish_time ASC, 
+												 bw.database_name ASC;
 	
 								
 									IF @database IS NOT NULL
@@ -889,19 +905,6 @@ LogShamer:
 							IF @Debug = 1 RAISERROR('No databases to back up right now, starting 3 second throttle', 0, 1) WITH NOWAIT;
 							WAITFOR DELAY '00:00:03.000';
 
-                            /* Check to make sure job is still enabled */
-		                    IF NOT EXISTS (
-						                    SELECT *
-						                    FROM msdb.dbo.sysjobs 
-						                    WHERE name LIKE 'sp_AllNightLog_Backup%'
-                                            AND enabled = 1
-						                    )
-                                BEGIN
-				                    RAISERROR('sp_AllNightLog_Backup jobs are disabled, so gracefully exiting. It feels graceful to me, anyway.', 0, 1) WITH NOWAIT;
-				                    RETURN;
-                                END        
-
-
 						END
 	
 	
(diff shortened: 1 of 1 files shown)


BrentOzar (changes requested):
Thanks for the pull request. There are changes in here that I don't quite understand - I'm guessing maybe you removed code you didn't want to, or tried reformatting the code. If you look at the pull request page here, the old code is on the left, and the changed code is on the right: https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/pull/3207/files

If you didn't mean to make all these changes, can you remove the parts you didn't mean to make? For example, on the left in the original code circa line 892, an entire query has been deleted with no explanation. Around lines 1016-1020, I can't tell what we changed, or why.

sp_AllNightLog.sql
```diff
@@ -889,19 +905,6 @@ LogShamer:
 							IF @Debug = 1 RAISERROR('No databases to back up right now, starting 3 second throttle', 0, 1) WITH NOWAIT;
 							WAITFOR DELAY '00:00:03.000';
 
-                            /* Check to make sure job is still enabled */
-		                    IF NOT EXISTS (
-						                    SELECT *
-						                    FROM msdb.dbo.sysjobs
```
DavidWiseman, line 895: This was moved to line 1066.  The check is now done at the end of the loop - so we process 1 database and perform this check, rather than wait for the queue to be empty (database is NULL).  The problem with waiting for the queue to be empty is that this will never occur if there is an error condition as the DB gets added back to the queue for immediate processing.

sp_AllNightLog.sql
```diff
@@ -1277,19 +1301,6 @@ IF @Restore = 1
 			                        RAISERROR('sp_AllNightLog_Backup jobs are enabled, so gracefully exiting. You do not want to accidentally do restores over top of the databases you are backing up.', 0, 1) WITH NOWAIT;
 			                        RETURN;
                                 END        
-
```
DavidWiseman, line 1280: This was moved to line 1507.  The check is now done at the end of the loop - so we process 1 database and perform this check, rather than wait for the queue to be empty (database is NULL).  The problem with waiting for the queue to be empty is that this will never occur if there is an error condition as the DB gets added back to the queue for immediate processing.

sp_AllNightLog.sql
```diff
@@ -181,6 +181,12 @@ IF NOT EXISTS (SELECT * FROM sys.procedures WHERE name = 'DatabaseBackup')
 			
 			RETURN;
 		END 
+IF NOT EXISTS (SELECT * FROM sys.tables WHERE name = 'CommandLog')
```
DavidWiseman, line 184: Added check to bail out early if CommandLog table doesn't exist

sp_AllNightLog.sql
```diff
@@ -1211,7 +1228,14 @@ IF @Restore = 1
 															AND state <> 1 /* Restoring */
 															AND NOT (state=0 AND d.is_in_standby=1) /* standby mode */
 														)
-										ORDER BY rw.last_log_restore_start_time ASC, rw.last_log_restore_finish_time ASC, rw.database_name ASC;
+										/*
```
DavidWiseman, line 1231: As the error handling is now working, an error condition causes the DB to be put back in the queue for immediate re-processing. The ordering has been changed so it won't always be the first DB to be pulled off the queue (preventing other DBs from processing if there is an error condition on 1 DB).

sp_AllNightLog.sql
```diff
@@ -952,74 +955,75 @@ LogShamer:
 																	        @Compress = 'Y', --This is usually a good idea
 																	        @LogToTable = 'Y'; --We should do this for posterity
 	
-
```
DavidWiseman, line 955: This should be done in the CATCH block.  If an error occurs this code wouldn't get executed.

sp_AllNightLog.sql
```diff
@@ -952,74 +955,75 @@ LogShamer:
...
-	
-						IF  @error_number IS NOT NULL
+						SELECT @error_number = ERROR_NUMBER(), 
+											   @error_severity = ERROR_SEVERITY(), 
+											   @error_state = ERROR_STATE();
+											   
 
+						/* 
+							Re-throw the error, preserving as much of the original error details as possible to assist  withtroubleshooting.
+						*/
```
DavidWiseman, line 979: Provide a detailed error, preserving as much of the original error information as possible to assist troubleshooing

sp_AllNightLog.sql
```diff
@@ -1365,76 +1376,74 @@ IF @Restore = 1
 																				   @Debug = @Debug
 	
 											END
-
```
DavidWiseman, line 1368: This should be done in the CATCH block. If an error occurs this code wouldn't get executed.

sp_AllNightLog.sql
```diff
@@ -1365,76 +1376,74 @@ IF @Restore = 1
...
-						IF  @error_number IS NOT NULL
 
+						SELECT @error_number = ERROR_NUMBER(), 
+								@error_severity = ERROR_SEVERITY(), 
+								@error_state = ERROR_STATE();
+	
+						
+						/* 
+							Re-throw the error, preserving as much of the original error details as possible to assist  withtroubleshooting.
+						*/
```
DavidWiseman, line 1399: Provide a detailed error, preserving as much of the original error information as possible to assist troubleshooing

sp_AllNightLog.sql
```diff
@@ -1365,76 +1376,74 @@ IF @Restore = 1
...
+					/*
+					
+					Catch any erroneous zones
+					
+					*/
 					BEGIN CATCH
-	
-						IF  @error_number IS NOT NULL
 
+						SELECT @error_number = ERROR_NUMBER(),
```
DavidWiseman, line 1392: Moved to CATCH block

sp_AllNightLog.sql
```diff
@@ -1455,7 +1464,7 @@ IF @Restore = 1
                             /* Make sure database actually exists and is in the restoring state */
                             IF EXISTS (SELECT * FROM sys.databases WHERE name = @database AND state = 1) /* Restoring */
 								BEGIN
-							        SET @msg = N'Updating backup_worker for database ' + ISNULL(@database, 'UH OH NULL @database') + ' for successful backup';
+							        SET @msg = N'Updating restore_worker for database ' + ISNULL(@database, 'UH OH NULL @database') + ' for successful restore';
```
DavidWiseman, line 1467: typo

sp_AllNightLog.sql
```diff
@@ -1468,7 +1477,7 @@ IF @Restore = 1
                                 END
                             ELSE /* The database doesn't exist, or it's not in the restoring state */
                                 BEGIN
-							        SET @msg = N'Updating backup_worker for database ' + ISNULL(@database, 'UH OH NULL @database') + ' for UNsuccessful backup';
+							        SET @msg = N'Updating restore_worker for database ' + ISNULL(@database, 'UH OH NULL @database') + ' for UNsuccessful restore';
```
DavidWiseman, line 1480: typo

sp_AllNightLog.sql
```diff
@@ -1492,7 +1501,19 @@ IF @Restore = 1
 							SET @database = NULL;
 
 
-						END; -- End update for successful backup	
+						END; -- End update for successful restore
```
DavidWiseman, line 1504: This was moved here. The check is now done at the end of the loop - so we process 1 database and perform this check, rather than wait for the queue to be empty (database is NULL). The problem with waiting for the queue to be empty is that this will never occur if there is an error condition as the DB gets added back to the queue for immediate processing.

sp_AllNightLog.sql
```diff
@@ -952,74 +955,75 @@ LogShamer:
...
 						/*
 						
-						If the ERROR() function returns a number, update the table with it and the last error date.
+						Update the table with last error number and the last error date.
 
 						Also update the last start time to 1900-01-01 so it gets picked back up immediately -- the query to find a log backup to take sorts by start time
 
 						*/
-	
-							BEGIN
```
DavidWiseman, line 984: BEGIN/END removed along with associated TAB spaces.  (See Removed IS NOT NULL comment above)

sp_AllNightLog.sql
```diff
@@ -1365,76 +1376,74 @@ IF @Restore = 1
...
 						/*
 						
-						If the ERROR() function returns a number, update the table with it and the last error date.
+						Update the table with last error number and the last error date.
 
 						Also update the last start time to 1900-01-01 so it gets picked back up immediately -- the query to find a log restore to take sorts by start time
 
 						*/
-	
-							BEGIN
```
DavidWiseman, line 1399: BEGIN/END removed along with associated TAB spaces (See Removed IS NOT NULL comment above)

DavidWiseman (commented):
Added comments to help make review easier.

DavidWiseman:
I've added comments to all the code changes, but just to clarify - the changes are all intentional.  If you need anything else from me, please let me know.

BrentOzar:
So, this is a tough one: it's been open for almost a year, and no one else has been able to do the code review due to the number of changes. I'm going to go ahead and close this, but going forward, when you contribute code to open source projects, here are a few tips to maximize your chance of getting the pull request accepted:
* Make one issue for each problem you're trying to solve
* Include a repro script that someone else can run on their machine in order to see the issue - [here's an example of a repro script](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3405#issuecomment-1847595377)
* Make one pull request for each issue, max (don't try to solve multiple problems in one pull request)

That way, someone can copy/paste your repro script, see the exact error or problem that you're trying to solve, and then run the repro script again with your changes, and see that it's fixed.

Otherwise, the way it is here, I don't know if this PR made things better or worse. I really wish I had the time to dig into pull requests like this, believe me, but only so many hours in each day.

DavidWiseman:
Thanks for the response - I'm happy for this to be killed off.  I normally prefer to do smaller PRs when contributing to other projects but this required some more substantial restructuring to fix the broken error handling.  I appreciate this made it hard to review.  

I think the overall principle behind sp_AllNightLog is good but it has a number of issues.  I ended up writing a console app as a replacement as I needed something that could do the same thing but also work from backups to Azure blob.

BrentOzar:
@DavidWiseman yeah, totally understand. I've worked with a couple of clients who started with sp_AllNightLog, and then gradually wanted something more powerful and flexible, and then built complex applications to do it.

PR #3218, opened 2023-01-26
Fix for #3215

### LINKED ISSUE #3215: sp_BlitzLock - Arithmetic overflow in the column wait_time_hms  of the chopsuey CTE when dp.wait_time adds up to more than 2147483647

VladDBA:
Closes #3215 
3rd time's the charm.

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -2441,6 +2441,76 @@ BEGIN
                      )
                  ),
              wait_time_hms =
+             /*the more wait time you rack up the less accurate this gets, 
+             it's either that or erroring out*/
+            CASE 
+                WHEN 
+                    SUM
+                    (
+                        CONVERT
+                        (
+                            bigint, 
+                            dp.wait_time
+                        )
+                    )/1000 > 2147483647
+                THEN 
+                   CONVERT
+                   (
+                       nvarchar(30),
+                       DATEADD
+                       (
+                            MINUTE,
+                            (
+                                 (
+                                    SUM
+                                    (
+                                       CONVERT
+                                       (
+                                           bigint, 
+                                           dp.wait_time
+                                       )
+                                    )
+                                 )/
+                                 60000
+                            ),
+                            0
+                       ),
+                       14
+                   )
+                WHEN 
+                    SUM
+                    (
+                        CONVERT
+                        (
+                            bigint, 
+                            dp.wait_time
+                        )
+                    ) BETWEEN 2147483648 AND 2147483647000
+                THEN 
+                   CONVERT
+                   (
+                       nvarchar(30),
+                       DATEADD
+                       (
+                            SECOND,
+                            (
+                                 (
+                                    SUM
+                                    (
+                                       CONVERT
+                                       (
+                                           bigint, 
+                                           dp.wait_time
+                                       )
+                                    )
+                                 )/
+                                 1000
+                            ),
+                            0
+                       ),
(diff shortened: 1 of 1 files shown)


VladDBA:
To add more context to the fix:
- if the aggregate sum of the wait_time column in the #deadlock_process table (used in the chopsuey CTE) is between 2147483648 and 2147483647000, the sum will be divided by 1000 and DATEADD will use SECOND instead of MILLISECOND, reducing the accuracy to one second instead of millisecond
- if (and this was my extremely unfortunate case) the aggregate sum of the wait_time column in the #deadlock_process table (used in the chopsuey CTE) is greater than 2147483647000, the sum will be divided by 60000 and DATEADD will use MINUTE instead, reducing the accuracy to one minute.
I know this is an edge case, but I figured that if anyone else is unfortunate to run into an instance with 32000 deadlocks in 7 days, they'd want more to see that than they would care for the potential reduction in wait time accuracy.

BrentOzar (approved):
Thanks for the pull request! I agree, lower accuracy totally makes sense here. Great idea. Merging this into the dev branch, will be in the next release with credit to you in the release notes.

Issue #3210, opened 2023-01-09
QRY_PROFILE_LIST_MUTEX blocking still a problem


sqlslinger:
**Version of the script**
8.11

**What is the current behavior?**
We are still seeing the issues described in this closed bug report:
https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2907

![image](https://user-images.githubusercontent.com/64641794/211401940-e46702ce-73d5-495b-af9c-c06750deccee.png)

This is an occurance of blocking locks where sp_Blitz is not the lead blocker, but one of the processes being blocked.  In this case, the lead blocker was the FormattedAddress function spid 160.  It blocked our Saturday job to do index maintenance, spid 57.  Spid 57 then blocked other processes including sp_Blitz spid 104.  You can see that spid 160 is waiting on QRY_PROFILE_LIST_MUTEX.  

To address the issue, the normal first action is to kill the lead blocker.  "KILL 160" did not do anything.  Since we have seen this before, we knew that killing sp_Blitz does get the jobs unstuck.  So "KILL 104" released the blocking locks and got the processes unstuck.

When this lock occurs, it represents a **serious** system issue.  All Sql Agent jobs stop operating.  We had attempted to address the issue by checking for sp_Blitz job running for more than 10 minutes.  However, since all jobs stopped executing once this block took place, it did not work around the issue.

**If the current behavior is a bug, please provide the steps to reproduce.**

Reproducing this issue is not easy.  We have sp_BlitzFirst running every 15 minutes on about 90 servers, and we see this error occur roughly once per week.  The consequence of the error is large so when it happens it is a crisis.

**What is the expected behavior?**

It is expected that the lock timeout added in the original issue would close the connection and free the mutex if the timeout expires.  That is apparently not happening.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
SQL Server Standard (and Enterprise) 2019, Windows Server 2016.  
The problem has occurred sporatically since installing the sp_Blitz job about 3 months back.

BrentOzar:
So if I understand you correctly, the SET LOCK TIMEOUT in the query isn't being honored by Microsoft SQL Server.

What steps would you like me to take?

sqlslinger:
[...]

PR #3221, opened 2023-01-30
sp_BlitzFirst SET LOCK_TIMEOUT 

sqlslinger:
SET LOCK_TIMEOUT 1000 when APPLY sys.dm_exec_query_statistics_xml to prevent mutex issues.  Set GitHub #3210

### DIFF
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -2293,8 +2293,9 @@ If one of them is a lead blocker, consider killing that query.'' AS HowToStopit,
 				  CROSS APPLY sys.dm_exec_query_plan(r.plan_handle) AS qp ';
 
 				IF EXISTS (SELECT * FROM sys.all_objects WHERE name = 'dm_exec_query_statistics_xml')
-					SET @StringToExecute = @StringToExecute + N' OUTER APPLY sys.dm_exec_query_statistics_xml(s.session_id) qs_live ';
-				  
+				/* GitHub #3210 */
+					SET @StringToExecute = N'
+                   SET LOCK_TIMEOUT 1000 ' + @StringToExecute + N' OUTER APPLY sys.dm_exec_query_statistics_xml(s.session_id) qs_live ';
 				  
 				SET @StringToExecute = @StringToExecute + N';
 
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -1794,7 +1794,7 @@ BEGIN
                 N'S',
                 N'IS'
             )
-        AND (dow.database_id = @DatabaseName OR @DatabaseName IS NULL)
+        AND (dow.database_id = @DatabaseId OR @DatabaseName IS NULL)
         AND (dow.event_date >= @StartDate OR @StartDate IS NULL)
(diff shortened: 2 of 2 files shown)


sp_BlitzFirst.sql
```diff
@@ -2293,8 +2293,9 @@ If one of them is a lead blocker, consider killing that query.'' AS HowToStopit,
 				  CROSS APPLY sys.dm_exec_query_plan(r.plan_handle) AS qp ';
 
 				IF EXISTS (SELECT * FROM sys.all_objects WHERE name = 'dm_exec_query_statistics_xml')
-					SET @StringToExecute = @StringToExecute + N' OUTER APPLY sys.dm_exec_query_statistics_xml(s.session_id) qs_live ';
-				  
+				/* GitHub #3210 */
+					SET @StringToExecute = N'
+                   SET LOCK_TIMEOUT 1000 ' + @StringToExecute + N' OUTER APPLY sys.dm_exec_query_statistics_xml(s.session_id) qs_live ';
```
BrentOzar, line 2298: Errr, that can't work. This puts SET LOCK_TIMEOUT in the middle of a query. Note what happens next in the query - it's an outer apply. That won't even compile.

You can see what I mean by trying to run:

SELECT * FROM sys.databases d
SET LOCK_TIMEOUT 1000
OUTER APPLY sys.dm_exec_query_statistics_xml(d.database_id) qs_live;

That doesn't even parse. So... with all due respect, I don't think you understand the problem you're trying to solve, so let's call this one off.

sqlslinger:
Brent,

Are you sure?  Look again.  I added the SET LOCK_TIMEOUT 1000 *before* the
existing @StringToExecute, for exactly the reason that it already had data
in it.

So if @StringToExecute contained 'SELECT * FROM sys.databases d', the
result of the statement would be

SET LOCK_TIMEOUT 1000
SELECT * FROM sys.databases d
OUTER APPLY sys.dm_exec_query_statistics_xml(d.database_id) qs_live;

As I said, this is compiled and working.

Vince

sp_BlitzFirst.sql
sqlslinger: @BrentOzar,

Can you please reopen this issue, or comment on my response?

>Brent,
>
>Are you sure?  Look again.  I added the SET LOCK_TIMEOUT 1000 *before* the
>existing @StringToExecute, for exactly the reason that it already had data
>in it.
>
>So if @StringToExecute contained 'SELECT * FROM sys.databases d', the
>result of the statement would be
>
>SET LOCK_TIMEOUT 1000
>SELECT * FROM sys.databases d
>OUTER APPLY sys.dm_exec_query_statistics_xml(d.database_id) qs_live;
>
>As I said, this is compiled and working.
>
>Vince

BrentOzar:
Sorry about that! I misunderstood the code. Just a gentle suggestion for next time - comments sure do help when folks are trying to understand your code. ;-)

BrentOzar (commented):
Sorry about that! I misunderstood the code. Just a gentle suggestion for next time - comments sure do help when folks are trying to understand your code. ;-)

PR #3223, opened 2023-02-09
Update sp_BlitzCache.sql

sm8680:
Fixed 3 case sensitive issues on column object_id for table sys.all_columns switching it from upper case to lower case to match the column in the table.


BrentOzar (approved):
Thanks for the pull request! Looks good, merging into the dev branch. Will be in the next monthly release with credit to you in the release notes.

PR #3240, opened 2023-03-05
Added @FileExtensionDiff param. Fixes #3234.

### LINKED ISSUE #3234: sp_DatabaseRestore - support @FileExtensionDiff parameter
sp_DatabaseRestore will only restore DIFF backups if they use the *.bak extension. If you use the @FileExtensionDiff = 'dif' or something similar during Ola's backups, the diffs are ignored during restore.

I would like sp_DatabaseRestore to support the same @FileExtensionDiff parameter that DatabaseBackup does, so that I do not have to tweak each new version of sp_DatabaseRestore :)

You could also modify the filter "WHERE BackupFile LIKE N'%.bak'" to just include ".dif" if we think that is the most common alternative to ".bak", but this seems like a poor solution.

Yes I have a modified version of 8.13 working.

spurdata:
Tested on SQL 2019 CU#19.
Very open to feedback, this is my first pull request.
Thanks!

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -31,6 +31,7 @@ ALTER PROCEDURE [dbo].[sp_DatabaseRestore]
 	@DatabaseOwner sysname = NULL,
 	@SetTrustworthyON BIT = 0,
     @Execute CHAR(1) = Y,
+	@FileExtensionDiff NVARCHAR(128) = NULL,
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good, works on my machine. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3241, opened 2023-03-06
Issue 3239: Fix dm_exec_query_profiles arithmetic overflow error in check 45

SQLLambert:
Check 45 looking for inaccurate query estimates was checking if the actual row count (row_count in dm_exec_query_profiles) was greater than estimate_row_count*10000. This was generating arithmetic overflow errors on queries containing estimates close to the bigint limit. Change the calculation to compare the row count/10000 to the estimate instead - same result but no overflow.


BrentOzar (approved):

PR #3266, opened 2023-05-01
sp_BlitzWho Increase accuracy of the grant_time column to match the rest of the date time columns

### LINKED ISSUE #3265: sp_BlitzWho Increase accuracy of the grant_time column to match the rest of the date time columns
**Version of the script**
8.14

If grant_time is available, it will be in the following format - May  1 2023 12:22PM
Not a bug necessarily 
Have the grant_time in a yyyy-mm-dd hh:mi:ss.mmm format when available
Applies to all versions.

Making a pull request for this right after filing this issue.

VladDBA:
Increased accuracy of the grant_time column to match the rest of the date time columns (i.e. login_time, request_time, run_date) Original format  `May  1 2023 12:22PM`, new `2023-05-01 12:53:27.290`
Closes #3265

### DIFF
--- a/sp_BlitzWho.sql
+++ b/sp_BlitzWho.sql
@@ -219,7 +219,7 @@ IF @OutputDatabaseName IS NOT NULL AND @OutputSchemaName IS NOT NULL AND @Output
 	[last_used_threads] [bigint] NULL,
 	[min_used_threads] [bigint] NULL,
 	[max_used_threads] [bigint] NULL,
-	[grant_time] [varchar](20) NULL,
+	[grant_time] [varchar](23) NULL,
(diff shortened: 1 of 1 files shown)


VladDBA:
Hi Brent,

I did modify the table creation command to use the new width for the grant_time column, but I failed to take into account scenarios with pre-existing tables.  

Regarding the reason: it's just my liking of uniformity between date time columns, so there is no actual benefit or fix to a problem.
I guess you can cancel the pull request. 

Thank you and have a great weekend!

BrentOzar:
Gotcha. Thanks anyway, and you have a good weekend too!

Issue #3267, opened 2023-05-01
sp_DatabaseRestore: Option to fixup orphan database users


sqlslinger:
**Is your feature request related to a problem? Please describe.**
A cross-server database restore needs to fixup database_principals that do not link correctly to server_principals.

**Describe the solution you'd like**
New Option @FixOrphanedUsers nchar(1) = 'Y'.  If 'Y', once restore is complete, sets database_principals.principal_id to the value of server_principals.principal_id where database_principals.name = server_principals.name.

**Describe alternatives you've considered**
Alternatively, this could be a separate procedure.  But it really only has utility after a cross-server database restore, so it seems to be a natural fit in a restore procedure.

**Are you ready to build the code for the feature?**
Yes, the code is not difficult and we have coded it in existing procedures.

PR #3270, opened 2023-05-03
sp_DatabaseRestore: Add @FixOrphanUsers option

sqlslinger:
Implements issue #3267
Fix database_principals that do not match server_principals after a cross-server restore.

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -11,10 +11,10 @@ ALTER PROCEDURE [dbo].[sp_DatabaseRestore]
     @MoveDataDrive NVARCHAR(260) = NULL, 
     @MoveLogDrive NVARCHAR(260) = NULL, 
     @MoveFilestreamDrive NVARCHAR(260) = NULL,
-	@MoveFullTextCatalogDrive NVARCHAR(260) = NULL, 
-	@BufferCount INT = NULL,
-	@MaxTransferSize INT = NULL,
-	@BlockSize INT = NULL,
+    @MoveFullTextCatalogDrive NVARCHAR(260) = NULL, 
+    @BufferCount INT = NULL,
+    @MaxTransferSize INT = NULL,
+    @BlockSize INT = NULL,
     @TestRestore BIT = 0, 
     @RunCheckDB BIT = 0, 
     @RestoreDiff BIT = 0,
@@ -27,15 +27,16 @@ ALTER PROCEDURE [dbo].[sp_DatabaseRestore]
     @StopAt NVARCHAR(14) = NULL,
     @OnlyLogsAfter NVARCHAR(14) = NULL,
     @SimpleFolderEnumeration BIT = 0,
-	@SkipBackupsAlreadyInMsdb BIT = 0,
-	@DatabaseOwner sysname = NULL,
-	@SetTrustworthyON BIT = 0,
+    @SkipBackupsAlreadyInMsdb BIT = 0,
+    @DatabaseOwner sysname = NULL,
+    @SetTrustworthyON BIT = 0,
+    @FixOrphanUsers BIT = 0,
     @Execute CHAR(1) = Y,
-	@FileExtensionDiff NVARCHAR(128) = NULL,
+    @FileExtensionDiff NVARCHAR(128) = NULL,
     @Debug INT = 0, 
     @Help BIT = 0,
     @Version     VARCHAR(30) = NULL OUTPUT,
-	@VersionDate DATETIME = NULL OUTPUT,
+    @VersionDate DATETIME = NULL OUTPUT,
     @VersionCheckMode BIT = 0
 AS
 SET NOCOUNT ON;
@@ -1374,7 +1375,6 @@ BEGIN
 		AND REPLACE( RIGHT( REPLACE( fl.BackupFile, RIGHT( fl.BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( fl.BackupFile ) ) ), '' ), 16 ), '_', '' ) > @StopAt
 		ORDER BY BackupFile;
 	END
-	
 	IF @ExtraLogFile IS NULL
 	BEGIN
 		DELETE fl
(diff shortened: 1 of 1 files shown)


sp_DatabaseRestore.sql
```diff
@@ -27,15 +27,16 @@ ALTER PROCEDURE [dbo].[sp_DatabaseRestore]
     @StopAt NVARCHAR(14) = NULL,
     @OnlyLogsAfter NVARCHAR(14) = NULL,
     @SimpleFolderEnumeration BIT = 0,
-	@SkipBackupsAlreadyInMsdb BIT = 0,
-	@DatabaseOwner sysname = NULL,
-	@SetTrustworthyON BIT = 0,
+    @SkipBackupsAlreadyInMsdb BIT = 0,
+    @DatabaseOwner sysname = NULL,
+    @SetTrustworthyON BIT = 0,
+    @FixOrphanUsers BIT = 1,
```
BrentOzar, line 33: Can we default FixOrphanUsers to 0 instead of 1? I just don't wanna fire new behavior without the users being aware.

sp_DatabaseRestore.sql
```diff
@@ -1584,6 +1584,30 @@ IF @DatabaseOwner IS NOT NULL
 			END
 		END;
 
+-- Link a user entry in the sys.database_principals system catalog view in the restored database to a SQL Server login of the same name
+		IF @FixOrphanUsers = 1 AND @Execute = 'Y'
```
BrentOzar, line 1588: The way this is written, it will only generate the user changes if Execute = Y. If someone runs it with Execute = N so that they can review the changes first, OR if they wanna apply 'em by hand, this portion of the dynamic SQL won't show up.

Can you change it to build the dynamic SQL even if Execute = N? If not, is there a reason why? (I only did a quick scan of this.)

sp_DatabaseRestore.sql
```diff
@@ -1584,6 +1584,30 @@ IF @DatabaseOwner IS NOT NULL
...
+
+			SELECT @sql = 'USE ' + @RestoreDatabaseName + '; 
+			    INSERT #OrphanUsers
+				SELECT ''ALTER USER ['' + d.name + ''] WITH LOGIN = ['' + d.name + '']; ''
+				FROM sys.database_principals d
+				INNER JOIN master.sys.server_principals s ON d.name COLLATE DATABASE_DEFAULT = s.name COLLATE DATABASE_DEFAULT
+				WHERE d.type_desc = ''SQL_USER'' 
+				AND d.name != ''guest''
+				AND d.sid <> s.sid'
+			EXECUTE whqms.[dbo].[CommandExecute] @DatabaseContext = 'master', @Command = @sql, @CommandType = 'UPDATE', @Mode = 1, @DatabaseName = @UnquotedRestoreDatabaseName, @LogToTable = 'Y', @Execute = 'Y';
```
BrentOzar, line 1600: This looks like a hard-coded server reference to whqms. (Shout out to DefaultBlameAcceptor on the live stream for catching this.)

sp_DatabaseRestore.sql
```diff
@@ -1584,6 +1584,30 @@ IF @DatabaseOwner IS NOT NULL
...
 		END;
 
+-- Link a user entry in the sys.database_principals system catalog view in the restored database to a SQL Server login of the same name
+		IF @FixOrphanUsers = 1 AND @Execute = 'Y'
+			BEGIN
+			CREATE TABLE #OrphanUsers (SqlToExecute NVARCHAR(MAX));
+
+			SELECT @sql = 'USE ' + @RestoreDatabaseName + '; 
+			    INSERT #OrphanUsers
+				SELECT ''ALTER USER ['' + d.name + ''] WITH LOGIN = ['' + d.name + '']; ''
```
BrentOzar, line 1594: This one's kinda tricky, and I hate even mentioning it, but is it possible that we could have SQL injection here? If there are other similar issues in the code, don't worry about it, I just try to use QUOTENAME or whatever whenever I can.

sp_DatabaseRestore.sql
```diff
@@ -1584,6 +1584,30 @@ IF @DatabaseOwner IS NOT NULL
...
+				FROM sys.database_principals d
+				INNER JOIN master.sys.server_principals s ON d.name COLLATE DATABASE_DEFAULT = s.name COLLATE DATABASE_DEFAULT
+				WHERE d.type_desc = ''SQL_USER'' 
+				AND d.name != ''guest''
+				AND d.sid <> s.sid'
+			EXECUTE whqms.[dbo].[CommandExecute] @DatabaseContext = 'master', @Command = @sql, @CommandType = 'UPDATE', @Mode = 1, @DatabaseName = @UnquotedRestoreDatabaseName, @LogToTable = 'Y', @Execute = 'Y';
+			
+			DECLARE @DatabasePrincipalsToFix NVARCHAR(MAX)
+			SELECT @DatabasePrincipalsToFix = 'USE ' + @RestoreDatabaseName + '; '
```
BrentOzar, line 1603: I think this will break on database names with spaces, probably need to quotename this.

sp_DatabaseRestore.sql
sqlslinger: Yes that makes sense.

sp_DatabaseRestore.sql
sqlslinger: Good point and Quotename is easy enough to stick in.

sp_DatabaseRestore.sql
sqlslinger: Quotename it is.

sp_DatabaseRestore.sql
sqlslinger: @BrentOzar,  the fixup orphan users functionality requires that the database already be present, so the database principal sid can be compared to the login sid.  If @Execute = 'N', the restored database likely does not yet exist.  In that case we cannot print what the fixup would do.

BrentOzar:
I know you're going to laugh at this, but.... could we do it as dynamic SQL that gets built and executed after the restore?---Tiny glass keyboardTypos flowing like riversAs winter snow thawsOn May 8, 2023, at 10:34 AM, sqlslinger ***@***.***> wrote:﻿
@sqlslinger commented on this pull request.



In sp_DatabaseRestore.sql:
> @@ -1584,6 +1584,30 @@ IF @DatabaseOwner IS NOT NULL
 			END
 		END;
 
+-- Link a user entry in the sys.database_principals system catalog view in the restored database to a SQL Server login of the same name
+		IF @FixOrphanUsers = 1 AND @Execute = 'Y'

@BrentOzar,  the fixup orphan users functionality requires that the database already be present, so the database principal sid can be compared to the login sid.  If @execute = 'N', the restored database likely does not yet exist.  In that case we cannot print what the fixup would do.

sp_DatabaseRestore.sql
sqlslinger: @RestoreDatabaseName is already sql-injection proof by this code on line 522:
```sql
SET @RestoreDatabaseID = DB_ID(@RestoreDatabaseName);
SET @RestoreDatabaseName = QUOTENAME(@RestoreDatabaseName);
SET @UnquotedRestoreDatabaseName = PARSENAME(@RestoreDatabaseName,1);
```

sp_DatabaseRestore.sql
sqlslinger: OK @BrentOzar, it took me a while but I got what you were requesting and how this can be done using dynamic sql.  If run with @Execute = 'N', you'll get output like the following after the RESTORE DATABASE commands:

```sql

-- Fixup Orphan Users by setting database user sid to match login sid
DECLARE @FixOrphansSql NVARCHAR(MAX), @OrphanUsers TABLE (SqlToExecute NVARCHAR(MAX));
USE [VinceTestRestored];

INSERT @OrphanUsers
SELECT 'ALTER USER [' + d.name + '] WITH LOGIN = [' + d.name + ']; '
	FROM  sys.database_principals d
	INNER JOIN master.sys.server_principals s ON d.name COLLATE DATABASE_DEFAULT = s.name COLLATE DATABASE_DEFAULT
	WHERE d.type_desc = 'SQL_USER'
		AND d.name NOT IN ('guest','dbo')
		AND d.sid <> s.sid
	ORDER BY d.name;

SELECT @FixOrphansSql = (SELECT SqlToExecute AS [text()] FROM @OrphanUsers FOR XML PATH (''), TYPE).value('text()[1]','NVARCHAR(MAX)');

IF @FixOrphansSql IS NULL 
	PRINT 'No orphan users require a sid fixup.';
ELSE
BEGIN
	PRINT 'Fix Orphan Users: ' + @FixOrphansSql;
	EXECUTE(@FixOrphansSql);
END;
```
Executing that code will either print and execute the ALTER USER statements needed, or print the line "No orphan users require a sid fixup."

BrentOzar:
There was a conflict in the code between your two pull requests, but I think it was just spacing. Merged it into the dev branch, but can you test the dev branch version to make sure it works on your end? It'll be in the June release either way, just wanted to give you the heads up. Thanks!

PR #3271, opened 2023-05-04
Update sp_Blitz.sql

sm8680:
Updated SQL Server 2016 from SP2 to SP3.


MichelZ:
Just to add some context so this is easier to digest I guess:

- sp_Blitz checks for unsupported SP versions
- The current check for SQL Server 2016 checks for SP2
- SP2 has become unsupported as of 2022/10/11 according to [Microsoft](https://learn.microsoft.com/en-us/lifecycle/products/sql-server-2016) and [Brent](https://sqlserverupdates.com/sql-server-2016-updates/)
- This change makes sure that only SP3+ is considered supported (SP3 = Build 13.0.6300.2)

@sm8680 you might want to include these details so it's easier for maintainers to review your request, after all they only have limited time to do these things :)

PR #3274, opened 2023-05-10
Update unsupported SQL Versions

MichelZ:
- Update script so that it considers SQL 2012 as unsupported [Microsoft Product Lifecycle SQL 2012](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-sql-server-2012)
- Update script so that it considers SQL 2016 < SP3 as unsupported [Microsoft Product Lifecycle for SQL 2016](https://learn.microsoft.com/en-us/lifecycle/products/sql-server-2016)

This also incorporates the change from PR #3271


BrentOzar:
Thanks for the pull request! Merged into the dev branch, will be in the June release with credit to you in the release notes.

PR #3289, opened 2023-06-19
Add @KeepCdc parameter

sqlslinger:
Add the KEEP_CDC option to the RESTORE DATABASE ... WITH RECOVERY command.


sqlslinger:
@BrentOzar, could you please review this change to allow for the KEEP_CDC option?

PR #3300, opened 2023-07-02
Add some permissions checks to sp_Blitz

### LINKED ISSUE #3292: sp_Blitz - Skip checks where the login doesn’t have enough permissions
I’m doing some work in a high security environment where most checks run but some things like hitting sys.traces and DBCC commands fail, etc. 

Check to see if the user is a sysadmin/has access to view traces before running some checks. 

Embarrassment 

Yep, have a prototype 😮‍💨

erikdarlingdata:
Closes #3292

This feels like a somewhat naïve set of checks. I don't know a ton about security. It may clash when someone uses `EXECUTE AS` or signs the procedure for execution for lower-privileged users. I just don't have the chops to test all that to a satisfactory degree.

It's also incomplete at the moment, because I need to round up commands that touch other databases we may not have read permissions in and exclude those. 

I fully expect this to get rejected, but it got me error-free runs in a tightly locked down SQL Server environment.

If anyone has feedback, I'm happy to take it.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -186,10 +186,127 @@ AS
             ,@CurrentComponentVersionCheckModeOK     BIT
             ,@canExitLoop                            BIT
             ,@frkIsConsistent                        BIT
-			,@NeedToTurnNumericRoundabortBackOn      BIT;
+			,@NeedToTurnNumericRoundabortBackOn      BIT
+			,@sa bit = 1
+			,@SUSER_NAME sysname = SUSER_SNAME()
+			,@SkipDBCC bit = 0
+			,@SkipTrace bit = 0
+			,@SkipXPRegRead bit = 0
+			,@SkipXPFixedDrives bit = 0
+			,@SkipXPCMDShell bit = 0
+			,@SkipMaster bit = 0
+			,@SkipMSDB bit = 0
+			,@SkipModel bit = 0
+			,@SkipTempDB bit = 0
+			,@SkipValidateLogins bit = 0;
+
+			DECLARE
+			    @db_perms table
+            (
+			    database_name sysname,
+				permission_name sysname
+			);
+
             
             /* End of declarations for First Responder Kit consistency check:*/
         ;
+
+		/*Starting permissions checks here, but only if we're not a sysadmin*/
+		IF
+        (
+            SELECT 
+                sa = 
+                    ISNULL
+                    (
+                        IS_SRVROLEMEMBER(N'sysadmin'), 
+                        0
+                    )
+        ) = 0
+        BEGIN
+		    SET @sa = 0; /*Setting this to 0 to skip DBCC COMMANDS*/
+
+		    IF NOT EXISTS
+		    (
+		        SELECT 
+                    1/0 
+                FROM sys.fn_my_permissions(NULL, NULL) AS fmp 
+                WHERE fmp.permission_name = N'VIEW SERVER STATE'
+		    )
+			BEGIN
+			    RAISERROR('The user %s does not have VIEW SERVER STATE permissions.', 0, 11, @SUSER_NAME) WITH NOWAIT;
+				RETURN;
+			END; /*If we don't have this, we can't do anything at all.*/
+
+            IF NOT EXISTS
+            (
+                SELECT
+                    1/0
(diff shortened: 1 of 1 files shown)


BrentOzar:
I love this idea. I did a really quick code review & test, and it worked at first glance. I need to check a few other things before merging the pull request:

* Test that it returns the same number of rows as the prior version (just to make sure it's not accidentally muting something it shouldn't)
* Test on Amazon RDS
* Test on Azure SQL DB Managed Instance
* Put "Starting permissions tests" in the Messages tab to troubleshoot long-running queries

erikdarlingdata:
Yeah, I don't know how often it comes up for other people, but I've run into this a few times. When you've done some initial checks, I'll expand things to:
* insert messages when things are skipped because of permissions
* expand the checks to user databases
* expand the checks to system database queries where we don't have permissions

If you come up with anything else, let me know.

sp_Blitz.sql
```diff
@@ -186,11 +186,128 @@ AS
...
+			    permission_name
+			)
+            SELECT
+                database_name =
+				    DB_NAME(d.database_id),
+                fmp.permission_name
+            FROM sys.databases AS d
+            CROSS APPLY fn_my_permissions(d.name, 'DATABASE') AS fmp
+            WHERE fmp.permission_name = N'SELECT'
+            AND   d.database_id < 5; /*Databases where we don't have read permissions*/
```
BrentOzar, line 308: Why is this only filtered for system databases? I think you want it for all databases because you wanna skip the ones you don't have access to, right?

erikdarlingdata: Yep, totally. I think I had done that to scope the initial skips to system database stuff, but long term it shouldn’t be there at all.

sp_Blitz.sql
```diff
@@ -331,6 +448,66 @@ AS
...
+			    1/0
+			FROM @db_perms AS dp
+			WHERE dp.database_name = DB_NAME(d.database_id)
+		);
+
+		/*Skip individial checks where we don't have permissions*/
+		INSERT #SkipChecks (DatabaseName, CheckID, ServerName)
+		SELECT
+		    v.*
+		FROM (VALUES(NULL, NULL, 29)) AS v (DatabaseName, CheckID, ServerName) /*Looks for user tables in model*/
```
BrentOzar, line 472: This is inserting the value 29 into the ServerName column, which is causing other problems, heh.

BrentOzar: Also, I think it's firing a false positive - I'm finding this live as I troubleshoot stuff, trying to track down why this new version is returning way less rows than the current version. I'm SA on this server, and it shouldn't be inserting a row for this check.

sp_Blitz.sql
```diff
@@ -331,6 +448,66 @@ AS
...
+		    DatabaseName
+		)
+		SELECT
+		    DB_NAME(d.database_id)
+		FROM sys.databases AS d
+		WHERE NOT EXISTS
+		(
+		    SELECT
+			    1/0
+			FROM @db_perms AS dp
```
BrentOzar, line 464: This section relies on @db_perms being populated, but that only has rows if the user is NOT sa. The population is done in an IF branch that starts up at line 215.

So if you ARE sa, the line 215 branch never fires, and it looks like you don't have any permissions, heh.

sp_Blitz.sql
```diff
@@ -186,11 +186,128 @@ AS
...
+                SELECT
+                    1/0
+                FROM fn_my_permissions(N'sp_validatelogins', N'OBJECT') AS fmp
+                WHERE fmp.permission_name = N'EXECUTE'
+            )
+            BEGIN
+                SET @SkipValidateLogins = 1;
+            END; /*Need execute on sp_validatelogins*/
+
+			INSERT
```
BrentOzar, line 295: Need to move this one insert up outside of the IF branch that starts around line 215, because it needs to run even if you're SA. (Since we use it down later to populate the list of databases to skip.)

BrentOzar (commented):
This new version produced way less rows on mine where I'm SA, but I narrowed down the root cause. Stopping for now, but will edit this later today to fix this easy stuff, and then test on RDS. Looking good overall though!

BrentOzar (approved):
Works now.

PR #3316, opened 2023-07-27
Correct the SARG brackets for Replication In Use check in sp_Blitz

DavidSchanzer:
Add bracketing to compound logical condition to correct behavior for the Replication In Use check.


BrentOzar (approved):
Thanks for the pull request. Makes sense. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3317, opened 2023-07-29
#3314 sp_BlitzCache high percentage of duplicate plans

### LINKED ISSUE #3314: sp_BlitzCache: In high DB count systems "duplicate" query plans in the Resource database overshadow duplicate plan count
**Version of the script**
Version = '8.15', VersionDate = '20230613'

When counting "duplicate" plans in the cache the script groups by the dbid plan attribute (line 1629 "pa.value") and that's fine for any queries _other than_ those based in the Resource database (dbid 32767). In that case one must also look at the dbid_execute attribute as the query is based in the Resource database, but it runs in each of the individual databases. This results in counting the same query multiple times, once for each database on the system. On low database-count systems that's not a huge deal, but on systems with 100s+ databases the numbers start to be real and it throws out the duplicate plan statistics. 

Run sp_BlitzCache on a machine with 100s of databases. You should see a Many Duplicate Plans finding (check 999) with a crazy duplicate percentage. e.g., "You have 32830 plans in your cache, and 254.00% are duplicates with more than ..."  

Technically, the script should deduplicate on both the dbid and dbid_execute attributes, but more simply, it could just omit plans in the Resource database since we're not going to "fix" those plans anyway. 

Microsoft SQL Server 2019 (RTM-CU21) (KB5025808) - 15.0.4316.3 (X64)  -- Though I expect it would be this way on any version. I initially found the problem on script 20221013, but upgraded to confirm it's still present.

PasswordIsMyPassword:
Fixes #3314

We had similar issue where it was showing 3 million duplicate plans and the percentage was over 2000%
We originally looked at this because sp_blitzFirst was taking over an hour to run.
Adding this change to the procedure brings the time down to under 1 minute and the duplicate plan percentage went from 2400% to 93%.

Here are our findings.

So why is the query slow on the server, and why does it return query_plan_hashes in millions? 
That’s because of the LEFT JOIN between sys.dm_exec_query_stats and sys.dm_exec_procedure_stats ON sql_handle.

The left input table in the LEFT JOIN i.e., sys.dm_exec_query_stats has thousands of duplicate sql_handles, and so does the right input sys.dm_exec_procedure_stats. Therefore, for each matching sql_handle on the left, it returns every matching row from the right. Hence, this “Cartesian product” of the join results in the output in millions. Apparently, sql_handle is not the best choice to join on as it contains duplicates in the sys.dm_exec_procedure_stats view, unlike plan_handle, which is unique in dm_exec_procedure_stats.

 The Fix? Simply replace the column to join on from sql_handle to plan_handle, and voila. The same original query with only the change to the join column ran in under 1 second compared to the 9+ minutes it ran with the sql_handle as the join column, and it returns an accurate plan_hash count too. Since all plan_handles are unique in dm_exec_procedure_stats (i.e. right side of the join), there’s only one match in the right table, thus avoiding duplicates.

When I put in the original change for this ticket, the run is still long and the duplicate plan percentage is still over 2400, so that change did not fix our problem. Maybe both changes are needed? 
Sorry if I did the change request wrong...i'm new to github.

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -1616,8 +1616,7 @@ WITH total_plans AS
         SELECT
 		    COUNT_BIG(qs.query_plan_hash) AS duplicate_plan_hashes
         FROM sys.dm_exec_query_stats qs
-        LEFT JOIN sys.dm_exec_procedure_stats ps 
-            ON qs.plan_handle = ps.plan_handle
+        LEFT JOIN sys.dm_exec_procedure_stats ps ON qs.plan_handle = ps.plan_handle
         CROSS APPLY sys.dm_exec_plan_attributes(qs.plan_handle) pa
         WHERE pa.attribute = N'dbid'
         AND   qs.query_plan_hash <> 0x0000000000000000
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -1617,7 +1617,7 @@ WITH total_plans AS
 		    COUNT_BIG(qs.query_plan_hash) AS duplicate_plan_hashes
         FROM sys.dm_exec_query_stats qs
         LEFT JOIN sys.dm_exec_procedure_stats ps 
-            ON qs.sql_handle = ps.sql_handle
+            ON qs.plan_handle = ps.plan_handle
         CROSS APPLY sys.dm_exec_plan_attributes(qs.plan_handle) pa
         WHERE pa.attribute = N'dbid'
         AND   qs.query_plan_hash <> 0x0000000000000000


BrentOzar:
Conceptually, I think I understand this because sys.dm_exec_query_stats keys on sql_handle AND statement_start_offset AND statement_end_offset, whereas sys.dm_exec_procedure_stats is only keying on sql_handle.

I've set up a repro query of the problem that sp_BlitzCache is trying to solve (unparameterized queries duplicated in the plan cache), NOT the query that this issue is addressing, and I'll put that in the main issue. This pull request works the same way the old code worked in this scenario, so I'm okay with the fix, although I'm still nervous about it, hahaha.

PR #3318, opened 2023-08-09
Update PercentMemoryGrantUsed in sp_BlitzCache.sql

### LINKED ISSUE #3313: sp_BlitzCache - Incorrect Percent Memory Grant Used Calculations
**Version of the script**
- 8.15 - 20230613

- PercentMemoryGrantUsed is calculated as:
```sql
CAST(ISNULL(NULLIF(( max_used_grant_kb * 1.00 ), 0) / NULLIF(min_grant_kb, 0), 0) * 100. AS MONEY) AS PercentMemoryGrantUsed
```
- This doesn't calculate the percent of memory grant that was used by the query because the two numbers aren't related.  There's also no reason to believe that the maximum/minimum used grant reflects the amount of memory used when the maximum/minimum grant was granted.

- Run SP

- A better calculation would use `total_used_grant_kb` / `total_grant_kb` as this will cover all memory usage since SP was compiled

- All versions of SQL Server and All OS's are affected.  This appears to have been how the metric has always been calculated.

mdpenguin:
Update to calculate as total_used_grant_kb / total_grant_kb. Fixes #3313.

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -2324,7 +2324,7 @@ BEGIN
            max_grant_kb AS MaxGrantKB,
            min_used_grant_kb AS MinUsedGrantKB,
            max_used_grant_kb AS MaxUsedGrantKB,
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request. Makes sense. Merging into the dev branch, will be in the next release with credit to you in the release notes.

Issue #3312, opened 2023-07-25
sp_BlitzQueryStore - Missing results related to PSPO executions


sqljared:
**Version of the script**
SELECT @Version = '8.15', @VersionDate = '20230613';

**What is the current behavior?**
If sp_BlitzQueryStore is executed with the @StoredProcName parameter, the results will not include any executed queries where Parameter Sensitive Procedure Optimization is used. 

- Any parent queries are excluded because their plans (the dispatch plans) do not execute. The query on line 1104 includes query_store_runtime_stats, so the query will filter out at that time.
- Variant queries have an object_id of 0 like ad hoc queries and won't be included by the filter.

Here is an example of the behavior from sp_BlitzQueryStore in a local example:
![image](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/assets/47091477/54977e40-91a9-41df-921a-fe0f8187ddaa)

For comparison, here is a query I ran to see all queries for this same procedure and any variant plans associated:
`WITH qpsp AS (
    SELECT
        ISNULL(qv.query_id,qsq.query_id) AS query_id,
        ISNULL(qvp.plan_id,qsp.plan_id) AS plan_id,
        ISNULL(qvp.query_plan,qsp.query_plan) AS query_plan,
        ISNULL(qv.query_text_id,qsq.query_text_id) AS query_text_id,
        ISNULL(qvp.plan_type_desc,qsp.plan_type_desc) AS plan_type_desc
    FROM sys.query_store_query qsq
    INNER JOIN sys.query_store_plan qsp
        ON qsp.query_id = qsq.query_id
    LEFT JOIN sys.query_store_query_variant vr
        ON vr.parent_query_id = qsq.query_id
    LEFT JOIN sys.query_store_query qv
        ON qv.query_id = vr.query_variant_query_id
    LEFT JOIN sys.query_store_plan qvp
        ON qvp.query_id = qv.query_id
    WHERE
        qsq.object_id = OBJECT_ID('Sales.GetOrders')
)
SELECT
    qpsp.query_id,
    qpsp.plan_id,
    qpsp.plan_type_desc,
    CAST(qpsp.query_plan as XML),
    qt.query_sql_text,
    rsi.end_time,
    (rs.avg_duration * rs.count_executions) as total_duration,
    rs.avg_duration,
    rs.count_executions,
    rs.avg_cpu_time,
    rs.avg_logical_io_reads,
[...]

PR #3322, opened 2023-08-18
PSPO fix for object_id issues in sp_QueryBlitzStore

sqljared:
This should resolve both issues documented in the [related bug](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3312).

The first commit adds a block to detect if the database config setting for PARAMETER_SENSITIVE_PLAN_OPTIMIZATION is present and enabled. 
If PSPO is enabled and the filter for @StoredProcName is used, we will include queries whose object_id = 0, but have a parent query that is related to the object matching @StoredProcName.

The second commit adds a block that attempts, if PSPO is enabled, to update #working_metrics.proc_or_function_name with the name of the parent query's object for any rows where proc_or_function_name IS NULL. The proc_or_function_name section does not depend on whether @StoredProcName is used; there may be variant queries in the temp table regardless.

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -330,6 +330,28 @@ SET @msg = N'New query_store_runtime_stats columns ' + CASE @new_columns
 							   END;
 RAISERROR(@msg, 0, 1) WITH NOWAIT;
 
+/*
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks good, works beautifully. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3332, opened 2023-09-13
keep the QueryText as it is for not changing line comment context

eschnepel:
this also includes changes to move the if block for @ExportToExcel = 1 to a later stage and do this for all rows at once. I accidentally included also a fix for a copy and paste error (and not opened an issue of it's own) where avg spills data got the update for the Pattern = ''avg memory grant''.


BrentOzar:
@eschnepel howdy, and thanks for the pull request. 

Some quick advice on pull requests:
* Open an issue first, explaining the problem you're trying to solve
* Don't change code "just because" - we need to be solving an actual problem, because change = risk, and you're asking other folks to take time to review your code and test it, plus it may break the way people rely on the code
* If there are multiple problems, open one issue per problem
* Open one pull request per issue

This helps people understand what changes you're making, and why. Right now, for example, I don't understand why we're trying to change the existing logic around query text. You might have a great reason, don't get me wrong - I just have no idea what that reason is.

I'm going to close this pull request for now, and I'll open a separate one about the avg spills & memory grant mixup, because that makes sense.

PR #3333, opened 2023-09-13
added return statement to avoid going into OutputResultsToTable block

eschnepel:
... otherwise the last result (e.g. spills) would be recorded twice into the output table.


BrentOzar:
I think you misunderstood the root cause - spills was being added twice due to a labeling problem, I think. Can you try with the latest dev branch, and if you're still experiencing the problem, can you open a Github issue with the exact steps to reproduce the problem? Thanks!

Issue #3334, opened 2023-09-14
sp_Blitz Fails when tricked in thinking you have read permissions in the model database (Check 29)


Montro1981:
**Version of the script**
Version 8.16
VersionDate 2023-08-20

**What is the current behavior?**
During the "declarations for First Responder Kit consistency check" step, a script is executed to retrieve all the databases where we have SELECT permissions, the model database can be included in this list.
Since we have SELECT permissions, the script assumes that we can also view the objects in the current security context.

**If the current behavior is a bug, please provide the steps to reproduce.**
Create a user on the SQL server with VIEW SERVER STATE rights that is not mapped to the master or model databases.
Execute: `SELECT 1/0 FROM model.sys.objects;`
This should fail.

**What is the expected behavior?**
A sanity check must be included to verify that we can view the objects in the model database. If we can't view the objects Check 29 must be added to the #SkipChecks table

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
SQL server 2017 (Web)

BrentOzar:
I closed the pull request because it added a confusing parameter. If you're okay with just skipping model, then use the existing @SkipModel parameter. Otherwise, to solve the problem you're discussing here, your original design would be correct:

> A sanity check must be included to verify that we can view the objects in the model database. If we can't view the objects Check 29 must be added to the #SkipChecks table

If you don't have the time to do that, I totally understand - just let me know and I can close this issue.

BrentOzar:
Thanks for the pull request! Looks good, merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3335, opened 2023-09-14
Issue #3334: sp_Blitz Fails because of permissions

Montro1981:
Fix for Issue #3334, Added another bit variable @SkipModelCheck which defaults to 0. Added a check block that sets @SkipModelCheck to 1 in case we don't have permissions in the model or if we are tricked into thinking we have access but in fact we can't view the objects in the model database.

If @SkipModelCheck = 1 check 29 will be added to the #SkipChecks table.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -198,7 +198,8 @@ AS
 			,@SkipMSDB bit = 0
 			,@SkipModel bit = 0
 			,@SkipTempDB bit = 0
-			,@SkipValidateLogins bit = 0;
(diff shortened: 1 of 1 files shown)


BrentOzar:
@SkipModel is already a parameter for sp_Blitz, so this feels confusing.

We'd definitely welcome a pull request that lines up with your original issue notes, though:

> A sanity check must be included to verify that we can view the objects in the model database. If we can't view the objects Check 29 must be added to the #SkipChecks table

PR #3340, opened 2023-09-20
Issue #3334: sp_Blitz Fails because of permissions

Montro1981:
Brent's comment from pull request #3335 
> @SkipModel is already a parameter for sp_Blitz, so this feels confusing.
> 
> We'd definitely welcome a pull request that lines up with your original issue notes, though:

My apologies I hadn't noticed the `@SkipModel` parameter.

Changed the code to use the `@SkipModel`  parameter and to the respect the `@SkipModel` value if the caller has set it to 1 by hand, as the old version of the code could override the passed value if we could access the objects in the model database. 
The rest stayed pretty much the same so in the permissions check we still check if we can access the model database and the objects within it.

If `@SkipModel = 1` check 29 will be added to the #SkipChecks table.

Fix for issue #3334

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -306,32 +306,35 @@ AS
             BEGIN
                 SET @SkipValidateLogins = 1;
             END; /*Need execute on sp_validatelogins*/
-
-			IF EXISTS
-            (
-                SELECT 1/0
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

Issue #3348, opened 2023-09-22
[sp_Blitz documentation] Explain @Debug values in the procedure


Montro1981:
**Version of the script**
Version: 8.16
VersionDate: 20230820

**What is the current behavior?**
There is a @Debug parameter to pass to the procedure but there is no explanation of what the different values do.

**What is the expected behavior?**
There is an explanation of what the different values for @Debug do.
0 - Silent running
1 - Output messages per step
2 - Outputs dynamic SQLs

No functional changes are needed.

BrentOzar:
Thanks for the pull request. Looks good, merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3349, opened 2023-09-22
#3348 Added explanation for @debug in the help in sp_Blitz and updated Readme.md

Montro1981:
No functional changes, just added some explanation for @debug in the help in sp_Blitz and updated Readme.md


Montro1981:
Don't want to mix functional fixes with textual fixes

Issue #3350, opened 2023-09-22
[sp_Blitz] DBCC dbinfo fires even if CheckID 68 or 2 is in #SkipChecks


Montro1981:
**Version of the script**
Version: 8.16
VersionDate: 20230820

**What is the current behavior?**
If the user is not part of the sysadmin group `@sa` gets set to 0:
`SET @sa = 0; /*Setting this to 0 to skip DBCC COMMANDS*/`

This later ensures that check 68 and 69 are added to #SkipChecks
```tsql
INSERT #SkipChecks (DatabaseName, CheckID, ServerName)
SELECT
    v.*
FROM (VALUES(NULL, 68, NULL)) AS v (DatabaseName, CheckID, ServerName) /*DBCC command*/
WHERE @sa = 0;

INSERT #SkipChecks (DatabaseName, CheckID, ServerName)
SELECT
    v.*
FROM (VALUES(NULL, 69, NULL)) AS v (DatabaseName, CheckID, ServerName) /*DBCC command*/
WHERE @sa = 0;
```

But the code checking if `DDBC DBInfo` can be run has a logical error in it:
```tsql
--Reformated for readablity
/*
Check 233 is not in #BlitzResults
Check 2 is not in #SkipChecks
Check 68 is in #SkipChecks
*/

IF NOT EXISTS
(
    SELECT 1
    FROM #BlitzResults 
    WHERE CheckID = 223 AND URL = 'https://aws.amazon.com/rds/sqlserver/')
    AND
    (
        NOT EXISTS
        (
            SELECT  1
            FROM    #SkipChecks
            WHERE   DatabaseName IS NULL AND CheckID = 2
        ) OR NOT EXISTS
        (
            SELECT  1
            FROM    #SkipChecks
            WHERE   DatabaseName IS NULL AND CheckID = 68
        )
)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Make a user that is not sysadmin but has VIEW SERVER STATE rights, and doesn't have read rights in some of the system databases
Execute sp_blitz:
```tsql
EXEC [dbo].[sp_Blitz] 
    @CheckUserDatabaseObjects = 1 ,
    @CheckProcedureCache = 0 ,
    @OutputType = 'TABLE' ,
    @OutputProcedureCache = 0 ,
    @CheckProcedureCacheFilter = NULL,
    @CheckServerInfo = 1 ,
    @Debug = 1
```
[...]

PR #3357, opened 2023-09-26
#3350 DBCC DBinfo fires incorrectly

Montro1981:
Fix for issue #3350

The old code had an logical error in it:
```
Check 233 is not in #BlitzResults
Check 2 is not in #SkipChecks
Check 68 is in #SkipChecks

It became:
IF True (Check 233 not in #BlitzResults)
AND
(
    True (Check 2 not in #SkipChecks)
    OR
    False (Check 68 in #SkipChecks)
)
```
So the IF returned true and the check fired.
Basically the OR needed to be an AND, this allowed the code to be simplified to:

```tsql
IF NOT EXISTS
(
	SELECT	1/0 
	FROM	#BlitzResults 
	WHERE	CheckID = 223 AND URL = 'https://aws.amazon.com/rds/sqlserver/'
) AND NOT EXISTS
(
	SELECT  1/0
	FROM    #SkipChecks
	WHERE   DatabaseName IS NULL AND CheckID IN (2, 68)
)
```

Easier to read as well.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -1109,17 +1109,17 @@ AS
 				least one of the relevant checks is not being skipped then we can extract the
 				dbinfo information.
 				*/
-				IF NOT EXISTS ( SELECT 1 
-							FROM #BlitzResults 
-							WHERE CheckID = 223 AND URL = 'https://aws.amazon.com/rds/sqlserver/')
-					AND (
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Wow, the logic in the original one makes my head hurt. Merging into the dev branch, will be in the next release with credit to you in the release notes.

Issue #3356, opened 2023-09-26
[sp_Blitz] Checks with stored procedures sp_validatelogins, xp_regread, and xp_readerrorlog fail because of permissions


Montro1981:
**Version of the script**
Version 8.16
VersionDate 2023-08-20

**What is the current behavior?**
The `fn_my_permissions` function returns that we can execute the stored procedures `sp_validatelogins` and `xp_regread` but when we try to execute them we get errors that we have no permissions.
Also permissions for stored procedure `xp_readerrorlog` are not checked.

**If the current behavior is a bug, please provide the steps to reproduce.**
Create a user on the SQL server with VIEW SERVER STATE rights that is not sysadmin.

*sp_validatelogins*
```tsql
SELECT [entity_name], [subentity_name], [permission_name]
FROM fn_my_permissions(N'sp_validatelogins', N'OBJECT') AS fmp
WHERE fmp.permission_name = N'EXECUTE';
```

| entity_name       | subentity_name | permission_name |
|-------------------|----------------|-----------------|
| sp_validatelogins |                | EXECUTE         |

```tsql
EXEC sp_validatelogins;
/*
Msg 15247, Level 16, State 1, Procedure sp_validatelogins, Line 6 [Batch Start Line 0]
User does not have permission to perform this action.
*/
```

*xp_regread*
```tsql
SELECT [entity_name], [subentity_name], [permission_name]
FROM fn_my_permissions(N'xp_regread', N'OBJECT') AS fmp
WHERE fmp.permission_name = N'EXECUTE'
```

| entity_name | subentity_name | permission_name |
|-------------|----------------|-----------------|
| xp_regread  |                | EXECUTE         |

```tsql
EXEC xp_regread @rootkey = N'HKEY_LOCAL_MACHINE', @key = N'', @value_name = N'';
/*
Msg 22001, Level 16, State 1, Line 31
xp_regread() returned error 5, 'Access is denied.'
*/
```

*xp_readerrorlog*
```tsql
SELECT [entity_name], [subentity_name], [permission_name]
FROM fn_my_permissions(N'xp_readerrorlog', N'OBJECT') AS fmp
WHERE fmp.permission_name = N'EXECUTE';
```

| entity_name | subentity_name | permission_name |
|-------------|----------------|-----------------|
|             |                |                 |

All these errors are not fatal and the procedure will continue to run.

**What is the expected behavior?**
[...]

PR #3359, opened 2023-09-26
#3356 Fix for sp_validatelogins

Montro1981:
#3356 Fix for sp_validatelogins

Rewritten the code so that the `#InvalidLogins` is created a little earlier and we attempt to fill the `#InvalidLogins` in the permissions check, if that fails we'll disable the check.


BrentOzar (approved):

PR #3360, opened 2023-09-26
#3356 Fix for xp_regread

Montro1981:
#3356 Fix for xp_regread

Rewritten the code so that the variables `@powerScheme`, `@cpu_speed_mhz`, `@cpu_speed_ghz` are declared at the start. During the permissions check we'll attempt to set these variables, if that fails we'll disable the check.


BrentOzar (approved):

PR #3361, opened 2023-09-26
#3356 Fix for xp_readerrorlog

Montro1981:
#3356 Fix for xp_readerrorlog

After reviewing the current code I noticed that bluntly disabling the check wouldn't be the preferred way forward. The code was doing checks for `Database Instant File Initialization` in other places as well that might succeed even if we don't have permissions on `xp_readerrorlog`, so chose to integrate a TRY CATCH block in the check itself.

Changed the `@@ROWCOUNT > 0` to a more descriptive IF statement


BrentOzar (approved):

PR #3379, opened 2023-10-11
#3377 Moved the xp_regread calls back to the check and add some messages to the user

### LINKED ISSUE #3377: [sp_Blitz] [Bug] Permissions checks are not working as intended
**Version of the script**
8.17
20231010

Mea culpa, some of the fixes I made for the permission checks are not working.
😢 

* #3360 xp_regread doesn't play nice with a TRY CATCH block, so we just have to swallow those errors, they are not breaking the run, they just annoy me in the messages tab but, oh well, that's what we get for using an undocumented stored procedure.
* #3358 the permission check for the msdb database is to broad
* Forgot the add a check for [master].[dbo].[sp_MSgetalertinfo]

Fix the code to correct the issue I introduced.
* xp_regread, it gets the `@powerScheme` alright after the first error. The cpu speed not so well. Going to ISNULL the speed to UNKW
* add additional permission check for the msdb database
* add permission checkfor [master].[dbo].[sp_MSgetalertinfo]

Will be separate Pull requests per fix.

Same errors different day, thus more fixes.

Montro1981:
Fixes #3377
PR #3360 Didn't work as I had intended, xp_regread doesn't play nice with TRY CATCH, removed the permission check at the start and moved it back to check 211.

Added so informational message to the messages tab to inform the user that the errors can be ignored.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -200,10 +200,6 @@ AS
 			,@SkipModel bit = 0
 			,@SkipTempDB bit = 0
 			,@SkipValidateLogins bit = 0
-			/* Variables for check 211: */
-			,@powerScheme varchar(36)
-			,@cpu_speed_mhz int
-			,@cpu_speed_ghz decimal(18,2);
 
 			DECLARE
 			    @db_perms table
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

PR #3380, opened 2023-10-11
#3377 Expanded msdb permission checks

### LINKED ISSUE #3377: [sp_Blitz] [Bug] Permissions checks are not working as intended

Montro1981:
Fixes #3377
The msdb permission check was too broad and didn't work as intended, one can see the objects in the database but may not query the dbo.sysjobs, dbo.sysalerts, and/or dbo.sysoperators

I split the `@SkipMSDB` into 2 separate variables
```
@SkipMSDB_objs
@SkipMSDB_jobs
```

And expanded the check to a general objects check, if that one passes we can view the user tables.
The `@SkipMSDB_jobs` is set to 1 if we can't query `msdb.dbo.sysjobs`
If `@SkipMSDB_jobs = 1` then we'll skip check related to jobs

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -198,8 +198,6 @@ AS
 			,@SkipMaster bit = 0
 			,@SkipMSDB_objs bit = 0
             ,@SkipMSDB_jobs bit = 0
-            ,@SkipMSDB_alerts bit = 0
-            ,@SkipMSDB_operators bit = 0
 			,@SkipModel bit = 0
 			,@SkipTempDB bit = 0
 			,@SkipValidateLogins bit = 0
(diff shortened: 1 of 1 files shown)


Montro1981:
Simplified the msdb jobs check to just 1 variable instead of 3 which was a bit overkill.

BrentOzar (approved):

PR #3383, opened 2023-10-14
Add missing drop if exists statements for temp tables

chadbaldwin:
Added drop if exists statements for two temp tables in order to make the contents of sp_blitzindex easier to run ad-hoc.


BrentOzar (approved):
Thanks for the pull request! Makes sense, merging into dev, will be in the next release with credit to you in the release notes.

PR #3397, opened 2023-11-20
Fix date math and improve MI compatibility

### LINKED ISSUE #3385: sp_BlitzLock last_execution_time error
**Version of the script**
Version = 8.17
VersionDate = 20231010

sp_BlitzLock returns an error when there is a last_execution_time value in sys.dm_exec_query_stats that returns '1900-01-01 00:00:00.000'

```
Msg 535, Level 16, State 0, Procedure sp_BlitzLock, Line 3758 [Batch Start Line 0]
The datediff function resulted in an overflow. The number of dateparts separating two date/time instances is too large. Try to use datediff with a less precise datepart.
```

Check sys.dm_exec_query_stats for any queries that have a last_execution_time of '1900-01-01', do a DATEDIFF using the SECOND datepart, and compare to GETDATE()

Repo code:

```
DECLARE @creationTime DATETIME = GETDATE()
DECLARE @lastExecutionTime DATETIME = '1900-01-01 00:00:00.000'

SELECT DATEDIFF(SECOND, @creationTime, @lastExecutionTime)
```

Imagine it should either exclude those rows or treat them as NULL, unsure what the expected behavior should be for them.

**Which versions of SQL Server and which OS are affected by this issue?**

SQL Server version - Standard 2019 (15.0.4316.3)
OS version - Windows Server 2019 Standard

**Did this work in previous versions of our procedures?**
Looks like [this section of sp_BlitzLock was added](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/commit/ff8a0f4869a3f040ea3741d134b877288a9b5ad7) by @erikdarlingdata on Jun 28.

### LINKED ISSUE #3392: sp_BlitzLock in Azure SQLMI not working due to Azure changes
**Version of the script** 
Look in the stored procedure, and it'll have a version date & number near the top. Put that in here. If it's not the current version (dated in the last month), then upgrade to the current version and test that before reporting a bug - we fix a lot of stuff in each new build. We'll flat out close bug reports for older builds.
--Version = '8.17', VersionDate = '20231010';

-- Script is erroring out and giving the below error message
image

-- Opened a New Query window in my Azure SQLMI. Executed "sp_BlitzLock" and "sp_BlitzLock @DatabaseName = [NameOfDB]". Below is the error that came from running both scripts

--This script checks for and analyzes deadlocks from the system health session or a custom extended event path

--Azure Managed Instance. OS: Server 2022 version 10.0. I've also tried to run the sp_BlitzLock in OS Server 2019 Azure SQLMI with no luck.  I've ran the sp_BlitzLock in a SQLExpress server and that worked just fine.

erikdarlingdata:
This pull request looks a little funny, because when I re-synced my fork, the available plans query was all beat up for some reason, and I had to paste in the version from the current master branch.

Anyway, the two actual changes in this PR:
 * Closes #3385
 * Closes #3392

Since I can't repro #3385, I'm making the smallest change possible that fixes the issue with mocked-up test data. It will need some community testing to validate.

For #3392 I added a local variable to detect MI, and then set the TargetSessionType to ring_buffer if it's NULL, and we're looking at the system_health extended event. It might be a good idea to have this happen for any session on MI, but I'm going to keep it limited at first.

This also addresses #3398 now too. It was a small change to add in.


BrentOzar (approved):
Looks good, although I have to confess I only did minimal testing and a quick scan of the code. Thanks for the pull request. Merging into the dev branch, will be in the next release with credit to you in the release notes.

erikdarlingdata:
@BrentOzar I've been testing using the dev branch with clients as I've made the changes, and all's well.

image

Issue #3401, opened 2023-12-04
sp_Blitz: Improve Markdown output alias and spacing


MikeScalise:
**Is your feature request related to a problem? Please describe.**
When I run sp_Blitz with the OutputType = 'markdown', I notice a few things:

1) The alias for the column returned is something like "XML_F52E2B61-10A2-18c2-B105-00805F499121", which isn't ideal.

2) When you click the hyperlink, the markdown text starts on line 2 instead of 1.

3) There is a line spacing inconsistency between some of the headings and their subheadings. For example, there's one line in between the finding "- Backups Not Performed Recently - Last backed up: <date>" and the indented list of databases. The same is true for "Last good DBCC CHECKDB over 2 weeks old - Last successful CHECKDB:  <date>." and the subheading list of databases underneath it. However, for the findings "- System Database on C Drive" and "- Tables in the Master Database" there are no lines between the heading and the subheading list of databases underneath them.

4) When pasting the markdown text into some viewers like "MarkdownViewer++" for Notepad++ (which I would argue is one of the more popular options for Windows users), the text for each item in the bulleted lists appears just underneath its associated bullet, across all of the lists--and while it's completely legible, the misalignment looks odd and incorrect.

**Describe the solution you'd like**
For 1), I would like the column to be called "Markdown"
For 2), I would like the markdown text to start on the first line.
For 3), I would like the spacing to be consistent between a given heading and its subheadings.
For 4), I would like the text for each bullet in a list to appear next to its associated bullet.

**Describe alternatives you've considered**
Since there's nothing functionally wrong with the current state of the markdown output, I haven't considered alternatives.

**Are you ready to build the code for the feature?**
Yes. I've written code to handle the four items I described, and I'm ready to commit them / create a pull request if I get the green light to move forward.

BrentOzar:
Looks good! Thanks for the pull request. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3403, opened 2023-12-04
#3401 sp_Blitz: Improve Markdown output alias and spacing

MikeScalise:
When users specify the parameter OutputType = 'markdown', they'll notice the following updates:

1) The column alias is now "Markdown". Achieved by using a slightly different method of retrieving valid XML.

2) When they click on the hyperlink, the markdown text starts on line 1 instead of line 2. Achieved using the STUFF function.

3) Consistent line spacing between a given heading and its subheadings. Achieved by adding a crlf to one of the CASE expression conditions.

4) Align item text in bulleted lists with its associated bullet (for some markdown viewers, including the MarkdownViewer++ plugin for Notepad++). Achieved by adding a single style tag at the end of the markdown text. Tested this on other markdown viewers and it didn't adversely affect the results.


BrentOzar (approved):

Issue #3399, opened 2023-11-22
sp_ineachdb [Enhancement] run only for writable copy of an AG database


DougTaft:
I want to run a script for all of the database that are writable, excluding databases that are not the writeable copy in an AG.

To achieve this I added a param, @is_ag_writeable_copy, bit with default value of 0.  In the existing section, from Andy Mallon, for testing if an AG is present, I added an IF block to remove database that are not the writable copies in an AG if the above param is set to 1.

If this makes sense I am happy to submit the change.

  IF @SQLVersion >= 11 AND 3 = (SELECT COUNT(*) FROM sys.all_objects WHERE name IN('availability_replicas','dm_hadr_availability_group_states','dm_hadr_database_replica_states'))
  BEGIN
    DELETE dbs FROM #ineachdb AS dbs
    WHERE EXISTS
          (
            SELECT 1 FROM sys.dm_hadr_database_replica_states AS drs 
              INNER JOIN sys.availability_replicas AS ar
              ON ar.replica_id = drs.replica_id
              INNER JOIN sys.dm_hadr_availability_group_states ags 
              ON ags.group_id = ar.group_id
              WHERE drs.database_id = dbs.id
              AND ar.secondary_role_allow_connections = 0
              AND ags.primary_replica <> @ServerName
          );
	/* Remove databases which are not the writeable copies in an AG. */
	IF @is_ag_writeable_copy = 1
	BEGIN
		DELETE dbs FROM #ineachdb AS dbs
		WHERE EXISTS
			  (
				SELECT 1 FROM sys.dm_hadr_database_replica_states AS drs 
				  INNER JOIN sys.availability_replicas AS ar
				  ON ar.replica_id = drs.replica_id
				  INNER JOIN sys.dm_hadr_availability_group_states AS ags 
				  ON ags.group_id = ar.group_id
				  WHERE drs.database_id = dbs.id
				  AND drs.is_primary_replica <> 1
				  AND ags.primary_replica <> @ServerName
			  );
	END
  END

BrentOzar:
This makes total sense. We'd love a pull request for it. Thanks!

BrentOzar:
Thanks for the pull request. Looks good, merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3408, opened 2023-12-14
#3399 modified sp_ineachdb to generate scripts only for writable databases in an AG

DougTaft:
sp_ineachdb run only for writable copy of an AG database To achieve this I added a param, @is_ag_writeable_copy, bit with default value of 0. In the existing section, from Andy Mallon, for testing if an AG is present, I added an IF block to remove database that are not the writable copies in an AG if the above param is set to 1.
modified:   sp_ineachdb.sql


BrentOzar (approved):

PR #3426, opened 2024-01-18
Skip xp_regread if we don't have sa perms

### LINKED ISSUE #3425: sp_Blitz - Skip xp_regread when @sa is set to 0
When I run sp_Blitz without adequate permissions for xp_regread, it throws an error. I realize work was done on this in #3377 to fix faulty permissions checks, and add explanatory error messages to the output, but realistically you're not going to be able to call this without sa permissions.

Just skip the xp_regead calls when the `@sa` variable is set to 0.

Being annoyed with my own incompetence around security and permissions checking.

It is DONE baby!

erikdarlingdata:
closes #3425

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -659,7 +659,7 @@ AS
 		SELECT
 		    v.*
 		FROM (VALUES(NULL, 211, NULL)) AS v (DatabaseName, CheckID, ServerName) /*xp_regread*/
-		WHERE @SkipXPRegRead = 1;
+		WHERE @sa = 0;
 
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

Montro1981:
@erikdarlingdata the issue I had with this is that check 211 ran mostly fine on my limited permission box, and xp_regread did work for the most part, that was the reason why I chose the solution that I build before ;-)

erikdarlingdata:
Yeah, I still couldn’t run it without error in the limited permissions environment I was in, which is why I did this. If you want to do a more granular permissions check, go for it.

PR #3431, opened 2024-01-31
Add FileNamePrefix parameter to sp_DatabaseRestore

LukaszBiegus:
Add `@FileNamePrefix` parameter to `sp_DatabaseRestore` - a prefix to add to the names of all restored files. Useful when you need to restore different backups of the same database into the same directory.


BrentOzar (approved):
Thanks for the pull request! I haven't tested this because I don't use sp_DatabaseRestore myself, but conceptually it looks good. Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3442, opened 2024-02-18
Update sp_DatabaseRestore.sql

gdoddsy:
Add VerifyRestoreWithStoredProcedure option to sp_databaseRestore.

Still needs work for the unit testing, but just want to check if the concept is valid first.


gdoddsy:
For issue #3442

BrentOzar (approved):
When I look at what the code does, it isn't necessarily testing the database - it's really just running a stored proc after the restore finishes. 

This functionality could be good for other stuff too, like obscuring data or changing permissions. I switched the parameter name from @VerifyRestoreWithStoredProcedure to @RunStoredProcAfterRestore.

PR #3443, opened 2024-02-18
Update sp_DatabaseRestore.sql

gdoddsy:
@RunStoredProcAfterRestore should not be inside the @TestRestore check


gdoddsy:
@BrentOzar - suspect there's a bug in the change you made. If the plan is to run this for any situation, then the stored proc should run whether `@TestRestore` is true or not.

BrentOzar:
Sorry about that! That's weird, I just did a search/replace on the parameter name and edited a description.

PR #3453, opened 2024-03-06
Update Install-Core-Blitz-No-Query-Store.sql

stanom:
modified code to be compatible with SQL Server version <= 2014


asos-martinsmith:
I think this fix should be done in the source proc (sp_BlitzIndex.sql) as well as all the places this proc definition is replicated. 

Rather than just fixing it in one copy of it

BrentOzar (commented):
Thanks for the pull request, but make sure to read the readme and the contributing guide. Changes to the scripts need to be made in their individual files rather than the installation scripts - the installation scripts are generated automatically.

PR #3455, opened 2024-03-09
Removed skipping of model and msdb in check for QS-enabled DBs

ReeceGoding:
Although it's sometimes considered a bad idea to turn on Query Store for these databases, that doesn't justify telling people that they haven't done so.

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -443,7 +443,7 @@ IF  (	SELECT COUNT(*)
 		WHERE d.is_query_store_on = 1
 		AND d.user_access_desc='MULTI_USER'
 		AND d.state_desc = 'ONLINE'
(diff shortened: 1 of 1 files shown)


BrentOzar:
Hmm, can you elaborate on a time where you've found it helpful to have Query Store enabled for these databases? I'm pretty suspicious.

ReeceGoding:
I've yet to find it useful in `msdb`, but it can be handy in `model` because [other databases inherit their Query Store settings from model](https://dba.stackexchange.com/questions/299301/how-to-check-query-store-settings-on-model-database). That said, there's obviously no useful data to look at in `model`.

But all of that is besides the point. What matters is that the current error message lies to the user. We shouldn't tell them that they don't have something enabled when they do.

BrentOzar:
That is a really fantastic point about model!

BrentOzar (approved):

PR #3458, opened 2024-03-15
Gave every warning check its own RAISERROR. Also indentation updates.

ReeceGoding:
One of the warning checks was missing a `RAISERROR`. This fixes that.

Also, the warning checks has some strange indentation. This changes that.

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -3022,6 +3022,9 @@ OPTION (RECOMPILE);
 /*
 This looks for parallel plans
 */
+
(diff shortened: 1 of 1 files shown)


BrentOzar:
Reece - can you open issues explaining what problem you're solving before submitting pull requests? That way we can discuss 'em and make sure we're on the right track.

In this case, I'm totally fine with adding errors, but let's not reformat code. Reformatting creates a big change request that human beings have to review - in this case, me, heh - and I need to focus on bang for the buck. Thanks for understanding.

I'm going to close this PR, but definitely feel free to reopen it with a change that only adds the error messages.

BrentOzar:

PR #3464, opened 2024-03-29
Closes #3461

### LINKED ISSUE #3461: sp_BlitzIndex - Mode 2 should fail more loudly when BringThePain is needed
If you run sp_BlitzIndex in Mode 2, and BringThePain is required, there's no canary row returned telling you to use it like in other modes. There's a message in the results pane, but it might be hard for someone familiar with the script not to catch.

It looks like the easiest thing would be to dump the select into a temp table, check the row count, and then either select from that with the big order by/case thing, or return a row informing our intrepid index tuner to use BringThePain and re-run.

Puttin' puddin' in my wooden shoe.

Yes, it's a small change.

erikdarlingdata:
Closes #3461

This dumps the mode 2 results into a temp table, and then either returns the contents or some situational details about the execution and instructions to use BringThePain when required.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -6036,52 +6036,77 @@ BEGIN
 						WHEN i.index_definition = '[HEAP]' THEN N''
 					    ELSE N'--' + ict.create_tsql END AS [Create TSQL], 
 					1 AS [Display Order]
+            INTO #Mode2Temp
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks for the pull request! Totally makes sense, merging, will be in the next release with credit to Cthulu in the release notes.

PR #3465, opened 2024-03-29
Update sp_BlitzIndex.sql

### LINKED ISSUE #3462: sp_BlitzIndex - Skip missing index details for Modes 1 and 2
Gathering missing index details can be time consuming on some servers (another issue about that to come), but it doesn't make sense to run this part at all in modes that don't do any analysis or return those results. 

Skip missing index details if using Mode 1 or 2.

Waiting. Endless... waiting.

Sure thing skipperoo.

erikdarlingdata:
Closes #3462

Skips doing some checks when using modes that don't report on the results.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1776,7 +1776,8 @@ BEGIN TRY
 		END; --End Check For @SkipPartitions = 0
 
 
-
+		IF @Mode NOT IN(1, 2)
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
I'm glad you did this because I've actually wondered in the past, "Are we gathering the data for every mode, every time?" and not gone into the code to look. Doh!

Issue #3449, opened 2024-02-28
sp_BlitzQueryStore prints 'Gathering highest log byte use plans' twice


ReeceGoding:
**Version of the script**
`@Version = '8.19', @VersionDate = '20240222';`

**What is the current behavior?**
'Gathering highest log byte use plans' prints twice. You can see this on [line 2186](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_BlitzQueryStore.sql#L2186C13-L2186C50) and [line 2234](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_BlitzQueryStore.sql#L2234).

**If the current behavior is a bug, please provide the steps to reproduce.**
Run the procedure with a `@DatabaseName` that supports the 2017 columns.

**What is the expected behavior?**
It should either only print once or it should print two different messages.

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
This isn't a SQL version or OS issue. It's a copy and paste error.

**Other Details**
I'm quite willing to fix this myself, but there's a question around what the fix should be. There are two obvious fixes:

1. Just delete one of the `RAISERROR` lines
2. Correct the `RAISERROR` that claims that it's checking for the "highest" when it's actually looking for the average.

Option two is both my preference and deeper than it looks. For consistency, if this pattern check has two distinct messages (one for "average" and one for "highest"), then we would need to include similar messages for all of the patterns that also check for "average" and "highest". By my count, that's seven more `RAISERROR` calls. We do seventeen pattern checks, but only print ten relevant messages. The seven categories with missing messages are: Duration, CPU, logical reads, physical reads, writes, memory use, and tempdb use. While we're at it, advertising in the documentation that we're checking seventeen patterns seems like a good idea. The docs currently read as if only nine are being done.

Again, I'm willing to do this myself. I just want some input on if we want option one or two.

ReeceGoding:
Related issue: The printed message "Gathering working plans" prints long after `#working_plans` is populated. The table that is actually populated at that point is called `working_plan_text`.

BrentOzar:
Fixed the stuff in the first issue note - if there are separate issues, let's get those documented individually rather than merging stuff together. Thanks for understanding.

PR #3469, opened 2024-03-30
Added missing RAISERROR to the check for parallel plans

ReeceGoding:
As requested in #3458 , reopened this as a PR that only includes the missing `RAISERROR`. Empty lines aside, this is just a one-line change so there's no need to open an issue or ask for resources.

The check for parallel plans was the only _warning_ check that was missing a `RAISERROR`, so I have added it in for consistency. There are _pattern_ checks that have oddities around what they `RAISERROR`, but that is [a separate issues](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3449) that requires some discussion before writing any code.

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -3022,6 +3022,9 @@ OPTION (RECOMPILE);
 /*
 This looks for parallel plans
 */
+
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

PR #3471, opened 2024-03-30
sp_BlitzQueryStore.sql: Make commenting style consistent

### LINKED ISSUE #3470: sp_BlitzQueryStore: Let's make the commenting style consistent
I'm always frustrated when I want to add a new feature to `sp_BlitzQueryStore`. The commenting style is extremely inconsistent, so I do not know how to comment my code! I find similar issues when I notice that a temp table has no comments (examples [here](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_BlitzQueryStore.sql#L1048) and [here](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_BlitzQueryStore.sql#L915)) and feel like fixing that myself.

The most common form is currently `/*Foo*/` so I'd say to apply that everywhere other than the multi-line block comments.

Aside from this style, we currently also have `--Foo`, `-- Foo`, `/*foo*/`, `/* Foo */`, and
```sql
/*
I am a single sentence on a single line. This comment block has no other lines of text.
*/
```
but I see no reason to pick those over `/*Foo*/`, which seems to be the most common.

Note that I am not suggesting building any kind of style guide. I suspect that once this change is made, anyone else who wants to write a comment in `sp_BlitzQueryStore` will be able to infer the commenting style from the code surrounding their change.

[Brent seems to prefer](https://www.brentozar.com/archive/2021/04/never-ever-ever-start-t-sql-comments-with-two-dashes/) `/* Foo */` over `/*Foo*/`, so we could go with that.

I've already done it. I just need to split it out from a PR where I got carried away and changed too much.

ReeceGoding:
Closes #3470. Also includes a very small number of typo fixes in the comments.

I could have waited until the issue had some discussion on it, but I already had the code ready from an old PR so there was no reason to not just throw this PR in.

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -10,11 +10,11 @@ GO
 
 DECLARE @msg NVARCHAR(MAX) = N'';
 
-	-- Must be a compatible, on-prem version of SQL (2016+)
+	/*Must be a compatible, on-prem version of SQL (2016+) or...*/
 IF  (	(SELECT CONVERT(NVARCHAR(128), SERVERPROPERTY ('EDITION'))) <> 'SQL Azure' 
 	AND (SELECT PARSENAME(CONVERT(NVARCHAR(128), SERVERPROPERTY ('PRODUCTVERSION')), 4)) < 13 
 	)
-	-- or Azure Database (not Azure Data Warehouse), running at database compat level 130+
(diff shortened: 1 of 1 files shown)


ReeceGoding:
@BrentOzar I've resolved the merge conflict.

PR #3475, opened 2024-03-30
sp_BlitzQueryStore.sql: DRY the SERVERPROPERTY checks

### LINKED ISSUE #3474: sp_BlitzQueryStore: Make the various calls to SERVERPROPERTY less repetitive and capture the current database's compatibility level in a variable
We use the following many times in `sp_BlitzQueryStore`, but do not store them in variables:
```sql
CONVERT(NVARCHAR(128), SERVERPROPERTY ('EDITION')),
PARSENAME(CONVERT(NVARCHAR(128), SERVERPROPERTY ('PRODUCTVERSION')), 4),
CONVERT(INT, SERVERPROPERTY ('ENGINEEDITION'));
```
How about we assign them to well-named variables and make the code a little less repetitive and a little more readable? While we're at it, capturing the current database's compatibility level would make one of the check for Azure nicer.

I'm thinking something like
```sql
DECLARE
	@edition NVARCHAR(128), /*Used to check if we are on Azure or not*/
	@major_version NVARCHAR(128), /*Used to see what checks we can do*/
	@engine_edition INT, /*Used to check which Azure we are on*/
	@current_database_compatibility_level INT; /*Used to check which Azure we are on*/

/*Assigns various repeated SERVERPROPERTY calls to variables*/
SELECT
	@edition = CONVERT(NVARCHAR(128), SERVERPROPERTY ('EDITION')),
	@major_version = PARSENAME(CONVERT(NVARCHAR(128), SERVERPROPERTY ('PRODUCTVERSION')), 4),
	@engine_edition = CONVERT(INT, SERVERPROPERTY ('ENGINEEDITION'));

/*Grabs current database's compatibility level*/
SELECT @current_database_compatibility_level = [compatibility_level]
FROM sys.databases
WHERE [name] = DB_NAME()
OPTION (RECOMPILE);
```
followed by using these variables in the appropriate places. The only trap is that you cannot use them at the very start of the code for `sp_BlitzQueryStore` because that code is for installing the procedure.

Just leaving it alone might be a good idea. I don't have the resources to test this myself and it's a much more risky change than anything else that I've recently proposed.

 I've already done it. I just need to split it out from a PR where I got carried away and changed too much.

ReeceGoding:
Closes #3474. However, this probably merits some discussion. I only made the PR because I already had the code written from an earlier one. This PR also servers as a good demonstration of what the issue is seeking to correct.

I don't have my own Azure box so this isn't easy to test by myself. Another snag is in the datatypes. Some of these were originally `NVARCHAR` and others were `VARCHAR`. I am not sure if there is a good reason behind this, so I have gone with my [the docs](https://learn.microsoft.com/en-us/sql/t-sql/functions/serverproperty-transact-sql?view=sql-server-ver16). `ENGINEEDITION` is an exception to this and has been stored as an `INT`.

### DIFF
--- a/sp_BlitzQueryStore.sql
+++ b/sp_BlitzQueryStore.sql
@@ -84,7 +84,11 @@ DECLARE /*Variables for the variable Gods*/
 		@is_azure_db BIT = 0, --Are we using Azure? I'm not. You might be. That's cool.
 		@compatibility_level TINYINT = 0, --Some functionality (T-SQL) isn't available in lower compat levels. We can use this to weed out those issues as we go.
 		@log_size_mb DECIMAL(38,2) = 0,
-		@avg_tempdb_data_file DECIMAL(38,2) = 0;
+		@avg_tempdb_data_file DECIMAL(38,2) = 0,
+		@edition NVARCHAR(128), /*Used to check if we are on Azure or not*/
+		@major_version NVARCHAR(128), /*Used to see what checks we can do*/
+		@engine_edition INT, /*Used to check which Azure we are on*/
+		@current_database_compatibility_level INT; /*Used to check which Azure we are on*/
 
 /*Grabs CTFP setting*/
 SELECT  @ctp = NULLIF(CAST(value AS INT), 0)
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):

ReeceGoding:
@BrentOzar I've resolved the merge conflict.

BrentOzar:
If this isn't fixing a real problem or adding functionality, let's hold off.

I know it sounds like I'm being pedantic here, but I'm literally a volunteer on the other side of the screen, trying to keep up with my day job while helping the community.

I'm worried about instances like this: https://doublepulsar.com/inside-the-failed-attempt-to-backdoor-ssh-globally-that-got-caught-by-chance-bbfe628fafdd

Where someone purposely (or accidentally) checks in something that's accidentally bad, or malicious. Because of that, I try to only focus on code changes that fix actual bugs, or add significant functionality. Going forward, let's focus on that. Thanks for understanding, and I really do wish I had more time to test things like this, but...

PR #3511, opened 2024-05-06
NumDatabases includes those already marked as unreachable

garyhuntddn:
There is code to identify databases which can't be analysed that results in secondary_role_allow_connections_desc = 'NO', but these aren't excluded from the NumDatabases count


BrentOzar (approved):

Issue #3527, opened 2024-05-29
sp_Blitz: Add more Query Store best practice checks.


ReeceGoding:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when `sp_Blitz` tells me that I have trace flag 3226 on and gives me a lovely description of it, but gives no lovely description for either of the trace flags that are considered best practice for Query Store (7745 and 7752). This got me thinking that we should do some more general checks for Query Store best practice.

**Describe the solution you'd like**
`sp_Blitz` should warn in all of the following scenarios:

1. Query Store enabled. but TF 7745 is off.
2. Query Store enabled on a pre-2019 box, but TF 7752 is off.
3. Query Store enabled on a 2019+ box, but TF 7752 is on. I might be mistaken, but I think that we can tell if an enabled-by-default trace flag has been deliberately turned on? If so, that means that a 2019+ box with TF 7752 enabled is worrying. It probably means that somebody did an upgrade and forgot to clear it from the list of startup parameters. 
4. TF 7745 or TF 7752 enabled on a box that doesn't support Query Store.
5. Query Store enabled, but Wait Stats Capture Mode disabled despite the box supporting it.
6. Query Capture Mode in an state other than AUTO. ALL is usually a stupid idea and, much like how we warn for RCSI being enabled, we should probably warn if the mode is CUSTOM because that probably means that somebody is doing some interesting non-default stuff.
7. Query Store not in ReadWrite state.
8. Maybe warn if Query Store is eating up a silly amount of RAM? I'm not sure what "silly" means. 1 GB?

In addition, the generic warning that TF 7745 or TF 7752 are on should be specific and say what they actually do.  Maybe [link to Erin](https://www.sqlskills.com/blogs/erin/query-store-trace-flags/)?

**Describe alternatives you've considered**
Leaving it alone comes to mind. [dbatools has beaten us to it](https://docs.dbatools.io/Test-DbaDbQueryStore).

**Are you ready to build the code for the feature?**
I sincerely can't decide.

BrentOzar:
Everything you're describing here makes sense. If you go for it, also update the Documentation/sp_Blitz_Checks_by_Priority.md file to add your new check IDs there.

For URLs for warnings, you definitely don't wanna link to BrentOzar.com stuff because we just don't have any kind of good guidance on there. Feel free to link to whatever URL makes the most sense for each one of the warnings.

ReeceGoding:
[...]

PR #3533, opened 2024-06-07
sp_Blitz: Add check for very high Query Store memory usage

ReeceGoding:
Closes item 8 in #3527. Other items in that issue remain open.

I really didn't do anything smart while making this. I just cracked open the docs for [sys.dm_os_memory_clerks](https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-os-memory-clerks-transact-sql?view=sql-server-ver16#types) and searched for references to Query Store.

1 GB was chosen utterly arbitrarily. I have no idea what can reasonably be considered a ridiculously high amount of Query Store memory usage. I've never seen a box crack 0.75 GB.

I've made this a priority 50 issue under the 'Performance' header. If your Query Store is eating a ridiculous amount of memory, then you have something interesting going on and ought to look. Everything with a number higher than 50 under the same header seemed far less important.

No useful URL has come to mind. Online research about Query Store overhead never mentions checking this metric, even though `sys.dm_os_memory_clerks` is the basis for a few of our checks. I could link to a generic guide on how to configure Query Store, but that wouldn't focus on the problem.

This is my first new check for sp_Blitz, so it's quite likely that it's missing something important. It's passed the limited suite of tests that I've tried on my local instance.

### DIFF
--- a/Documentation/sp_Blitz_Checks_by_Priority.md
+++ b/Documentation/sp_Blitz_Checks_by_Priority.md
@@ -6,8 +6,8 @@ Before adding a new check, make sure to add a Github issue for it first, and hav
 
 If you want to change anything about a check - the priority, finding, URL, or ID - open a Github issue first. The relevant scripts have to be updated too.
 
-CURRENT HIGH CHECKID: 261.
-If you want to add a new one, start at 262.
+CURRENT HIGH CHECKID: 262.
+If you want to add a new one, start at 263.
 
 | Priority | FindingsGroup | Finding | URL | CheckID |
 |----------|-----------------------------|---------------------------------------------------------|------------------------------------------------------------------------|----------|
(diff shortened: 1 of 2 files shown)


BrentOzar:
The idea is interesting, but I did a quick check on the data from SQL ConstantCare, and there are indeed tons of boxes with >1GB of memory in Query Store, going as high as 18GB, for boxes that have 256GB+ memory.

I'm not saying 18GB is normal, but <10% of the box's memory doesn't seem like a big consumer. Because of that, in order to set a threshold, we should probably be able to answer these questions:
* What's the problem we're trying to solve?
* What's the evidence we have that the user is experiencing degraded performance because of this?
* What's the specific action we recommend users take to fix it?

I don't have those answers handy, so I don't think this is a good fit for sp_Blitz. If you do run into a box where you're running into specific, provable problems due to high memory consumption by Query Store, then let's have a conversation from there, but until then let's avoid adding general advice that may not actually fix a problem.

PR #3535, opened 2024-06-15
sp_Blitz: Add check for Query Store Wait Stats being turned off

ReeceGoding:
Closes item 5 in #3527  Only items 6 and 7 are now awaiting PRs.

I've added this under the same check number as my other new check, because it's rude to assume that both will get merged.

I've made this a priority 200 issue under the 'Performance' header. This is the consistent with the check for having Query Store completely disabled.

I couldn't find a better URL than the one that I've given. One must surely exist.

This is my first new mutli-database check for `sp_Blitz`, so it's quite likely that it's missing something important. It's passed the limited suite of tests that I've tried on my local instance.


BrentOzar (approved):

PR #3536, opened 2024-06-15
sp_Blitz: Added check for Query Store not being in READ_WRITE state and check for it not being in desired state.

ReeceGoding:
Closes item 7 in https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3527 . Only item 6 is now awaiting a PR. This change includes two new checks. Adding a check for Query Store not being in its desired state was such a natural follow-up from checking for READ_WRITE that I couldn't resist.

I've made this a priority 200 issue under the 'Performance' header. This is the consistent with the check for having Query Store completely disabled. After all, these two issues effectively disable Query Store.

The URL I've added is the best one that I know for this topic. Both checks have the same URL.

I've only been able to test the check for the READ_WRITE state. I've only ever seen Query Store leave its desired state on busy production databases. I'm unsure how to simulate it to test it myself. I'm open to ideas. Maybe just make the settings for Query Store ridiculously low and hit your server very hard?

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -6758,6 +6758,67 @@ IF @ProductVersionMajor >= 10
 									  AND N''?'' NOT IN (''master'', ''model'', ''msdb'', ''tempdb'', ''DWConfiguration'', ''DWDiagnostics'', ''DWQueue'', ''ReportServer'', ''ReportServerTempDB'') OPTION (RECOMPILE)';
 							END;
 
+						IF NOT EXISTS ( SELECT  1
+										FROM    #SkipChecks
+										WHERE   DatabaseName IS NULL AND CheckID = 262 )
+                            AND EXISTS(SELECT * FROM sys.all_objects WHERE name = 'database_query_store_options')
+							BEGIN
+
+								IF @Debug IN (1, 2) RAISERROR('Running CheckId [%d].', 0, 1, 262) WITH NOWAIT;
+
+								EXEC dbo.sp_MSforeachdb 'USE [?];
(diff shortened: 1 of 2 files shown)


sp_Blitz.sql
```diff
@@ -6758,6 +6758,67 @@ IF @ProductVersionMajor >= 10
...
+                                        SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
+			                            INSERT INTO #BlitzResults
+			                            (CheckID,
+			                            DatabaseName,
+			                            Priority,
+			                            FindingsGroup,
+			                            Finding,
+			                            URL,
+			                            Details)
+		                              SELECT TOP 1 262,
```
BrentOzar, line 6779: In another pull request, you're using 262 as the ID for a different check. Can you fix the check numbers in this and the documentation, and then I can accept it? Thanks!

BrentOzar (changes requested):
In another pull request, you're using 262 as the ID for a different check. Can you fix the check numbers in this and the documentation, and then I can accept it? Thanks!

ReeceGoding:
@BrentOzar That should fix it.

ReeceGoding:
>  I'm unsure how to simulate it to test it myself. I'm open to ideas. Maybe just make the settings for Query Store ridiculously low and hit your server very hard?

Tested this. It worked. You don't even have to hit the server hard. Just make the Query Store settings super low.

BrentOzar (approved):
Thanks for the PR! Looks good, and thanks for doing the documentation changes too. Makes sense, merging into the dev branch. I'm doing a July release a little early to get a couple of fixes into the Consultant Toolkit, and this will be in there.

PR #3539, opened 2024-06-21
sp_Blitz: Added checks for disabled but recommended Query Store Trace Flags

ReeceGoding:
It occurred to me that items 1 and 2 from #3527 were never actually done. We did plenty of checks for what Trace Flags were on, but not what were off. I've bolted this on to the checks that we already have for Trace Flags. It seems sensible to have all of the Trace Flag checks in one place. If you don't do it like this, then you have to be careful about the state of the both the `@QueryStoreInUse` variable and the temp table that holds the Trace Flags. This also puts all of the relevant logic in one place and provides a copy-and-paste way for anyone to add any future checks for recommended-but-disabled Trace Flags.

Also removed some redundant logic for checking the enabled Query Store Trace Flags. We don't need to check if your version supports Query Store if we already know that you have it enabled.


BrentOzar (approved):
Thanks for the PR! Looks good, and thanks for doing the documentation changes too. Makes sense, merging into the dev branch. I'm doing a July release a little early to get a couple of fixes into the Consultant Toolkit, and this will be in there.

BrentOzar:
Argh, this one can't be merged due to making too many changes in different PRs to the same area of code. I'm going to close this, but feel free to reopen it with a fixed version that'll merge.

PR #3541, opened 2024-06-22
sp_Blitz: Added check for unusual Query Store query capture mode

ReeceGoding:
Closes the last open item from #3527 .

I've made this a priority 200 issue under the 'Performance' header. This is the consistent with the check for having Query Store completely disabled. After all, the None mode isn't very different from that.

My claim that the All query capture mode can turn Query Store off is not in the URL provided. The documentation for that is actually [here](https://learn.microsoft.com/en-us/sql/relational-databases/performance/best-practice-with-the-query-store?view=sql-server-ver16#Verify).


BrentOzar:
Argh, this one can't be merged due to making too many changes in different PRs to the same area of code. I'm going to close this, but feel free to reopen it with a fixed version that'll merge.

PR #3563, opened 2024-09-01
sp_Blitz.sql: Updated #IgnorableWaits to match the 2021 version of Paul Randal's script

### LINKED ISSUE #3562: Add the SQL Server 2019+ ignorable waits to sp_Blitz's #IgnorableWaits.
I'm always frustrated when `sp_Blitz` tells me that I need to worry about my "PWAIT_EXTENSIBILITY_CLEANUP_TASK" waits, but then links to Paul Randal's website that [tells me that I don't](https://www.sqlskills.com/help/waits/pwait_extensibility_cleanup_task/). I suspect that we've not kept `#IgnorableWaits` up to date. [I'm not the first to say this](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2930).

dbatools had the same problem. Before I updated it, their list was last updated in 2018. I don't know when we last updated `#IgnorableWaits` (I can see [some updates from two years ago](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/1798)?), but my plan is to copy the same updates as [what I gave them](https://github.com/dataplat/dbatools/pull/9249/files).

I could submit one ticket per wait and we could review each, but I think we trust Paul's judgement here.

I've also considered doing the same work for `sp_BlitzFirst`, but I'm unsure if it has the same problems. I've yet to notice it complaining about waits that it shouldn't, but I also haven't found any code that prevents it from doing that.

Yes. It shouldn't take long.

ReeceGoding:
Closes #3562 

I expect that adding CXCONSUMER may be controversial. Everything else seems clear-cut.


sp_Blitz.sql
```diff
@@ -1108,6 +1108,7 @@ AS
 		INSERT INTO #IgnorableWaits VALUES ('CLR_AUTO_EVENT');
 		INSERT INTO #IgnorableWaits VALUES ('CLR_MANUAL_EVENT');
 		INSERT INTO #IgnorableWaits VALUES ('CLR_SEMAPHORE');
+		INSERT INTO #IgnorableWaits VALUES ('CXCONSUMER');
```
BrentOzar, line 1111: Let's take this one back out - we show why it's not ignorable in the Mastering Server Tuning class. Thanks!

BrentOzar (changes requested):
Let's take CXCONSUMER out - I explain why in the Mastering Server Tuning class. Thanks!

BrentOzar (approved):
Thanks! I appreciate it!

PR #3575, opened 2024-10-01
sp_Blitz: Fix checks 191 and 73

Tisit:
### Check 73 - No Failsafe Operator Configured

This check currently works only when user has EXECUTE on [master].[dbo].[sp_MSgetalertinfo] and does  NOT have sa. Changed to check for permissions on sp_MSgetalertinfo in case the user isn't sa

### Check 191 - Mismatch between the number of TempDB files in sys.master_files versus tempdb.sys.database_files

User not having sa may get this incorrectly since sys.master_files may return 0 rows for tempdb, while tempdb.sys.database_files returns the right number of rows. Reading through [this article](https://learn.microsoft.com/pl-pl/archive/blogs/sql_pfe_blog/tempdb-misconfiguration-when-sql-server-fails-to-create-a-secondary-data-file) it seems normally sys.master_files = tempdb.sys.database_files. When there are problems sys.master_files > tempdb.sys.database_files. So state where sys.master_files < tempdb.sys.database_files is invalid because we don't see all the rows. Based on this I changed comparision between those tables from <> to >. Kinda stupid, but I feel how SQL Server works here is stupid too.

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -334,18 +334,15 @@ AS
 			    END CATCH;
 			END; /*Need execute on sp_validatelogins*/
             
-            IF ISNULL(@SkipGetAlertInfo, 0) != 1 /*If @SkipGetAlertInfo hasn't been set to 1 by the caller*/
-			BEGIN
-			    BEGIN TRY
-			        /* Try to fill the table for check 73 */
-					INSERT INTO #AlertInfo
-                    EXEC [master].[dbo].[sp_MSgetalertinfo] @includeaddresses = 0;
-			
-			    	SET @SkipGetAlertInfo = 0; /*We can execute sp_MSgetalertinfo*/
-			    END TRY
-			    BEGIN CATCH
-			    	SET @SkipGetAlertInfo = 1; /*We have don't have execute rights or sp_MSgetalertinfo throws an error so skip it*/
-			    END CATCH;
+			IF NOT EXISTS
+            (
+                SELECT
+                    1/0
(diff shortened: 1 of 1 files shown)


sp_Blitz.sql
```diff
@@ -334,18 +334,15 @@ AS
...
-			        /* Try to fill the table for check 73 */
-					INSERT INTO #AlertInfo
-                    EXEC [master].[dbo].[sp_MSgetalertinfo] @includeaddresses = 0;
-			
-			    	SET @SkipGetAlertInfo = 0; /*We can execute sp_MSgetalertinfo*/
-			    END TRY
-			    BEGIN CATCH
-			    	SET @SkipGetAlertInfo = 1; /*We have don't have execute rights or sp_MSgetalertinfo throws an error so skip it*/
-			    END CATCH;
+			IF NOT EXISTS
```
BrentOzar, line 337: Why did we remove @SkipGetAlertInfo? That's a parameter that users can set to skip getting alerts.

BrentOzar:
Thanks for the pull request, but a couple of notes:

* Please submit an issue before doing work. That way, we can talk through things and make sure they make sense. In this example, I don't see the tempdb file check as a bug - it's working as designed. That post about tempdb files that you linked to is true, but it's not the ONLY thing that causes problems when those two system tables don't agree. If you want to add a filter for sys.master_files <> 0 to avoid false alarms, though, that'd be fine.
* Please only submit one pull request per issue - don't try to fix multiple things in the same pull request.

I'm going to close this, but you can definitely submit issues & pull requests going forward. Thanks for understanding, and sorry about the delay!

PR #3588, opened 2024-10-24
INT data type prevents reporting indexes < 1 MB

stone-de-croze:
Updated data type of reserved MB fields in [#dm_db_partition_stats_etc] to match similar fields in other temp tables.
BlitzIndex for a single table will now report "size" for indexes less than 1 MB.


stone-de-croze:
I noticed this rather trivial issue when demonstrating BlitzIndex using a table with few rows.

BrentOzar (approved):

PR #3590, opened 2024-10-25
Update sp_BlitzIndex.sql

VincenzoMarchese:
The select from the #BlitzIndexResults table is present 2 times in the same block


BrentOzar (approved):

PR #3619, opened 2025-04-05
sp_BlitzIndex: Added reports of resumable index operations (closes #3609 )

### LINKED ISSUE #3609: sp_BlitzIndex: add warning for paused resumable index operations
[Unlike `sp_Blitz`](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/pull/2011), there is nothing in `sp_BlitzIndex` that reads `sys.index_resumable_operations`. This can cause problems, because a hidden resumable operation could change the results of any analysis that I'm conducting with `sp_BlitzIndex`.

Something almost identical to the `sp_Blitz` output should be added to the mode 4 output of `sp_BlitzIndex`. A notice should also be added to the output that comes from passing in a table by name. I am unsure if said notice should be its own new table (hidden if empty) or somehow bolted on to the table that currently tells us about missing indexes.

I should look at what we currently do for online index operations. I'm not sure what DMV they're in, but there must be a good reason for whatever we're currently doing with them (even if what we're doing is nothing at all).

Sure. I should have a free weekend soon.

ReeceGoding:
Added reports of resumable index operations to the mode 0 and mode 4 outputs, as well as the mode where you name one table. The priority set here is very high, but that seems correct. They can prevent DDL!

[Brent warned](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3609#issuecomment-2693163590) that I should check if the `sys.index_resumable_operations` honours `NOLOCK`, but as far as I can tell it's impossible to put a resumable operation in a transaction.

My production experience of resumable operations is very limited, so the warnings that I have given about them relied on either trusting the documentation or doing what I could easily verify. If there are much nastier things to warn of, then somebody should change my strings.

I think that this is my first contribution to `sp_BlitzIndex`. It is certainly my first non-trivial one. All of my testing says that I've done the work correctly. However, you shouldn't trust a first timer. 

Not included: Any reference to the 2022 feature that auto-kills paused indexes. It is in [the ALTER DATABASE SCOPED CONFIGURATION docs](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-database-scoped-configuration-transact-sql?view=sql-server-ver16#paused_resumable_index_abort_duration_minutes). I will make an issue for it.

### DIFF
--- a/Documentation/sp_BlitzIndex_Checks_by_Priority.md
+++ b/Documentation/sp_BlitzIndex_Checks_by_Priority.md
@@ -6,13 +6,15 @@ Before adding a new check, make sure to add a Github issue for it first, and hav
 
 If you want to change anything about a check - the priority, finding, URL, or ID - open a Github issue first. The relevant scripts have to be updated too.
 
-CURRENT HIGH CHECKID: 121
-If you want to add a new check, start at 122.
+CURRENT HIGH CHECKID: 123
+If you want to add a new check, start at 124.
 
 | Priority | FindingsGroup           | Finding                                                         | URL                                             | CheckID |
 | -------- | ----------------------- | --------------------------------------------------------------- | ----------------------------------------------- | ------- |
 | 10       | Over-Indexing           | Many NC Indexes on a Single Table                               | https://www.brentozar.com/go/IndexHoarder       | 20      |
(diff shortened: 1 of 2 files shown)


BrentOzar:
Looks good! Tested in my lab, works. Thanks for the pull request, merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3639, opened 2025-04-16
2 Checks are missing Column alias

DiHo78:
I just realized  the SELECT queries for checks 60 and 116 are missing some column aliases. This should be a non-breaking change,making  the code more consistent.


BrentOzar:
Just to be fair to everyone's time, can we focus our pull requests on things that will make a noticeable difference for end users? Thanks for understanding.

PR #3641, opened 2025-05-02
Issue 3640

### LINKED ISSUE #3640: sp_BlitzLock - New table mode breaks other stuff
**Version of the script**
Current

I was quite foolishly working on this feature across two issue branches without realizing it, and I had changes on one that did not make it to the other, primarily:
1. Hard-coded object names in table mode
2. Incomplete query logic for parsing XML data

Just run the darn thing. The `#dd` temp table doesn't populate correctly with the current code because the `@xe` and `@xd` variables remain `NULL` unless the new "table" code branch is accessed. 

Erik should mind his branches in the future. 

All versions of ErikOS contain this bug.

erikdarlingdata:
Closes #3640

Fixes hardcoded table and column names
Fixes #dd temp table not populating when not in "table" mode.

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -1255,7 +1255,7 @@ BEGIN
     END;
 
     /* If table target */
-    IF @TargetSessionType = 'table'
+    IF LOWER(@TargetSessionType) = N'table'
     BEGIN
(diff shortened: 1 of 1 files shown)


BrentOzar:
Thanks for the pull request! Merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #3651, opened 2025-06-16
Add Query Store Parameter to sp_ineachdb

bwiggin10:
This adds a new parameter to allow filtering of databases that have query store either enabled or disabled. This treats the model database as if it has query store disabled since it's marked as enabled in SQL 2022+ but is not a true query store enabled database.


BrentOzar (approved):
Thanks for the pull request, and sorry about the earlier comment - I had tunnel vision and I was still thinking of sp_Blitz, hahaha. Looks good, merging into the dev branch, will be in the next release with credit to you in the release notes.

Issue #3648, opened 2025-06-03
Additional Guidance for Empty AD Groups


jpalmer-n4:
**Is your feature request related to a problem? Please describe.**
When checking that Users/Groups exist in AD (invalid logins check), sp_blitz reports that an empty AD group doesn't exist at all. This is using sp_validatelogins.sql, so not code that is native to sp_blitz. Apparently this is well known to the Windows SysAdmins, but this DBA is still learning.

**Describe the solution you'd like**
Instead of the details reading:
_'Windows user or group [Domain\Group] is mapped to a SQL   Server principal but no longer exists in the Windows environment.'_

They should read:
_'Windows user or group [Domain\Group] is mapped to a SQL   Server principal but no longer exists in the Windows environment. Verify that this is not an empty group before taking action.'_

**Describe alternatives you've considered**
Learning things the hard way.

**Are you ready to build the code for the feature?**
Yes. It's an edit to https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_Blitz.sql#L1865
Someone just needs to approve the wording.

BrentOzar:
Thanks for the pull request! Makes sense, looks good, merging into the dev branch. It'll be in the next release with credit to you in the release notes.

PR #3652, opened 2025-06-19
Update sp_Blitz.sql

jpalmer-n4:
Extend details for Invalid Logins - empty AD Groups will occasionally show up here (it's an AD thing), and it's a pain to delete that group and then have to reinstate it in SQL when it turns out you do need it.

This relates to issue https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3648


BrentOzar (approved):
Thanks for the pull request! Makes sense, looks good, merging into the dev branch. It'll be in the next release with credit to you in the release notes.

PR #3654, opened 2025-06-26
Exclude Model Database From @QueryStoreInUse Check

bwiggin10:
The model database always returns 1 for is_query_store_on in sys.databases in SQL 2022, so this will always show query store is in use even if no user databases have it enabled.


BrentOzar:
Thanks for the pull request! Makes sense, looks good, merging into the dev branch. It'll be in the next release with credit to you in the release notes.

PR #3687, opened 2025-08-19
Added @UsualOwnerOfJobs to be used in CheckID 6 Jobs Owned By Users

JamesDavis-Aviagen:
to set your default SA user.
I did not use @UsualDBOwner you do not have these the same. Amend


BrentOzar (changes requested):
Each pull request needs to be confined to just one thing we want to fix. Can you revert out the unrelated changes, and just focus on the things in the title of this pull request? Thanks for understanding!

BrentOzar (approved):

PR #3705, opened 2025-09-08
Correctly filter out new indexes

TooMuchBlue:
A few checks in sp_BlitzIndex were written to filter out tables created in the last 7 days or modified in the last 2 days, but the less-than/greater-than check was reversed.  This fixes them.

check_id 21: More Than 5 Percent NC Indexes Are Unused
check_id 29: NC indexes with 0 reads and < 10,000 writes

The relevant lines also had tabs where surrounding lines used spaces, so I fixed that too.

I verified the direction of all other newer/older checks in sp_BlitzIndex and they seem to be correct.  I searched the rest of the codebase for similar filters and found none.


BrentOzar (approved):
Great catch! Thanks for the pull request, merging into the dev branch, will be in the next release with credit to you in the release notes.

Issue #3712, opened 2025-09-29
sp_Blitz: shows redundant results for RDS server triggers


yoni-yad2:
**Version of the script**
2025-07-04

**What is the current behavior?**
```
Server Trigger [rds_alter_database_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_audit_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_create_database_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_create_login_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_credential_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_das_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_deny_backups_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_drop_database_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_drop_login_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
Server Trigger [rds_extended_events_trigger] is enabled.  Make sure you understand what that trigger is doing - the less work it does, the better.
```
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
These server triggers with prefix `rds_` should be ignored by `sp_Blitz`

**Which versions of SQL Server and which OS are affected by this issue? Did this work in previous versions of our procedures?**
SQL Server 2019 on Amazon RDS

yoni-yad2:
Also, all checks on database `rdsadmin` are redundant as users don't have control over it.

BrentOzar:
OK, great! Did you plan to submit a pull request to change these? This would definitely be a pull request we'd be interested in merging.

PR #3713, opened 2025-09-29
Exclude RDS objects and rdsadmin DB from sp_Blitz checks

yoni-yad2:
https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/3712

### DIFF
--- a/sp_Blitz.sql
+++ b/sp_Blitz.sql
@@ -570,8 +570,7 @@ AS
         SELECT
 		    DB_NAME(d.database_id)
         FROM sys.databases AS d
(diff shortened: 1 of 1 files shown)


BrentOzar (changes requested):
Hi! Thanks for the pull request, but it looks like maybe some unrelated changes snuck in? Or can you explain these other changes, not sure why they were made? Sorry about the delay.

BrentOzar:
@yoni-yad2 yes, just busy - can you resubmit a pull request with only the relevant changes? I know it sounds dumb, but it's volunteers here on the other side of the screen, and I need to be careful with what we have to test. Hope that's fair.

Issue #3765, opened 2025-12-02
sp_BlitzLock - Duplicate available plans


Tisit:
**Version of the script**
8.28

**What is the current behavior?**

There are at least 3 scenarios where we can get dupliate entries for available plans from sp_BlitzLock:

1. Procedure involved in deadlock has many queries
2. Procedure is called with different parameters
3. The query involves many tables and deadlock may occur on any of them (I think)

Setup:

```
DROP TABLE IF EXISTS dbo.LockTest

CREATE TABLE dbo.LockTest (
    Id INT NOT NULL PRIMARY KEY,
    Value INT NOT NULL
);
GO

CREATE OR ALTER PROC p_UpdateLockTest (@p1 int) AS
BEGIN

    SELECT count(1) c into #t2 from sys.indexes

    UPDATE dbo.LockTest
    SET Value = Value + 1
    WHERE Id = @p1;

END

GO

INSERT INTO dbo.LockTest (Id, Value) VALUES (1, 100), (2, 200);
GO

--clean out plan cache to better see the issue
DBCC FREEPROCCACHE()
```
Session A:

```
BEGIN TRAN;

EXEC p_UpdateLockTest @p1 = 1

GO

EXEC p_UpdateLockTest @p1 = 2

--ROLLBACK
```
Session B:

```
BEGIN TRAN;

EXEC p_UpdateLockTest @p1 = 2

GO

EXEC p_UpdateLockTest @p1 = 1

--ROLLBACK
```

1. Set up table, procedure and insert data
2. Open 2 sessions A and B
3. In session A start the transaction and execute the first procedure
4. In session B start the transaction and execute the first procedure
5. In session A execute the second procedure. It will hang
6. In session B execute the second procedure. It will succeed or become a deadlock victim
7. execute sp_BlitzLock and observe the available plans

<img width="1698" height="150" alt="Image" src="https://github.com/user-attachments/assets/8f23a040-e73a-49cf-889f-8714273efcfa" />

This reproducesr shows first 2 of the issues mentioned. The last one is quite hard to reproduce for me. Sorry.

Even though only one procedure was involved we can see it duplicated 6 times in my case. Some entries are obviosly duplicate of others.
Other seem to provide data for unrelated query from the same procedure. You can see the statistics and statement offfsets are different.

**What is the expected behavior?**
Only one row is returned for involved statement

[...]

PR #3773, opened 2025-12-10
sp_BlitzLock - fix duplicate plans

Tisit:
Fix for #3765 (or at least improvements)

Turns out there are many causes of this:
- if there are many statements in procedure we need to use specific statement start and end columns, otherwise we get something close to cartesian join
- We may get duplicate query plans in #deadlock_results -> I take just max query_xml  and group by rest of the columns. Seemed easiest way to fix for me
- one query can have deadlock on many different tables. This caused same SQL to be reported many times. At least that is my interpretation, because removing dow.object_name on line 4145 (which was unused anyway) and aggregating seemed to fix this particular problem.

In the environment where I saw all these problems the number of rows returned dropped from 10k+ to 60. Much more managable in my opinion.

Thanks

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -4316,8 +4316,8 @@ BEGIN
                         ) AS deps
                         WHERE deqs.sql_handle = ap.sql_handle
                         AND   deps.dbid = ap.database_id
-                        AND   deqs.statement_start_offset = ap.stmtstart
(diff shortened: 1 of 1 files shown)


BrentOzar:
Sorry about the delay! Got it tested & working, merged into the dev branch, will be in the next release with credit to you in the release notes. Thanks!

PR #3783, opened 2026-01-02
sp_Blitz - Use Last Hardened Time for AG Behind Check if Last Commit Time is NULL

iaint-at-scc:
Fixes 3775 - sp_Blitz Does not report a problem with the AG If an AG Secondary is Offline


BrentOzar (approved):

Issue #3791, opened 2026-01-13
sp_BlitzCache - Performance issues due to duplicate sys.dm_exec_query_plan_stats calls


connormoolman:
**Version of the script**
Version = '8.28', VersionDate = '20251124';

**What is the current behavior?**
The procedure runs slowly with certain sort orders ('avg cpu' for example) due to the existence of a redundant sys.dm_exec_query_plan_stats CROSS APPLY. FNGETQUERYPLANSTATS is being executed for almost all rows in sys.dm_exec_query_stats before Top N filtering is applied.

<img width="598" height="370" alt="Image" src="https://github.com/user-attachments/assets/34c8064f-d66a-4c0b-b583-f98201614f92" />

Another FNGETQUERYPLANSTATS Table Valued Function operator exists in the plan with more reasonable times as it is only running once for each row after the Top N has been applied:

<img width="624" height="243" alt="Image" src="https://github.com/user-attachments/assets/79b97193-ed16-46f1-be8e-442a92a7b8e4" />

The procedure is adding this to the body variable twice:

<img width="1258" height="138" alt="Image" src="https://github.com/user-attachments/assets/97b45b7c-583f-467f-b816-367ebc90af5c" />

<img width="1249" height="124" alt="Image" src="https://github.com/user-attachments/assets/fe4cf735-9df9-4658-9f86-44b3278d8fc7" />

This can be seen in the output of Debug = 1:

```
FROM   (SELECT TOP (@Top) x.*, xpa.*,
               CAST((CASE WHEN DATEDIFF(mi, creation_time, GETDATE()) > 0 AND execution_count > 1
                          THEN DATEDIFF(mi, creation_time, GETDATE()) 
                          ELSE NULL END) as MONEY) as age_minutes,
               CAST((CASE WHEN DATEDIFF(mi, creation_time, last_execution_time) > 0 AND execution_count > 1
                          THEN DATEDIFF(mi, creation_time, last_execution_time) 
                          ELSE Null END) as MONEY) as age_minutes_lifetime
        FROM   sys.dm_exec_query_stats x
               CROSS APPLY (SELECT * FROM sys.dm_exec_plan_attributes(x.plan_handle) AS ixpa 
                            WHERE ixpa.attribute = 'dbid') AS xpa 
     CROSS APPLY sys.dm_exec_query_plan_stats(x.plan_handle) AS deqps -- HERE
        WHERE  1 = 1 
[...]

PR #3793, opened 2026-01-22
sp_BlitzCache - fix for #3791

ConnorMoolman:
Improves performance for calculated sort orders (avg, xpm, etc.) by removing a redundant sys.dm_exec_query_plan_stats call as per issue #3791.


connormoolman:
Hey Brent, only the first instance of that cross apply is commented out. There's another deqps (qs.plan_handle instead of x.plan_handle), and only this one is used:

image


When reviewing the debug output of the current procedure, both the x.plan_handle and qs.plan_handle iterations exist. The first is in the 'qs' subquery, which is not actually exposing deqps - only 'x' and 'xpa' are being selected from. 

image

There is also no case where only one of these are added to the executing dynamic SQL. If @VersionShowsAirQuoteActualPlans = 1, they are both added.

PR #3921, opened 2026-04-15
sp_BlitzFirst: add Top 5 server/login/app breakdown to High Number of Connections

### LINKED ISSUE #3903: sp_BlitzFirst: add status line for number of connections
When I'm troubleshooting a high connections problem, sp_BlitzFirst does have an alert for High Number of Connections. When that alerts, I'd like the Details column to include a list of the top 5 server names, top 5 login names, top 5 app names, sorted by the number of connections descending, and the min & max query finish time for each one. This will help track down if an app or server is leaving too many connections open, or if they're actually using their connections to run queries.

The list should be separated with lines, like:
Top 5 Servers:
AppServer14 - 106 connections, most recent query finished 3 seconds ago, oldest query finished 4 hours 15 minutes ago
WebServer11 - 98 connections, most recent query finished 14 seconds ago, oldest query finished 3 hours 15 minutes ago

BrentOzar:
Closes #3903.

## Summary

When CheckID 49 (High Number of Connections) fires, the `Details` column now includes three ranked sections so operators can see **where** the connection pressure is concentrated:

- **Top 5 Servers** — by `sys.dm_exec_sessions.host_name`
- **Top 5 Logins** — by `login_name`
- **Top 5 Apps** — by `program_name`

Each line shows the connection count plus the most-recent and oldest query-finish ages for that group:

```
AppServer14 - 106 connections, most recent query finished 3 seconds ago, oldest query finished 4 hours 15 minutes ago
WebServer11 - 98 connections, most recent query finished 14 seconds ago, oldest query finished 3 hours 15 minutes ago
```

## Implementation notes

- **DMVs read once.** Connection + session attributes land in a `TABLE` variable that's aggregated three different ways. No correlated triple-scan.
- **Gated on the alert firing.** Total connection count is computed into `@TotalConnections` first; the breakdown work only runs when `@TotalConnections > @max_worker_threads`. No extra cost on healthy servers.
- **Epoch sentinel handled.** `sys.dm_exec_sessions.last_request_end_time` defaults to `1900-01-01` for sessions that have never run a request. That gets `NULLIF`'d so the row renders as "unknown" rather than "45000 days ago".
- **Unknown buckets.** NULL/empty `host_name`, `login_name`, `program_name` collapse into `(unknown host)` / `(unknown login)` / `(unknown app)` so internal or anonymous connections don't drop out of the count.
- **Entity-safe concatenation.** Uses `.value(N'.[1]', N'NVARCHAR(MAX)')` — the same pattern already used elsewhere in sp_BlitzFirst — so `program_name` values containing `<`, `>`, or `&` don't truncate the output.
- **Deterministic ordering.** `ORDER BY ConnectionCount DESC, ConnectionGroup` — no ties cause random row shuffling.

Alternative to this PR: #3904 (same feature, different approach). This version avoids the scalar-aggregate + CROSS APPLY compile issue flagged in review there and collapses the age-formatting CASE from 12 copies to 6.

## Test plan

- [ ] On a server with < `@max_worker_threads` connections, confirm the check is silent (no row in `#BlitzFirstResults`).
- [ ] Flood the server with connections (e.g. `ostress -n3000` or a load generator against a VM with low core count so `@max_worker_threads` is low), confirm the alert fires and the three Top-5 sections populate.
- [ ] Verify a mix of connection sources: different hosts, logins, and `program_name` values — confirm counts sum correctly to the total.
- [ ] Confirm connections from a brand-new session (no request ever run) show "unknown" rather than a huge day count.
- [ ] Confirm an app whose `program_name` contains `<`, `>`, or `&` still appears in the Top 5 Apps list (exercises the `.[1]` XPath fix).
- [ ] Confirm the check is skipped entirely on non-64-bit Windows / Azure SQL DB (existing guard unchanged).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


BrentOzar:
Thanks @Copilot. Addressed in `dc10aef`:

- **XPath comments (lines 2735/2772/2809):** Valid catch. Switched all three `.value('text()[1]', 'nvarchar(max)')` to `.value(N'.[1]', N'NVARCHAR(MAX)')` — matches the existing FRK pattern at line 2487 and protects against `program_name` values containing `<`, `>`, or `&`, which get entity-escaped into separate text nodes.
- **Missing line break after headers (line 2823):** Valid catch. Applied the suggested fix.
- **Version bump (line 2671):** Not applying. The `@Version` / `@VersionDate` line around line 50 of each sp is maintained by the build process; the earlier commit `cd914b7` reverted my original bump for that reason. I've dropped the version-bump line from the PR description so it's no longer misleading.

PR #3968, opened 2026-04-26
sp_BlitzBackups: priority changes and doc for #3963

### LINKED ISSUE #3963: sp_BlitzBackups - separate checks into different priority levels, document them
The title.
IMHO there are some findings that would warrant higher priorities.
For example findings like "Damaged backup" and "No CHECKSUMS" could be considered more impactful as opposed to something like "Uncompressed backups".

VladDBA:
Closes #3963 

**Priority 10 — Fix immediately**
- **Damaged backups** (8): Backups taken with `CONTINUE_AFTER_ERROR` may not be restorable at all.
- **Recovery model switched** (11): Breaks the log chain. Point-in-time recovery is compromised right now.
- **Backup to NUL without COPY_ONLY** (14): Backups don't exist *and* the backup chain is broken.

**Priority 20 — Fix soon**
- **No CHECKSUMS** (7): You won't know about corruption until restore time (silent risk).
- **Backup to NUL** (14): Backups don't exist, but at least the chain isn't broken (COPY_ONLY or SIMPLE recovery).

**Priority 50 — Worth investigating**
- **Single user mode backups** (6): Operationally dangerous but not an immediate data loss risk.
- **Big Diffs/Logs** (13): Signals a backup strategy that may not meet RTO. Worth reviewing, not an emergency.

**Priority 80 — Best practice gap**
- **Uncompressed backups** (12): Wastes time and space but data is still safe and recoverable.

**Priority 100 — Informational/investigate**
- Non-Agent backups, compatibility level changes, password backups, encrypted backups, bulk logged backups. All worth knowing about and investigating, but none directly threaten recoverability right now.

**Priority 150/200 — Noise**
- **Read-only state backups** (5): Explicitly called out as potentially normal (AG secondaries).
- **Snapshot backups** (4): The code comment literally says "purely informational."

### DIFF
--- a/Documentation/sp_BlitzBackups_Checks_by_Priority.md
+++ b/Documentation/sp_BlitzBackups_Checks_by_Priority.md
@@ -4,7 +4,7 @@ This table lists all checks ordered by priority.
 
 Before adding a new check, make sure to add a Github issue for it first, and have a group discussion about its priority, and description.
 
-If you want to change anything about a check - the priority, finding,, or ID - open a Github issue first. The relevant scripts have to be updated too.
+If you want to change anything about a check - the priority, finding, or ID - open a Github issue first. The relevant scripts have to be updated too.
 
 CURRENT HIGH CHECKID: 14.
 If you want to add a new one, start at 15.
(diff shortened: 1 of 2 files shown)


BrentOzar (approved):

BrentOzar (approved):
Good catch.

PR #3974, opened 2026-04-27
fix for #3973 - CheckID 11 not working

### LINKED ISSUE #3973: sp_BlitzBackups: CheckID 11 "Recovery model switched" would never show up with its current filtering
**Version of the script**
8.30

If you have a database and have switched recovery models between FULL and SIMPLE for multiple backups, then CheckID 11 would never return a finding.

```sql
	SET @StringToExecute += N'SELECT 
		11 AS CheckId,
		10 AS [Priority],
		b.database_name AS [Database Name],
		''Recovery model switched'' AS [Finding],
		''The database '' + QUOTENAME(b.database_name) + '' has changed recovery models between FULL and SIMPLE '' + CONVERT(VARCHAR(10), COUNT(DISTINCT b.recovery_model)) + '' times. This breaks the log chain and is generally a bad idea.'' AS [Warning]
	FROM   ' + QUOTENAME(@MSDBName) + '.dbo.backupset AS b
	WHERE b.recovery_model <> ''BULK-LOGGED''
	GROUP BY b.database_name
	HAVING COUNT(DISTINCT b.recovery_model) > 4;' + @crlf;
```

Reason being the filtering that's done in the `HAVING` clause (`HAVING COUNT(DISTINCT b.recovery_model) > 4`) since we'll always have a max of 2 distinct recovery models that condition will always exclude those backups.

```sql
ALTER DATABASE [AdventureWorks2019] SET RECOVERY FULL WITH NO_WAIT;
GO
BACKUP DATABASE [AdventureWorks2019] TO DISK = N'NUL';
GO
ALTER DATABASE [AdventureWorks2019] SET RECOVERY SIMPLE WITH NO_WAIT;
GO
BACKUP DATABASE [AdventureWorks2019] TO DISK = N'NUL';
GO
ALTER DATABASE [AdventureWorks2019] SET RECOVERY FULL WITH NO_WAIT;
GO
BACKUP DATABASE [AdventureWorks2019] TO DISK = N'NUL';
GO
ALTER DATABASE [AdventureWorks2019] SET RECOVERY SIMPLE WITH NO_WAIT;
GO
BACKUP DATABASE [AdventureWorks2019] TO DISK = N'NUL';
GO
```

```sql
EXEC sp_BlitzBackups;
```

There should be a row for the recovery model switched finding here, but there isn't.

Image

Image

all of them

VladDBA:
Closes #3973

### DIFF
--- a/sp_BlitzBackups.sql
+++ b/sp_BlitzBackups.sql
@@ -1063,16 +1063,21 @@ IF @ProductVersionMajor >= 12
 
 	SET @StringToExecute =N'SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;' + @crlf;
 
-	SET @StringToExecute += N'SELECT 
+	SET @StringToExecute += N';WITH RecoveryHistory AS (
+		SELECT b.database_name, b.recovery_model,
+			LAG(b.recovery_model) OVER (PARTITION BY b.database_name ORDER BY b.backup_finish_date, b.backup_set_id) AS prev_recovery_model
+		FROM ' + QUOTENAME(@MSDBName) + '.dbo.backupset AS b
+		WHERE b.recovery_model <> ''BULK-LOGGED''
+	)
+	SELECT
 		11 AS CheckId,
 		10 AS [Priority],
-		b.database_name AS [Database Name],
+		database_name AS [Database Name],
 		''Recovery model switched'' AS [Finding],
-		''The database '' + QUOTENAME(b.database_name) + '' has changed recovery models between FULL and SIMPLE '' + CONVERT(VARCHAR(10), COUNT(b.recovery_model)) + '' times. This breaks the log chain and is generally a bad idea.'' AS [Warning]
-	FROM   ' + QUOTENAME(@MSDBName) + '.dbo.backupset AS b
-	WHERE b.recovery_model <> ''BULK-LOGGED''
-	GROUP BY b.database_name
-	HAVING COUNT(DISTINCT b.recovery_model) = 2;' + @crlf;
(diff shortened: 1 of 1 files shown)


BrentOzar (approved):
Thanks for the pull request! Looks like it makes sense, merging into the dev branch, will be in the next release with credit to you in the release notes.

BrentOzar:
@VladDBA — pushed a follow-up commit on top of yours to address Copilot's review note on the warning text.

The original `COUNT(b.recovery_model)` was counting backup rows (so a database with frequent log backups would report it had "changed recovery models" thousands of times). Replaced the query with a CTE that uses `LAG` over `backup_finish_date` to count actual FULL ↔ SIMPLE transitions. The `HAVING` was rewritten to `SUM(...transitions...) > 0`, which is equivalent to your `COUNT(DISTINCT b.recovery_model) = 2` but lets the same expression drive both the filter and the count in the message.

Happy to back this out if you'd rather take a different approach — your call since it's your PR.

(All written/coded by Claude Code - Brent)

VladDBA:
> @VladDBA — pushed a follow-up commit on top of yours to address Copilot's review note on the warning text.
> 
> The original `COUNT(b.recovery_model)` was counting backup rows (so a database with frequent log backups would report it had "changed recovery models" thousands of times). Replaced the query with a CTE that uses `LAG` over `backup_finish_date` to count actual FULL ↔ SIMPLE transitions. The `HAVING` was rewritten to `SUM(...transitions...) > 0`, which is equivalent to your `COUNT(DISTINCT b.recovery_model) = 2` but lets the same expression drive both the filter and the count in the message.
> 
> Happy to back this out if you'd rather take a different approach — your call since it's your PR.
> 
> (All written/coded by Claude Code - Brent)

Awesome! I was just about to switch this to draft and make the lazy version of the fix "has changed recovery models between FULL and SIMPLE multiple times" 😅

BrentOzar:
I really love Claude Code for that exact reason - I find that the quality of stuff I ship has gone way up because I can now afford to say, "Lemme ask the robot to do this the right way, instead of the lazy way I would usually do it," hahaha.

PR #3982, opened 2026-05-03
sp_BlitzIndex: HAS_DBACCESS preflight to avoid hang on inaccessible DBs

### LINKED ISSUE #3981: sp_BlitzIndex hangs at 100% CPU when caller has no user mapping in target database
## Symptom

`sp_BlitzIndex` hangs indefinitely at 100% CPU (no waits) when the executing principal has no user mapping in a database it is asked to analyze. The proximate cause is a query against `sys.dm_db_partition_stats` via three-part name (e.g. `[YourDb].sys.dm_db_partition_stats`).

Reproduces on SQL Server 2016 SP3 through 2025 CU4.

## Root cause

This is a SQL Server engine bug, not an FRK bug per se: the compiled plan trips a permission check at execute time, and the engine misclassifies that as "plan needs recompile." It then loops trying to recompile forever instead of returning Msg 916. Every other catalog view referenced in `sp_BlitzIndex` (`sys.indexes`, `sys.objects`, `sys.partitions`, `sys.allocation_units`, `sys.dm_db_index_usage_stats`) fast-fails with Msg 916 in the same scenario; `sys.dm_db_partition_stats` is the odd one out.

Tracked in detail under https://github.com/erikdarlingdata/PerformanceMonitor/issues/915, where it was first reported against `sp_IndexCleanup` (DarlingData) and fixed there in https://github.com/erikdarlingdata/DarlingData/pull/778.

## Where in sp_BlitzIndex

- **Multi-database path (`@GetAllDatabases = 1`)** has a partial defense — there is an existing TRY/CATCH preflight against `sys.indexes` that drops inaccessible DBs into `#DatabasesNoAccess`. That works (because `sys.indexes` Msg 916s). However it is gated by `IF IS_SRVROLEMEMBER('sysadmin') = 0`, and the comment at the top of that block already acknowledges that Azure SQL DB principals like `##MS_DatabaseConnector##` / `CONNECT ANY DATABASE` report as sysadmin-equivalent yet can still be denied CONNECT to specific databases. Those principals skip the preflight and can hang on `sys.dm_db_partition_stats`.
- **Single-database path (explicit `@DatabaseName`)** has no access preflight at all. A non-sysadmin login that calls `EXEC sp_BlitzIndex @DatabaseName = N'NoAccessDb'` hangs as soon as the analysis loop hits `sys.dm_db_partition_stats` (first reference is in the main partition_stats insert; further references exist in the helpful-index check and the internal-tables join).

## Manual repro

```sql
USE master;
CREATE LOGIN low_priv WITH PASSWORD = 'StrongP@ssw0rd!';
GRANT VIEW SERVER STATE TO low_priv;
-- intentionally no CREATE USER in any user database

-- as low_priv:
EXEC dbo.sp_BlitzIndex @DatabaseName = N'SomeUserDb';
-- spins at 100% CPU forever
```

To clear: `KILL` the session.

The fix the user has to apply manually (per database) is:

```sql
CREATE USER [low_priv] FOR LOGIN [low_priv];
GRANT VIEW DATABASE STATE TO [low_priv];
GRANT VIEW DEFINITION TO [low_priv];
```

## Proposed fix

Filter database enumeration via `HAS_DBACCESS(name) = 1` before any three-part-name query is issued:

1. Add `AND HAS_DBACCESS(DB_NAME(database_id)) = 1` to the `@GetAllDatabases = 1` insert into `#DatabaseList`.
2. In the single-database `ELSE` branch, preflight `HAS_DBACCESS(@DatabaseName)`; if it returns 0 (or NULL), `RAISERROR` with severity 16 telling the operator which `CREATE USER` / `GRANT` calls to run, and `RETURN`.

`HAS_DBACCESS` is per-session and reflects the actual principal, so it correctly gates the Azure-special "sysadmin-equivalent but denied CONNECT" case that the existing TRY/CATCH bypasses. It also avoids issuing dynamic SQL just to probe access.

PR to follow.

erikdarlingdata:
Fixes #3981.

## Summary

When the executing principal has no user mapping in a target database, `sp_BlitzIndex`'s three-part-name queries against `sys.dm_db_partition_stats` hang at 100% CPU forever instead of returning `Msg 916`. This is a SQL Server engine bug: the compiled plan trips a permission check at execute time, the engine misclassifies that as "plan needs recompile," and loops. Reproduces on 2016 SP3 through 2025 CU4.

We can't fix the engine, so this PR preflights access and skips cleanly. The same fix pattern was applied to `sp_IndexCleanup` in DarlingData (https://github.com/erikdarlingdata/DarlingData/pull/778), which is where the issue was originally tracked: https://github.com/erikdarlingdata/PerformanceMonitor/issues/915.

## Changes

`sp_BlitzIndex.sql`:

- **`@GetAllDatabases = 1` path**: add `AND HAS_DBACCESS(DB_NAME(database_id)) = 1` to the `#DatabaseList` insert. This catches the Azure SQL DB case where principals like `##MS_DatabaseConnector##` / `CONNECT ANY DATABASE` report as sysadmin-equivalent (so they bypass the existing `IS_SRVROLEMEMBER('sysadmin') = 0`-gated TRY/CATCH preflight) but can still be denied `CONNECT` per database. The comment in the existing block already acknowledges this scenario.
- **Explicit `@DatabaseName` path**: preflight `HAS_DBACCESS(@DatabaseName)`. If 0/NULL, `RAISERROR` (severity 16) with a clear message telling the operator which `CREATE USER` / `GRANT` calls to run, then `RETURN`. Without this, the procedure runs to the analysis loop and hangs on the first `sys.dm_db_partition_stats` reference.

The existing TRY/CATCH preflight against `sys.indexes` is left in place as defense-in-depth — it works (because `sys.indexes` does fast-fail with `Msg 916`) and removing it would be churn unrelated to this fix.

No public parameter surface change. No unrelated rewrites.

## Manual repro

```sql
USE master;
CREATE LOGIN low_priv WITH PASSWORD = 'StrongP@ssw0rd!';
GRANT VIEW SERVER STATE TO low_priv;
CREATE USER low_priv FOR LOGIN low_priv;       -- in master, just so they can EXECUTE sp_BlitzIndex
GRANT EXECUTE ON dbo.sp_BlitzIndex TO low_priv;
-- intentionally no CREATE USER for low_priv in any other user database

-- as low_priv:
EXEC dbo.sp_BlitzIndex @DatabaseName = N'SomeUserDb';
-- before this PR: spins at 100% CPU forever (KILL the session to clear)
-- after this PR:  errors in <1 second with a GRANT hint

EXEC dbo.sp_BlitzIndex @GetAllDatabases = 1;
-- before this PR: hangs on the first inaccessible user DB
-- after this PR:  finishes cleanly; inaccessible DBs are silently skipped
```

The fix the user has to apply manually (per database) once they see the error is:

```sql
CREATE USER [low_priv] FOR LOGIN [low_priv];
GRANT VIEW DATABASE STATE TO [low_priv];
GRANT VIEW DEFINITION TO [low_priv];
```

## Test plan

- [x] Deploys cleanly on SQL Server 2022 (`CREATE OR ALTER` round-trips).
- [x] Happy path: `EXEC sp_BlitzIndex @DatabaseName = N'master'` as `sa` returns results normally.
- [x] Single-DB repro: low-priv login + `@DatabaseName = N'<inaccessible>'` errors in <1s with the new message instead of hanging.
- [x] Multi-DB repro: low-priv login + `@GetAllDatabases = 1, @BringThePain = 1` returns in ~1s with no hang; inaccessible DBs are excluded from `#DatabaseList`.
- [ ] Reviewer to confirm on a 2016/2017/2019 instance if available — `HAS_DBACCESS` exists on all of those.

### DIFF
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1216,6 +1216,13 @@ IF @AI > 0
 
 IF @GetAllDatabases = 1
     BEGIN
+        /* HAS_DBACCESS gates databases the current principal can actually query.
+           Without this filter, three-part-name queries against sys.dm_db_partition_stats
+           hang indefinitely under a SQL Server permission/recompile bug instead of
+           erroring with Msg 916. The TRY/CATCH preflight below catches the same case
+           for legacy non-sysadmin paths, but HAS_DBACCESS also covers Azure principals
+           (##MS_DatabaseConnector##, CONNECT ANY DATABASE) that report as
+           sysadmin-equivalent yet can still be denied CONNECT per database. */
         INSERT INTO #DatabaseList (DatabaseName)
         SELECT  DB_NAME(database_id)
         FROM    sys.databases
@@ -1226,6 +1233,7 @@ IF @GetAllDatabases = 1
         AND DB_NAME(database_id) NOT LIKE 'rdsadmin%'
 		AND LOWER(name) NOT IN('dbatools', 'dbadmin', 'dbmaintenance', 'gcloud_cloudsqladmin')
         AND is_distributor = 0
+        AND HAS_DBACCESS(DB_NAME(database_id)) = 1
 		OPTION    ( RECOMPILE );
 
         /*Check if sysadmin 
@@ -1325,10 +1333,23 @@ IF @GetAllDatabases = 1
     END;
 ELSE
     BEGIN
+        /* Preflight: if an explicit @DatabaseName was supplied but the current principal
+           has no access to it, fail loud rather than letting downstream queries hang.
+           sp_BlitzIndex queries sys.dm_db_partition_stats via three-part name, which
+           under a SQL Server permission/recompile bug spins at 100% CPU forever instead
+           of erroring with Msg 916. */
+        IF @DatabaseName IS NOT NULL AND @DatabaseName <> N''
+        AND DB_ID(@DatabaseName) IS NOT NULL
+        AND ISNULL(HAS_DBACCESS(@DatabaseName), 0) = 0
+            BEGIN
+                RAISERROR(N'The current login has no access to database %s. Run, against that database: CREATE USER [<login>] FOR LOGIN [<login>]; GRANT VIEW DATABASE STATE; GRANT VIEW DEFINITION;', 16, 1, @DatabaseName);
+                RETURN;
+            END;
+
         INSERT INTO #DatabaseList
                 ( DatabaseName )
-        SELECT CASE 
-		            WHEN @DatabaseName IS NULL OR @DatabaseName = N'' 
+        SELECT CASE
+		            WHEN @DatabaseName IS NULL OR @DatabaseName = N''
 		            THEN DB_NAME()
                     ELSE @DatabaseName END;
                END;
--- a/sp_BlitzIndex.sql
+++ b/sp_BlitzIndex.sql
@@ -1342,7 +1342,8 @@ ELSE
         AND DB_ID(@DatabaseName) IS NOT NULL
         AND ISNULL(HAS_DBACCESS(@DatabaseName), 0) = 0
             BEGIN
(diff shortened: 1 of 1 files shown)


BrentOzar:

PR #3996, opened 2026-05-13
sp_BlitzCache: fix empty <ai_prompt /> on parent procedure rows

### LINKED ISSUE #3995: sp_BlitzCache: AI Prompt column shows empty <ai_prompt /> on parent procedure rows when their statements are also in the result set
**Version of the script**
8.32 (sp_BlitzCache.sql:301)

When running `sp_BlitzCache` with `@AI = 1` or `@AI = 2`, the **AI Prompt** column on some rows is rendered as a self-closing, empty XML element:

```xml
<ai_prompt />
```

The affected rows are `Procedure or Function:` parent rows whose corresponding `Statement (parent ...)` rows are also present in the same result set.

In my local reproduction (SQL Server 2025, StackOverflow workload), running `EXEC dbo.sp_BlitzCache @AI = 2, @Top = 50` returned 7 of 59 rows with `<ai_prompt />`. All 7 were procedure-parent rows for procs whose statements were also in the result.

**Steps to reproduce**

1. Check out the current `dev` branch.
2. Install `sp_BlitzCache.sql` into any database.
3. Run against a server whose plan cache contains stored procedures with multiple statements (so both the parent proc row and one or more child statement rows show up in the top-N result):

   ```sql
   EXEC dbo.sp_BlitzCache @AI = 2, @Top = 50;
   ```

   `@AI = 2` generates the prompt without invoking the external API, so this reproduces with no AI credentials required and without freeing the plan cache.

4. Inspect the **AI Prompt** column on rows where `Query Type` starts with `Procedure or Function:` and there is also a matching `Statement (parent ...)` row in the output.

For `@AI = 2`, every returned row should carry a usable `ai_prompt` — the user is going to paste that prompt into an external AI tool, and a parent-procedure prompt is legitimately useful (it lets the user reason about the procedure as a whole rather than statement-by-statement).

For `@AI = 1`, parent-with-statements rows should not trigger a redundant API call (each statement is analyzed separately), but the **AI Advice** column on those parent rows should still carry an explanation rather than appearing blank.

**Root cause**

Three pieces of logic interact to produce the empty element:

1. **The metrics UPDATE intentionally skips parent rows that have child statements in the result set.**
   `sp_BlitzCache.sql:5249-5354`. The `WHERE` clause excludes any row where `QueryType LIKE 'Procedure or Function:%'` *and* a matching `Statement (parent …)` row exists for the same plan handle. This was added in commit `70c0acf` (#3669 - AI - skip calls for parents with kids) to avoid asking the model to redundantly analyze a parent proc when we are already going to send each statement individually.

2. **The fallback "no text or plan" UPDATE does not cover this case.**
   `sp_BlitzCache.sql:5357-5362` sets `ai_prompt = N'Prompt not generated because we can''t find the query text or query plan.'` — but only when **both** `QueryText IS NULL AND QueryPlan IS NULL`. Parent rows have non-NULL `QueryText` *and* non-NULL `QueryPlan`, so they do not match. Their `ai_prompt` stays `NULL`.

3. **The output projection concatenates the system prompt with a NULL `ai_prompt`, producing NULL, which `FOR XML PATH` renders as `<ai_prompt/>`.**
   `sp_BlitzCache.sql:5724-5728` (Expert Mode 2) and `5862-5866` (default / `@AI = 1`):

   ```sql
   [AI Prompt] = CASE WHEN QueryText IS NULL AND QueryPlan IS NULL THEN (
       SELECT ai_prompt AS [text()] FOR XML PATH('ai_prompt'), TYPE)
   ELSE (
       SELECT (@AISystemPrompt + NCHAR(13) + NCHAR(10) + NCHAR(13) + NCHAR(10) + ai_prompt) AS [text()] FOR XML PATH('ai_prompt'), TYPE) END,
   ```

   Because `QueryText`/`QueryPlan` are populated on the parent row, the `ELSE` branch runs. Standard SQL string concatenation with a `NULL` operand returns `NULL`, and `FOR XML PATH('ai_prompt'), TYPE` over `NULL` emits the self-closing `<ai_prompt/>` element.

   Standalone proof:

   ```sql
   DECLARE @ai_prompt NVARCHAR(MAX) = NULL;
   DECLARE @AISystemPrompt NVARCHAR(4000) = N'system';
   SELECT (SELECT (@AISystemPrompt + NCHAR(13) + NCHAR(10) + NCHAR(13) + NCHAR(10) + @ai_prompt)
           AS [text()] FOR XML PATH('ai_prompt'), TYPE);
   -- Returns: <ai_prompt/>
   ```

The two related concerns — *"generate a usable prompt for display"* and *"don't make redundant AI API calls"* — are currently both gated by the same `ai_prompt IS NULL` state. Separating them resolves both `@AI = 1` and `@AI = 2` correctly.

**Suggested fix**

1. **Always populate `ai_prompt` for every row** — remove the parent-with-kids exclusion from the metrics UPDATE at `sp_BlitzCache.sql:5249`. Every row, including parent procedures, then has a full prompt available in the **AI Prompt** column.
2. **Move the parent-with-kids exclusion onto the AI cursor** at `sp_BlitzCache.sql:5382`. This is the only place where "don't call AI twice for the same optimization" actually matters. The cursor's existing `DISTINCT SqlHandle, QueryHash, PlanHandle` already dedupes exact handle/hash duplicates; the parent-with-kids check is the remaining piece.
3. **Stamp a clear `ai_advice` on the skipped parent rows** when `@AI = 1`, e.g. *"AI advice not generated for this parent procedure because its individual statements are being analyzed separately. See the AI Advice on each Statement (parent ...) row."* That keeps the column from rendering blank.

After applying the fix to my local repro:

- `@AI = 2`: 0 of 59 rows have NULL `ai_prompt` (was 7). All 7 parent-with-kids rows now carry full prompts (15K–50K chars each).
- `@AI = 1`: all 7 parent-with-kids rows carry the new `ai_advice` stub; 0 of them triggered an API call; the remaining 32 non-parent rows were processed by the cursor as expected.
- The rendered output contains 0 empty `<ai_prompt/>` elements.

**Which versions of SQL Server and which OS are affected by this issue?**

Reproduced on SQL Server 2025 (RTM-CU4-GDR) / Windows 11 with the latest `dev` branch. The affected logic was introduced in commit `70c0acf` (#3669) and is not version-gated, so all SQL versions where `@AI` mode is supported are affected.

Yes — PR to follow.

BrentOzar:
Fixes #3995.

## Summary

When `sp_BlitzCache` runs with `@AI = 1` or `@AI = 2`, some rows render the **AI Prompt** column as a self-closing, empty XML element (`<ai_prompt />`). The affected rows are `Procedure or Function:` parent rows whose corresponding `Statement (parent ...)` rows are also in the result set. Full diagnosis is in #3995.

## What changed

Two concerns were conflated under `ai_prompt IS NULL`:

| Concern | Where it applies | Where it was enforced before | Where it is enforced now |
|---|---|---|---|
| Generate a usable prompt for display | `@AI = 1` and `@AI = 2` | Skipped on parent-with-kids rows in the metrics `UPDATE` | Always applied — every row gets a prompt |
| Avoid redundant API call for the same optimization | `@AI = 1` only | Implicit (parent rows had NULL `ai_prompt`, so the cursor skipped them) | Explicit `NOT EXISTS` predicate on the cursor's `WHERE` |

Three edits in `sp_BlitzCache.sql`:

1. **Metrics `UPDATE` (around line 5258)** — removed the parent-with-kids exclusion. Every row in `##BlitzCacheProcs` now gets an `ai_prompt`.
2. **AI cursor `WHERE` (around line 5395)** — added the parent-with-kids `NOT EXISTS` clause. The cursor still skips parents whose statements are also in the result set, so no redundant API calls.
3. **New `UPDATE` inside `IF @AI = 1` (around line 5351)** — stamps `ai_advice` on the skipped parents with: *"AI advice not generated for this parent procedure because its individual statements are being analyzed separately. See the AI Advice on each Statement (parent ...) row."* — so the column shows an explanation instead of being blank.

`@AI = 2` users now see a full prompt on every row — including parent procedures, which is legitimately useful when the user wants the model to reason about a procedure as a whole rather than statement-by-statement. `@AI = 1` users continue to pay for AI calls only on the statement rows, not on their parent procs.

## Test results

Run against my local plan cache (no plan cache reset) before and after the fix:

| Test | Before | After |
|---|---|---|
| `@AI = 2`: rows with NULL `ai_prompt` | 7 of 59 | **0 of 59** |
| `@AI = 2`: parent-with-kids rows with a full prompt | 0 of 7 | **7 of 7** (15K-50K chars each) |
| Rendered output: empty `<ai_prompt />` elements | several | **0** |
| `@AI = 1`: parent-with-kids rows that triggered an API call | 7 of 7 | **0 of 7** |
| `@AI = 1`: parent-with-kids rows with the stub `ai_advice` | 0 of 7 | **7 of 7** |
| `@AI = 1`: non-parent rows processed by the cursor | 32 | **32** (unchanged) |

## Test plan

- [ ] Install `sp_BlitzCache.sql` from this branch on a server whose plan cache contains stored procedures with multiple statements.
- [ ] `EXEC dbo.sp_BlitzCache @AI = 2, @Top = 50;` — verify the **AI Prompt** column has content on every row, including `Procedure or Function:` rows whose statements are also present.
- [ ] Inspect the same result in SSMS — verify no row renders as `<ai_prompt />`.
- [ ] If you have AI credentials configured, run `EXEC dbo.sp_BlitzCache @AI = 1, @Top = 50;` and verify:
  - Parent-with-statements rows show the new stub message in **AI Advice** rather than a real AI response.
  - Non-parent rows still receive real AI advice as before.
  - The total number of API calls equals the number of non-parent rows (no extra calls on parents).

### DIFF
--- a/sp_BlitzCache.sql
+++ b/sp_BlitzCache.sql
@@ -5325,7 +5325,38 @@ XML Execution Plan:
 Thank you.'
     FROM ##BlitzCacheProcs p
     WHERE p.SPID = @@SPID
-    AND NOT (p.QueryType LIKE 'Procedure or Function:%'     /* This and the below exists query makes sure that we don't get advice for parent procs, only their statements, if the statements are in our result set. */
+    /* Generate a prompt for every row, including parent procedures whose
+       statements are also in the result set. When @AI = 1 we still avoid
+       calling the API on those parents (see the cursor below), but @AI = 2
+       users get a full prompt on every row so they can paste it elsewhere. */
+    OPTION (RECOMPILE);
+
+    /* If both query text and query plan are null, override with a simple message - no metrics or system prompt needed */
+    UPDATE ##BlitzCacheProcs
+    SET ai_prompt = N'Prompt not generated because we can''t find the query text or query plan.'
+    WHERE SPID = @@SPID
+    AND QueryText IS NULL
+    AND QueryPlan IS NULL
+    OPTION (RECOMPILE);
+
+    IF @Debug = 2
+        SELECT 'After setting up ai_prompt, before calling AI' AS ai_stage, SqlHandle, QueryHash, PlanHandle, QueryPlan, ai_prompt, ai_advice, ai_raw_response
+            FROM ##BlitzCacheProcs
+            WHERE SPID = @@SPID;
+        
+    IF @AI = 1
+    BEGIN
+        RAISERROR('Calling AI endpoint for query plan analysis - starting loop', 0, 1) WITH NOWAIT;
+
+        /* Parent procedures whose statements are also in the result set are
+           skipped by the AI cursor below (each statement is analyzed
+           individually). Stamp a clear ai_advice on those parent rows so the
+           column shows an explanation instead of being blank in the output. */
+        UPDATE p
+        SET p.ai_advice = N'AI advice not generated for this parent procedure because its individual statements are being analyzed separately. See the AI Advice on each Statement (parent ...) row.'
+        FROM ##BlitzCacheProcs p
+        WHERE p.SPID = @@SPID
+        AND p.QueryType LIKE 'Procedure or Function:%'
         AND EXISTS
         (
             SELECT 1
@@ -5350,25 +5381,7 @@ Thank you.'
                           - (LEN('Statement (parent ') + 1)
                     )))
         )
-    )
-    OPTION (RECOMPILE);
-
-    /* If both query text and query plan are null, override with a simple message - no metrics or system prompt needed */
-    UPDATE ##BlitzCacheProcs
-    SET ai_prompt = N'Prompt not generated because we can''t find the query text or query plan.'
-    WHERE SPID = @@SPID
-    AND QueryText IS NULL
-    AND QueryPlan IS NULL
-    OPTION (RECOMPILE);
-
-    IF @Debug = 2
-        SELECT 'After setting up ai_prompt, before calling AI' AS ai_stage, SqlHandle, QueryHash, PlanHandle, QueryPlan, ai_prompt, ai_advice, ai_raw_response
-            FROM ##BlitzCacheProcs
-            WHERE SPID = @@SPID;
-        
-    IF @AI = 1
-    BEGIN
-        RAISERROR('Calling AI endpoint for query plan analysis - starting loop', 0, 1) WITH NOWAIT;
+        OPTION (RECOMPILE);
 
         DECLARE @CurrentSqlHandle VARBINARY(64);
         DECLARE @CurrentQueryHash BINARY(8);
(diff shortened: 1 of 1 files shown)


BrentOzar:
Addressed Copilot's review note in 3843f4e — extracted the parent-with-kids detection (the SUBSTRING/CHARINDEX QueryType parse) into a one-shot `#parents_with_kids` temp table at the top of the `IF @AI = 1` block. The stub `UPDATE` now `INNER JOIN`s to it and the cursor uses `NOT EXISTS` against it, so the predicate lives in one place. Re-verified against the local plan cache: 0 NULL `ai_prompt` and 0 redundant API calls on parent-with-kids rows.

PR #4012, opened 2026-05-28
Added Emergency Mode to sp_blitzfirst, like what we have for sp_kill

### LINKED ISSUE #4011: sp_BlitzFirst: add option to skip statistics check
I'm always frustrated when I run `sp_BlitzFirst` during an emergency on a large database and it takes over a minute. Without fail, it will be waiting on CheckID 44, which generates SQL like this.
```sql
INSERT INTO #UpdatedStats(HowToStopIt, RowsForSorting)
    SELECT HowToStopIt = 
                QUOTENAME(DB_NAME()) + N'.' +
                QUOTENAME(SCHEMA_NAME(obj.schema_id)) + N'.' +
                QUOTENAME(obj.name) +
                N' statistic ' + QUOTENAME(stat.name) + 
                N' was updated on ' + CONVERT(NVARCHAR(50), sp.last_updated, 121) + N',' + 
                N' had ' + CAST(sp.rows AS NVARCHAR(50)) + N' rows, with ' +
                CAST(sp.rows_sampled AS NVARCHAR(50)) + N' rows sampled,' +  
                N' producing ' + CAST(sp.steps AS NVARCHAR(50)) + N' steps in the histogram.',
        sp.rows
    FROM sys.objects AS obj WITH (NOLOCK)
    INNER JOIN sys.stats AS stat WITH (NOLOCK) ON stat.object_id = obj.object_id  
    CROSS APPLY sys.dm_db_stats_properties(stat.object_id, stat.stats_id) AS sp  
    WHERE sp.last_updated > DATEADD(MI, -15, GETDATE())
    AND obj.is_ms_shipped = 0
    AND '[?]' <> '[tempdb]';
```
When I've waited for this check to finish, the information is often useless. If I have a large number of tables, then it's expected that plenty of them will have had statistics updates recently. I understand that this check is really there to warn of parameter sniffing emergencies, but the noise defeats the signal on a big database with a lot of tables and even the signal isn't worth the wait.

The worst I've ever seen was 7 minutes just for this check on a pretty much idle database with only 2,500 statistics. I don't know what these DMVs are doing internally and I don't trust what `SET STATISTICS IO ON` says about them. It's as if there's something hard about finding the page with the statistics properties on it. `sp_WhoIsActive` can report millions of reads in these cases with (I think) about 100,000 physical reads.

I find that for the above query to be tremendously slow, you need all three of these to be true:

1. The statistics data is not already in RAM.
2. You have a lot of statistics. Over 10,000 seems to do it, provided that your tables aren't so small that a few big reads can grab all of their data.
3. Your I/O latency is bad. I'm talking about the kinds of systems where 50 ms `PAGEIOLATCH_SH` waits are easy to find. You will struggle to replicate this on an SSD-equipped laptop unless you subject it to a cruel Governor (every demo I've made for that felt fake).

I'm not the first to take issue with this statistics check. We already have safeguards because it's [provably slow if you have many databases](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2439) and [fights with index creation](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/issues/2548).

Just add an off switch, `@CheckStatistics TINYINT = 1`. If not 1, then skip CheckID 44 which is the one and only statistics check. We can add it on [the same line as the other safeguards](https://github.com/BrentOzarULTD/SQL-Server-First-Responder-Kit/blob/dev/sp_BlitzFirst.sql#L2528).

We could carry on with the current model and keep on adding safeguards until we think that we've caught them all. I don't think that will work. Somebody will always find a new way to break it. For example, I can [steal this idea](https://learn.microsoft.com/en-us/answers/questions/175497/dynamic-management-objects-dm-db-stats-properties) and make it go extremely slow even on an un-throttled commodity SSD by making lots of nearly empty tables with incremental statistics. I definitely wouldn't have thought of that ahead of time. I suspect that bad columnstore maintenance is another way to make it awful.

Another alternative is tuning the query, but I have no ideas for it. The nasty part is always doing the `sys.dm_db_stats_properties` read, which is always near-last in the execution plans I've seen. I don't know how to make those reads faster. We've already done almost all of the filtering that we can at that point and there isn't any way to force parallelism into this.

We could maybe use the `STATS_DATE` function instead of `sys.dm_db_stats_properties` but I have no reason to believe that will be better and it will definitely return less information. 

**Do you want to build this feature yourself?**
If you want to build it, we'll leave this issue open and discuss it, giving our feedback about ideas on how to build it. If you DON'T want to build it, no problem, we can give it a shot with AI-assisted coding to see if we can knock it out quickly.
Yes. I was hoping to do it tonight, but I might have spent too long writing up the issue. Maybe tomorrow night.

ReeceGoding:
Added Emergency Mode to `sp_blitzfirst`, like what we have for `sp_kill` but on by default rather than off.

Currently only has two states: either equal to 1 (the default) and therefore on, or not and therefore off. All that turning this on does is skip the statistics check, check ID 44.

Closes #4011.

## Screenshot
image

## Test script

```sql
DROP TABLE IF EXISTS Whatever;
SELECT * INTO Whatever FROM sys.messages;
CREATE STATISTICS Whatever ON Whatever(severity);
UPDATE STATISTICS Whatever;

SELECT 'Emergency Mode off, statistics check fires.';
EXEC sp_BlitzFirst @EmergencyMode = 0;
SELECT 'Emergency Mode on, statistics check does not fire.';
EXEC sp_BlitzFirst @EmergencyMode = 1;
SELECT 'Emergency Mode set to nonsense, statistics check does not fire.';
EXEC sp_BlitzFirst @EmergencyMode = 6;
SELECT 'Emergency Mode set to NULL, statistics check does not fire.';
EXEC sp_BlitzFirst @EmergencyMode = NULL;
SELECT 'Emergency Mode not mentioned, statistics check does not fire.';
EXEC sp_BlitzFirst;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ Common sp_BlitzFirst parameters include:
 * @Seconds = 5 by default. You can specify longer samples if you want to track stats during a load test or demo, for example.
 * @ShowSleepingSPIDs = 0 by default. When set to 1, shows long-running sleeping queries that might be blocking others.
 * @ExpertMode = 0 by default. When set to 1, it calls sp_BlitzWho when it starts (to show you what queries are running right now), plus outputs additional result sets for wait stats, Perfmon counters, and file stats during the sample, then finishes with one final execution of sp_BlitzWho to show you what was running at the end of the sample. When set to 2, it does the same as 1, but skips the calls to sp_BlitzWho.
+* @EmergencyMode = 1 by default. When set to 0, it runs potentially slow checks that are skipped otherwise.
 
 ### Logging sp_BlitzFirst to Tables
 
--- a/sp_BlitzFirst.sql
+++ b/sp_BlitzFirst.sql
@@ -24,6 +24,7 @@ ALTER PROCEDURE [dbo].[sp_BlitzFirst]
     @OutputXMLasNVARCHAR TINYINT = 0 ,
     @FilterPlansByDatabase VARCHAR(MAX) = NULL ,
     @CheckProcedureCache TINYINT = 0 ,
+    @EmergencyMode TINYINT = 1 ,
     @CheckServerInfo TINYINT = 1 ,
     @FileLatencyThresholdMS INT = 100 ,
     @SinceStartup TINYINT = 0 ,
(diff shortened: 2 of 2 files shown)


BrentOzar (approved):

PR #4014, opened 2026-06-02
sp_BlitzLock: add @SkipExecutionPlans to skip plans and return only findings (#4013)

### LINKED ISSUE #4013: sp_BlitzLock: add @skip_execution_plans to skip plans and go straight to findings rollup
## Summary

Add a `@skip_execution_plans bit = 0` parameter to `sp_BlitzLock` that lets callers bypass the execution plans result set and jump straight to the findings rollup.

## Motivation

When triaging deadlocks, the execution plans result set is often the slowest part to produce (it gathers `#available_plans` / `#dm_exec_query_stats` and calls `sys.dm_exec_text_query_plan`) and is frequently not what the user wants — they just want the findings rollup. A flag to skip it speeds up the common "just show me the findings" case and avoids the plan-fetching work entirely.

## Current structure

`sp_BlitzLock` already returns the two result sets back-to-back:

- Execution plans result set: the `#available_plans` SELECT (~lines 4492-4576)
- Findings rollup: the `#deadlock_findings` SELECT (~lines 4583-4593)

So the change is well-scoped: gate the plan-gathering work and the plan result set behind the new flag, then fall through to the existing findings rollup.

## Proposed behavior

- `@skip_execution_plans = 0` (default): unchanged behavior.
- `@skip_execution_plans = 1`: skip building `#available_plans` / `#dm_exec_query_stats`, skip the `sys.dm_exec_text_query_plan` calls, and skip the execution plans result set; return only the findings rollup.

## Prior art

The same feature was just added to `sp_HumanEventsBlockViewer` in the DarlingData repo, which has the identical execution-plans-then-rollup shape. The approach (a guard that skips straight to the rollup, plus `@help` entries for description/valid inputs/defaults) applies here.

## Notes

- Findings checks read from the `#deadlock_*` temp tables, not from `#available_plans`, so skipping the plan section does not affect the rollup.
- Add the parameter to the `@help = 1` output (description, valid inputs, default).

erikdarlingdata:
## Summary

Adds a `@SkipExecutionPlans bit = 0` parameter to `sp_BlitzLock` that lets callers bypass the execution plans result set and return only the findings rollup.

Fixes #4013.

## What changed

When `@SkipExecutionPlans = 1`, the client-output path:

- skips building `#available_plans` / `#dm_exec_query_stats`,
- skips the `sys.dm_exec_text_query_plan` calls, and
- skips the execution plans result set,

then falls through to the existing deadlock findings rollup. It uses a `GOTO ReturnDeadlockFindings` guard placed right after the deadlock results are returned to the client, with the label immediately before the findings rollup — so the only thing skipped is the plan gathering and the plan result set.

Default behavior (`@SkipExecutionPlans = 0`) is unchanged.

## Why

The execution plans result set is often the slowest part to produce and is frequently not what the user wants when triaging deadlocks — they just want the findings. This flag skips that work and result set entirely.

## Notes

- The findings rollup reads from the `#deadlock_*` temp tables, not from `#available_plans`, so skipping the plan section does not affect the rollup.
- The `@Debug` dumps for `#available_plans` and `#dm_exec_query_stats` are already guarded by `OBJECT_ID(...) IS NOT NULL`, so skipping their creation is safe under `@Debug = 1`.
- Documented in `@Help` and surfaced in the `@Debug` parameter dump.

## Testing

Installed and compiled clean on SQL Server 2022 (16.0.4255.1). Verified:

- `@SkipExecutionPlans = 1`: returns deadlock results then findings rollup, with no "Getting available execution plans" step.
- `@SkipExecutionPlans = 0` (default): still runs the execution plans gathering and result set as before.

### DIFF
--- a/sp_BlitzLock.sql
+++ b/sp_BlitzLock.sql
@@ -32,7 +32,8 @@ ALTER PROCEDURE
     @OutputDatabaseName sysname = NULL,
     @OutputSchemaName sysname = N'dbo',      /*ditto as below*/
     @OutputTableName sysname = N'BlitzLock', /*put a standard here no need to check later in the script*/
-    @ExportToExcel bit = 0
+    @ExportToExcel bit = 0,
+    @SkipExecutionPlans bit = 0 /*skip the execution plans result set and return only the findings rollup*/
 )
 WITH RECOMPILE
 AS
@@ -103,7 +104,10 @@ BEGIN
        
         @TargetTimestampColumnName: The name of the datetime column for filtering by date range (optional)
 
-   
+        /*Output options*/
+        @SkipExecutionPlans: Set to 1 to skip the execution plans result set and return only the findings rollup
+
+
     To learn more, visit http://FirstResponderKit.org where you can download new
     versions for free, watch training videos on how it works, get more info on
     the findings, contribute your own code, and more.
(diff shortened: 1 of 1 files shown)


BrentOzar:
I love this! Great idea. Thanks for the pull request, merging into the dev branch, will be in the next release with credit to you in the release notes.

PR #4020, opened 2026-06-06
Fix DIFF backup silently skipped when filename has no numeric split-file suffix

vanterx:
## Problem

When \sp_DatabaseRestore\ is called with \@RestoreDiff = 1\, the DIFF restore is silently skipped if the DIFF backup file does not have a trailing numeric split-file index (e.g. \_1.bak\).

Root cause: \PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) )\ incorrectly matches digit pairs inside the datetime portion of non-split filenames (e.g. \_190000\ in \...DIFF_20260605_190000.bak\), producing garbled datetime strings like \ODDIFF20260605\ instead of \20260605190000\.

## Fix

Replaced all 13 occurrences of the broken pattern with a \CASE\ expression:
\\\sql
CASE WHEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) <= 7
     THEN PATINDEX( '%[_]%', REVERSE( BackupFile ) )  -- strip _NN.bak/_NN.trn
     ELSE CHARINDEX( '.', REVERSE( BackupFile ) )       -- strip .bak/.trn only
END
\\\
If the last underscore is within 7 chars from the end (indicating a 1-2 digit split suffix), strip from there. Otherwise fall back to stripping only the file extension. Backward-compatible: split backups produce identical results.

## Validation

| Filename | Last \_\ from end | Result |
|---|---|---|
| \...173813_10.bak\ | 7 → ≤ 7 → strip \_10.bak\ | \20260530173813\ |
| \...173813_9.bak\ | 6 → ≤ 7 → strip \_9.bak\ | \20260530173813\ |
| \...190000.bak\ (non-split) | 11 → > 7 → strip \.bak\ | \20260605190000\ |
| \...000100_01.trn\ | 7 → ≤ 7 → strip \_01.trn\ | \20260606000100\ |
| \...000100.trn\ (non-split) | 11 → > 7 → strip \.trn\ | \20260606000100\ |

### DIFF
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -753,16 +753,7 @@ BEGIN
 			FROM @FileList
 			WHERE BackupFile LIKE N'%[_][0-9][0-9].' + @FileExtensionBak
 			AND	BackupFile LIKE N'%' + @Database + N'%'
-			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 18 ), '_', '' ) > @StopAt);
 			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, CASE WHEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) <= 7 THEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) ELSE CHARINDEX( '.', REVERSE( BackupFile ) ) END ), '' ), 18 ), '_', '' ) > @StopAt);
-
-			DELETE
-			FROM @FileList
-			WHERE BackupFile LIKE N'%.bak' /*delete backups that aren't striped and new "enough"*/
-			AND BackupFile NOT LIKE N'%[_][0-9].bak' /*skip remaining striped backups since those were deleted above*/
-			AND BackupFile NOT LIKE N'%[_][0-9][0-9].bak'
-			AND	BackupFile LIKE N'%' + @Database + N'%'
-			AND (REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 16 ), '_', '' ) > @StopAt);
 		END;
 
     -- Find latest full backup
--- a/sp_DatabaseRestore.sql
+++ b/sp_DatabaseRestore.sql
@@ -747,13 +747,22 @@ BEGIN
 			FROM @FileList
 			WHERE BackupFile LIKE N'%[_][0-9].' + @FileExtensionBak
 			AND	BackupFile LIKE N'%' + @Database + N'%'
-			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 16 ), '_', '' ) > @StopAt);
+			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, CASE WHEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) <= 7 THEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) ELSE CHARINDEX( '.', REVERSE( BackupFile ) ) END ), '' ), 16 ), '_', '' ) > @StopAt);
 
 			DELETE
 			FROM @FileList
 			WHERE BackupFile LIKE N'%[_][0-9][0-9].' + @FileExtensionBak
 			AND	BackupFile LIKE N'%' + @Database + N'%'
 			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 18 ), '_', '' ) > @StopAt);
+			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, CASE WHEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) <= 7 THEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) ELSE CHARINDEX( '.', REVERSE( BackupFile ) ) END ), '' ), 18 ), '_', '' ) > @StopAt);
+
+			DELETE
+			FROM @FileList
+			WHERE BackupFile LIKE N'%.bak' /*delete backups that aren't striped and new "enough"*/
+			AND BackupFile NOT LIKE N'%[_][0-9].bak' /*skip remaining striped backups since those were deleted above*/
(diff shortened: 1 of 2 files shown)


sp_DatabaseRestore.sql
```diff
@@ -747,13 +747,22 @@ BEGIN
...
 
 			DELETE
 			FROM @FileList
 			WHERE BackupFile LIKE N'%[_][0-9][0-9].' + @FileExtensionBak
 			AND	BackupFile LIKE N'%' + @Database + N'%'
 			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 18 ), '_', '' ) > @StopAt);
+			AND	(REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, CASE WHEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) <= 7 THEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) ELSE CHARINDEX( '.', REVERSE( BackupFile ) ) END ), '' ), 18 ), '_', '' ) > @StopAt);
+
+			DELETE
+			FROM @FileList
```
BrentOzar: 🛑 **SQL syntax error — old line wasn't removed.**

The old `PATINDEX('%_[0-9][0-9]%' ...)` line on 759 is still here, and the new `CASE WHEN` line is appended below it. The first line ends the `DELETE` statement with `;`, so line 760's leading `AND ...` no longer belongs to a `WHERE` clause and will fail to parse.

The companion replacement in `Install-All-Scripts.sql` (line ~35228) correctly removed the old line — please do the same here. Line 759 should be deleted, leaving only the new `CASE WHEN` expression on line 760.

---
_Generated by [Claude Code](https://claude.ai/code)_

sp_DatabaseRestore.sql
```diff
@@ -747,13 +747,22 @@ BEGIN
...
+			WHERE BackupFile LIKE N'%.bak' /*delete backups that aren't striped and new "enough"*/
+			AND BackupFile NOT LIKE N'%[_][0-9].bak' /*skip remaining striped backups since those were deleted above*/
+			AND BackupFile NOT LIKE N'%[_][0-9][0-9].bak'
+			AND	BackupFile LIKE N'%' + @Database + N'%'
+			AND (REPLACE( RIGHT( REPLACE( BackupFile, RIGHT( BackupFile, PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) ) ), '' ), 16 ), '_', '' ) > @StopAt);
 		END;
 
     -- Find latest full backup
```
BrentOzar: 🛑 **This new DELETE block reintroduces the very bug this PR is fixing, and hardcodes `.bak`.**

Two problems:

1. **Uses the broken pattern.** Line 768 still uses `PATINDEX( '%_[0-9][0-9]%', REVERSE( BackupFile ) )` — the exact expression the rest of the PR is replacing. Because the `WHERE` filters for non-striped files (`NOT LIKE N'%[_][0-9].bak'`), this block targets *precisely* the failure case described in the PR (e.g. `..._190000.bak`), so the bug will silently fire here. It needs the same `CASE WHEN PATINDEX( '%[_]%', REVERSE( BackupFile ) ) <= 7 THEN ... ELSE CHARINDEX( '.', REVERSE( BackupFile ) ) END` treatment.

2. **Hardcodes `.bak` instead of `@FileExtensionBak`.** The two DELETE blocks above this one use `+ @FileExtensionBak` so callers can override the extension. This block uses `N'%.bak'`, `N'%[_][0-9].bak'`, `N'%[_][0-9][0-9].bak'` — for any user who sets `@FileExtensionBak` to something other than `bak`, this block is a no-op.

Also — is this new DELETE block actually part of the documented fix? The PR description and validation table describe only the 13 expression replacements, and the corresponding hunk in `Install-All-Scripts.sql` doesn't include this block. If it was added by accident, the cleanest fix is to drop it. If it's intentional, please describe what scenario it covers in the PR body, and fix items (1) and (2) above.

---
_Generated by [Claude Code](https://claude.ai/code)_

BrentOzar (changes requested):
Thanks for tracking this down — the root-cause analysis is spot on, and the `CASE WHEN PATINDEX( '%[_]%', REVERSE(...) ) <= 7 THEN ... ELSE CHARINDEX('.', REVERSE(...)) END` replacement is the right fix. Validation table checks out.

A few things to address before this is ready:

### Please don't modify `Install-All-Scripts.sql`

`Install-All-Scripts.sql` is generated by our build process from the individual script files — it's not hand-edited. Any changes there get blown away on the next build. Please revert your edits to `Install-All-Scripts.sql` and keep this PR scoped to `sp_DatabaseRestore.sql`. The build will pick up the fix from the source file.

(Side benefit: this also avoids the divergence I flagged below, where the new DELETE block exists in `sp_DatabaseRestore.sql` but not in `Install-All-Scripts.sql`.)

### Issues in `sp_DatabaseRestore.sql`

See the two inline comments:

1. **Syntax error in the second DELETE block (~lines 759-760)** — the old broken `PATINDEX('%_[0-9][0-9]%' ...)` line wasn't removed, just appended to. The first line ends the statement with `;`, so the new line becomes a stray `AND ...` that won't parse.
2. **The new "non-striped backup" DELETE block (~lines 762-768)** uses the same broken `PATINDEX('%_[0-9][0-9]%' ...)` pattern this PR exists to fix, and hardcodes `.bak` instead of using `@FileExtensionBak`. Also unclear whether this block is intended to be part of this fix at all.

### Looks good

The other 12 expression replacements in `sp_DatabaseRestore.sql` are clean substitutions and preserve the existing `<= 7` split-suffix heuristic. Nice work on the diagnosis — once the above is sorted this is a solid fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

vanterx:
Thanks for the detailed review! Addressed everything in commit 668f5f2:

1. **Syntax error fixed** — removed the stale `PATINDEX('%_[0-9][0-9]%', ...)` line that was left alongside the new `CASE WHEN` line in the `_[0-9][0-9].bak` `@StopAt` DELETE block. Only the corrected expression remains.
2. **Non-striped DELETE block dropped** — the accidental block (lines 762–768) is gone. It wasn't among the 13 locations listed in the issue, reintroduced the broken pattern, and hardcoded `.bak` instead of `@FileExtensionBak`.
3. **`Install-All-Scripts.sql` reverted** — the file is back to its pre-PR state; the build process will pick up the fix from `sp_DatabaseRestore.sql` after merge.

BrentOzar (approved):
Verified commit `668f5f2` against the three asks — all addressed cleanly:

- ✅ Second DELETE block now has exactly one `AND (REPLACE(...CASE WHEN...)) > @StopAt);` clause; the orphan/duplicate `AND` is gone.
- ✅ The accidental non-striped DELETE block is dropped.
- ✅ `Install-All-Scripts.sql` reverted — PR is now scoped to `sp_DatabaseRestore.sql` only, 13/13 clean substitutions.

Nice diagnosis and quick turnaround on the cleanup. LGTM 👍

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4027, opened 2026-07-02
Trigger sp_kill smoke test

BrentOzar:
## Summary

Adds a harmless whitespace-only tweak in `sp_kill.sql` so the new SQL Server smoke-test workflow has a root-level `sp_*.sql` change to test.

## Expected result

The newly merged workflow should pick up `sp_kill.sql`, install it individually, and report the current standalone compile issue near `tempdb`.

## Validation

This PR is intentionally meant to validate the GitHub Actions workflow behavior rather than change product behavior.

### DIFF
--- a/sp_kill.sql
+++ b/sp_kill.sql
@@ -32,7 +32,7 @@ BEGIN
 	SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
 
 	/*
-	sp_kill from http://FirstResponderKit.org
+	sp_kill  from http://FirstResponderKit.org
 
 	This script helps you kill queries during a performance emergency.
 


BrentOzar:
The new SQL Server smoke-test workflow did run on this PR and passed.

The important bit: this confirms the workflow is active on `main` and that a root-level `sp_*.sql` change (`sp_kill.sql`) is picked up by the PR workflow. My earlier local `Incorrect syntax near 'tempdb'` result came from the Homebrew Go-based `sqlcmd`; GitHub Actions uses Microsoft `mssql-tools18` `sqlcmd`, and the script compiles successfully there.

Issue #4033, opened 2026-07-10
sp_ineachdb add option to sort by database name instead of id for execution order


jbelina:
**Is your feature request related to a problem? Please describe.**
I am frustrated by the default order that sp_ineachdb commands are executed. The current order is by database_id which in my environment seems very random. This behavior is less of a problem if I'm inserting records into a temporary table since I can sort my output.  But for scripts that are run on an adhoc basis, I would prefer to have the execution be in database name order.

**Describe the solution you'd like**
I would like the execution order used by sp_ineachdb be in database_name order instead of by database_id.

**Describe alternatives you've considered**
I considered adding the original order when processing a @database_list, but that ended up being impractical based on the way database names are parsed.

**Do you want to build this feature yourself?**
Yes, please. I have the @database_order feature coded and ready for review.  I also have two fun variants below but did not include them in the actual commit.

```ORDER BY
  CASE @database_order
  WHEN N'NAME' THEN name
  WHEN N'RANDOM' THEN TRY_CAST(NEWID() AS NVARCHAR(36)) -- Just for fun #1 Random order
  WHEN N'EMAN' THEN REVERSE(name) -- Just for fun #2 EMAN (NAME backwards)
  ELSE NULL END ASC
  , id ASC;

BrentOzar:
Thanks for the pull request! Makes sense, looks good, merging into dev, will be in the next release with credit to you in the release notes.

PR #4034, opened 2026-07-10
    Add optional database execution order to sp_ineachdb.

jbelina:
I had fun with other sort options, but RANDOM and EMAN didn't seem practical.

```
ORDER BY
          CASE @database_order
          WHEN N'NAME' THEN name
          WHEN N'RANDOM' THEN TRY_CAST(NEWID() AS NVARCHAR(36)) -- Just for fun #1 Random order
          WHEN N'EMAN' THEN REVERSE(name) -- Just for fun #2 EMAN (NAME backwards)
          ELSE NULL END ASC
          , id ASC;
```

#4033


BrentOzar (approved):
Thanks for the pull request! Makes sense, looks good, merging into dev, will be in the next release with credit to you in the release notes.

PR #4039, opened 2026-07-27
sp_ineachdb.sql: default to include DBs regardless of read_only

aprokosch:
The current default of @is_read_only = 0 excludes readonly databases from sp_ineachdb output by default. This is counterintuitive and apparently undocumented. The suggested new default of NULL includes both readonly and read-write databases, allowing the user to specifically override the behavior in either direction (=0 or =1) if desired.

### DIFF
--- a/sp_ineachdb.sql
+++ b/sp_ineachdb.sql
@@ -20,7 +20,7 @@ ALTER PROCEDURE [dbo].[sp_ineachdb]
   @recovery_model_desc  nvarchar(120)  = NULL,
   @compatibility_level  tinyint        = NULL,
   @state_desc           nvarchar(120)  = N'ONLINE',
-  @is_read_only         bit = 0,
+  @is_read_only         bit = NULL,
   @is_auto_close_on     bit = NULL,
   @is_auto_shrink_on    bit = NULL,
   @is_broker_enabled    bit = NULL,


BrentOzar:
Howdy! Lemme try to rephrase what you're saying to make sure I understand:

The @is_read_only parameter has 3 options:
* = 1 means only read-only databases
* = 0 means no read-only databases (like when you need to make writeable changes) - and this is the default
* = null means all databases

And because that's not documented, you want to change the default to null.

If I captured that correctly, let's not change the code: let's document the switch. Otherwise, changing the default settings on code that's out there in the wild, used on production servers, is kinda dangerous. People will suddenly have different behavior if they actually wanted/expected their stuff to run on non-read-only databases, and we change the code behind them.

I'm going to close this pull request, create an issue for the documentation change, and then make that change. Going forward, whenever you spot something you'd like to change, just check out the readme file in the repo first, which links to info on how to create issues, how we work in here (issues first w/discussions before pull requests), etc. I just wanna make sure your time isn't wasted.

Thanks for the report though!

aprokosch:
Thanks Brent! Your solution addresses my concern (=previous surprise) very well.

Quick note if you use the end of your response in other contexts -- I actually had read through the README on this repo previously. Technically, I think you wanted me to have reviewed the CONTRIBUTING file before going off and submitting a PR. I've now done that, and will follow that procedure in future.

Thanks for all the tools,
A

PR #4043, opened 2026-08-06
Docs: document sp_ineachdb @is_read_only, add Other Utilities TOC section

### LINKED ISSUE #4042: sp_ineachdb: document @is_read_only setting
Right now, the @is_read_only parameter has 3 options:
* = 1 means only read-only databases
* = 0 means no read-only databases (like when you need to make writeable changes) - and this is the default
* = null means all databases

And because that's not documented, people can expect the defaults will run in all databases, when it's actually skipping the read-only ones.

On the readme.md for the repo, in the sp_ineachdb section, add this parameter to the parameters section.

Also, while we're in there, in the Navigation table of contents at the top, above "Parameters Common to Many of the Stored Procedures", add a section for "Other Utilities", and in there put a hyperlink to the sp_ineachdb anchor. Add sp_BlitzUpdate documentation as well.

BrentOzar:
Closes #4042.

README-only change, no T-SQL touched.

### 1. Document `@is_read_only` in the sp_ineachdb parameters section

It was buried in a lumped "filter by database property" bullet with no mention that its default silently skips databases. Now it has its own breakdown of all three values, leading with the surprise:

* `0` (the default) - only databases that are *not* read-only
* `1` - only read-only databases
* `NULL` - all databases

While verifying that against the code, `@state_desc` turned out to have the identical trap — it defaults to `N'ONLINE'`, so offline, restoring, and recovering databases get skipped too, and it was sitting in that same lumped bullet. Documented it the same way. Confirmed against the `DELETE` at sp_ineachdb.sql:295 that passing `NULL` does disable the filter.

The remaining property filters all default to `NULL`, so the bullet now says so explicitly.

### 2. "Other Utilities" section in the Navigation table of contents

Added above "Parameters Common to Many of the Stored Procedures", with links to the sp_ineachdb and sp_BlitzUpdate anchors.

### 3. sp_BlitzUpdate documentation

New section after sp_ineachdb, sourced from the proc's `@Help` text and the code:

* What it does and the one-command usage
* Installation note that it lives in `OptionalScripts` and is not in `Install-All-Scripts.sql` or `Install-Azure.sql`
* Requirements: SQL Server 2025 / Azure SQL DB for `sp_invoke_external_rest_endpoint`, the `external rest endpoint enabled` sp_configure toggle, outbound HTTPS to api.github.com, the Azure outbound-REST allowlist and its `Msg 31612`, and the compat-level-130 OPENJSON dependency
* Parameter table, examples, and the trust model — every call EXECs whatever GitHub serves at the requested ref, so pin `@Branch` to a release tag

### Note for the reviewer

sp_BlitzPlanCompare is also an OptionalScript but is listed under "Performance Tuning" in the TOC. I left it there — the TOC grouping reads as by-purpose rather than by-install-bundle. Easy to move if you'd rather group the optional scripts together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,9 @@ Navigation
  - Backups and Restores:
    - [sp_BlitzBackups: How Much Data Could You Lose](#sp_blitzbackups-how-much-data-could-you-lose)  
    - [sp_DatabaseRestore: Easier Multi-File Restores](#sp_databaserestore-easier-multi-file-restores)  
+ - Other Utilities:
+   - [sp_ineachdb: Run a Command in Each Database](#sp_ineachdb-run-a-command-in-each-database)
+   - [sp_BlitzUpdate: Update the First Responder Kit from GitHub](#sp_blitzupdate-update-the-first-responder-kit-from-github)
  - [Parameters Common to Many of the Stored Procedures](#parameters-common-to-many-of-the-stored-procedures)
  - [License MIT](#license)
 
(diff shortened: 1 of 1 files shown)


BrentOzar (commented):
Looks good, merging to dev.

PR #4089, opened 2026-09-12
Fix Uninstall.sql targeting the wrong procedure schema

### LINKED ISSUE #4066: Uninstall.sql: schema-blind matching can remove master procedures from a user-database uninstall
## Version and environment
Reproduced on dev `2e2ac07e` with SQL Server 2022 Linux in a disposable container.

## Problem
Both procedure-discovery queries match sys.procedures by name without filtering schema_id, then generate every DROP against dbo. This can remove a procedure from **master even with @allDatabases=0**.

When the current database contains dbo.sp_Blitz and custom.sp_Blitz, the script generates:
```sql
DROP PROCEDURE dbo.sp_Blitz;
DROP PROCEDURE dbo.sp_Blitz;
```
The first removes the current database's dbo.sp_Blitz. SQL Server's sp_ name resolution allows the second to remove master.dbo.sp_Blitz. The custom procedure remains. Both DROPs succeed, so there is no error indicating the unintended cross-database removal.

## Repro (disposable server only)
1. Install current root sp_Blitz.sql in master.
2. Create a throwaway user database and execute there:
```sql
CREATE SCHEMA custom;
GO
CREATE PROCEDURE custom.sp_Blitz AS SELECT 1;
GO
CREATE PROCEDURE dbo.sp_Blitz AS SELECT 2;
GO
SELECT DB_NAME() AS ContextDB;
SELECT name,schema_id FROM sys.procedures;
SELECT name FROM master.sys.procedures WHERE name='sp_Blitz';
```
3. Still in the user database, run unchanged root Uninstall.sql with defaults (@allDatabases=0, @printOnly=0).
4. Repeat the catalog queries. Actual: custom.sp_Blitz remains, but both the current dbo.sp_Blitz AND master.dbo.sp_Blitz are gone.

If only custom.sp_Blitz exists and master has no sp_Blitz, the same incorrect discovery produces error 3701 for nonexistent dbo.sp_Blitz.

## Expected / suggested fix
Uninstall only objects in the selected database and intended schema. Since the kit installers target dbo, filter discoveries to dbo and qualify emitted drops to prevent sp_ fallback to master. Apply the same rule to current-database and all-databases modes. Include before/after master checks in regression tests.

The reproduction was confined to a disposable local container, master.sp_Blitz was restored afterward, and temporary databases were removed. No remote installed procedures were touched. Fix PR will follow the full root-script review pass.

BrentOzar:
Uninstall.sql discovers procedures by name alone, which can generate duplicate dbo drops and accidentally fall through to the master system procedure. Restrict both current- and all-database discovery to dbo, matching the kit installers, and quote the actual schema and procedure names. Preserve unrelated custom-schema procedures.

The all-databases path also preserves full quoted database names and parameterizes raw names when building USE statements.

Closes #4066.

Validation: added a CI regression that executes the actual uninstall script in both modes after the smoke matrix on the disposable server. It verifies master preservation in current mode, custom/unrelated object preservation, dbo kit removal, quoted Unicode names, and a 128-character database name. The regression passes on SQL Server 2022 and fails on the prior PR head with 'Custom procedure was removed.' Shell syntax and git diff --check pass.

### DIFF
--- a/.github/scripts/build-uninstall-regression.py
+++ b/.github/scripts/build-uninstall-regression.py
@@ -1,0 +1,60 @@
+"""Generate a destructive uninstall regression for the disposable CI server only.
+
+Run after the smoke matrix: all-database mode intentionally removes the kit.
+The production script is embedded verbatim except for its allDatabases setting.
+"""
+from pathlib import Path
+
+source = (Path(__file__).resolve().parents[2] / 'Uninstall.sql').read_text()
+def literal(value):
+    return "N'" + value.replace("'", "''") + "'"
+def identifier(value):
+    return '[' + value.replace(']', ']]') + ']'
+
+names = ["FRKUninstall'雪]", 'FRKUninstall' + 'x' * 115 + ']']
+print('USE master; SET NOCOUNT ON;')
+print("DECLARE @DataPath nvarchar(4000) = CONVERT(nvarchar(4000), SERVERPROPERTY('InstanceDefaultDataPath'));")
+print("DECLARE @CreateDatabase nvarchar(max);")
+print("IF OBJECT_ID(N'dbo.sp_Blitz') IS NULL THROW 51000, 'Install the kit before testing uninstall.', 1;")
+for name in names:
+    print(f"IF DB_ID({literal(name)}) IS NOT NULL THROW 51000, 'Uninstall fixture already exists.', 1;")
+    if len(name) < 100:
+        print(f'EXEC({literal("CREATE DATABASE " + identifier(name))});')
+    else:
+        prefix = 'CREATE DATABASE ' + identifier(name) + " ON PRIMARY (NAME=N'FRKUninstallLong', FILENAME=N'"
+        middle = "FRKUninstallLong.mdf') LOG ON (NAME=N'FRKUninstallLong_log', FILENAME=N'"
+        suffix = "FRKUninstallLong_log.ldf');"
+        print(f"SET @CreateDatabase = {literal(prefix)} + REPLACE(@DataPath, '''', '''''') + {literal(middle)} + REPLACE(@DataPath, '''', '''''') + {literal(suffix)};")
+        print('EXEC(@CreateDatabase);')
+    setup = f'''USE {identifier(name)};
+EXEC(N'CREATE SCHEMA custom');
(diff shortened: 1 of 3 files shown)


BrentOzar:
Restricted both discovery paths to dbo, matching the kit installers, while retaining explicit schema/name quoting. Same-named custom-schema procedures are now preserved.

Added a CI regression generator that embeds the actual Uninstall.sql, changing only its allDatabases setting. It runs last on the disposable SQL Server because the all-database test intentionally removes the installed kit. It exercises current/all modes, verifies current mode preserves master.dbo.sp_Blitz, verifies custom procedures and unrelated objects survive, and tests quoted Unicode and 128-character database names (with short physical filenames).

The regression passes on SQL Server 2022 with this fix and fails against the prior PR head with 'Custom procedure was removed.' Shell syntax and git diff --check pass. Generated installers remain unchanged per CONTRIBUTING.md.

BrentOzar:
Addressed the caller/target collation finding with exact binary procedure-name matching in both paths. The fixture databases now use a case-sensitive collation and contain both dbo.sp_Blitz and unrelated dbo.SP_BLITZ; the latter must survive current/all modes. Master preservation is asserted using the saved object ID from master.sys.procedures, rather than OBJECT_ID name resolution.

Updated regression passes on SQL Server 2022 and fails against the previous PR head with 'Differently cased procedure was removed.' Fixture databases were removed, and git diff --check passes.

BrentOzar:
Changed the match to follow each target database's identifier semantics. Current mode uses DATABASE_DEFAULT; all-database discovery now executes in the target database context before applying DATABASE_DEFAULT. Thus a case-sensitive database preserves distinct dbo.SP_BLITZ, while a case-insensitive database removes the equivalent installed spelling.

Added a third, case-insensitive fixture with only dbo.SP_BLITZ and tested it in both current and all modes, alongside the two case-sensitive fixtures. All cases pass. The prior binary-match head fails the new CI-database assertion with 'dbo kit procedure remains.' Existing tests continue to prove custom-schema and distinct CS names survive.

BrentOzar:
Added the requested fail-fast generation guard in a58a676e. Before emitting any SQL, the generator now requires exactly one expected allDatabases declaration; unexpected formatting or duplicate declarations raise an error rather than silently testing current-database mode twice. Verified the generated script contains exactly two current-mode executions and one all-databases execution.

PR #4095, opened 2026-09-12
sp_BlitzCache: validate unsupported sort and query-filter combinations

### LINKED ISSUE #4083: sp_BlitzCache: explicit query filters bypass guards for unsupported sort metrics
Reproduced on unchanged dev `2e2ac07e3ea78f0d0156a4de457da888c87f78e2`, SQL Server 2022 Linux:

```sql
EXEC master.dbo.sp_BlitzCache @Top=1,
    @QueryFilter='procedures', @SortOrder='memory grant';
-- Error 207: Invalid column name 'max_grant_kb'.

EXEC master.dbo.sp_BlitzCache @Top=1,
    @QueryFilter='functions', @SortOrder='spills';
-- Error 207: Invalid column name 'max_spills'.
```

The procedure/function collection conditions exclude unsupported sorts for `@QueryFilter='all'`, but append an `OR LEFT(@QueryFilter,3)='pro'/'fun'` that bypasses those guards. The shared dynamic SQL then sorts those DMVs by columns they do not expose.

Validate these combinations before collecting plan-cache data, with a clear message directing users to statement metrics, or apply a documented supported fallback. Include the corresponding average/unused-grant variants when validating. Avoid silently presenting unrelated metrics as if the requested sort succeeded.

BrentOzar:
Explicit procedure/function filters bypass the normal exclusions for sorts whose columns their DMVs do not expose. For example, procedure memory-grant sorting fails with `Invalid column name 'max_grant_kb'`, and function spill sorting fails with `Invalid column name 'max_spills'`.

Reject unsupported combinations with an actionable message directing the caller to statement statistics or another sort. Cover memory-grant/average/unused and duplicate sorts for procedures/functions, plus spills/average spills for functions. Preserve procedure spill sorting and supported statement/all filters.

Closes #4083.

Validation: SQL Server 2022 tests verified the intended error messages for ten unsupported combinations and successful execution for six supported controls: procedure CPU/spills, function CPU, statement memory grants/spills, and all with memory grants. No invalid-column errors occurred. `git diff --check` passed.

### DIFF
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -141,6 +141,58 @@ EXEC dbo.sp_BlitzCache
      @OutputSchemaName   = 'dbo',
      @OutputTableName    = 'BlitzCache';
 
+--#STEP: sp_BlitzCache rejects unsupported filters before reanalysis
+/* Populate real results in this session so @Reanalyze cannot silently fall
+   back to a fresh collection. */
+EXEC dbo.sp_BlitzCache @Top = 1;
+IF OBJECT_ID(N'tempdb..##BlitzCacheResults') IS NULL
+    THROW 51000, 'Reanalysis fixture was not created.', 1;
+DECLARE @Cases TABLE (SortOrder varchar(50), QueryFilter varchar(10));
+INSERT @Cases
+SELECT s.SortOrder, f.QueryFilter
+FROM (VALUES ('memory grant'), ('avg memory grant'), ('unused grant'), ('duplicate')) s(SortOrder)
+CROSS JOIN (VALUES ('procedures'), ('functions')) f(QueryFilter);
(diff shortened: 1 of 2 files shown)


BrentOzar:
Addressed the early-validation and reanalysis feedback in 8d1eef19. Sort aliases and query filters are now normalized before plan-cache preprocessing; incompatible combinations return before the query-hash/duplicate scans, stored-procedure setup, or @Reanalyze jump. Query-hash prefix parsing is separated from its collection step so composite aliases retain their behavior.

Added automated smoke steps covering all ten rejected combinations with @Reanalyze = 0 and 1, three alias/composite cases with both modes, and six supported controls (procedure/function CPU, procedure spills, statement memory grants, and query-hash aliases/default). The tests passed on SQL Server 2022. Against the previous PR head, the regression confirms that reanalysis bypasses the friendly validation and enters Results, then fails with Invalid object name '#plan_creation'. The fixed code returns the expected validation error instead. git diff --check passed.

PR #4096, opened 2026-09-12
sp_BlitzCache: validate reserved output names before quoting them

### LINKED ISSUE #4088: sp_BlitzCache: reserved global output-table names bypass validation
Confirmed on dev `2e2ac07e3ea78f0d0156a4de457da888c87f78e2`, SQL Server 2022 on Linux, using the unchanged procedure in a disposable instance.

```sql
EXEC dbo.sp_BlitzCache @Top=1, @OutputTableName='##BlitzCacheProcs';
```

Expected: the existing friendly error that this table name is reserved.

Actual: execution reaches `Writing results to table` and fails with Msg 207, `Invalid column name 'ServerName'`.

The global output branch detects the `##` prefix inside the already QUOTENAME-wrapped `@OutputTableName` (`SUBSTRING(...,2,2)`), but its reserved-name check compares that same wrapped value against unquoted `##BlitzCacheProcs` and `##BlitzCacheResults`. The check therefore cannot match normal inputs, allowing the output path to operate on the procedure's own internal tables.

Validate against normalized raw names or correctly quoted reserved names, preferably before performing the expensive analysis. Cover both reserved names and an ordinary global output table in regression validation. The observed failure is an invalid-column error; this report does not claim successful deletion of the internal table.

BrentOzar:
Output directed to the internal ##BlitzCacheProcs or ##BlitzCacheResults tables can bypass the reserved-name check because the old comparison runs after QUOTENAME. Validate raw names with an explicit case-insensitive collation before plan-cache preprocessing and return the explanatory error immediately, including when the procedure is installed in a case-sensitive database.

Schema qualification does not make a ## name permanent on SQL Server, so qualified reserved names remain rejected.

Closes #4088.

Validation: added two automated SQL smoke matrix steps. A case-sensitive database hosts the procedure and tests both reserved names, case variants, duplicate/query-hash sorts, and qualified destinations. A real cached workload verifies an ordinary global output table contains rows and the expected schema. These regressions passed on SQL Server 2022 Linux; the previous PR head fails the new regression with an invalid-column error after a lowercase name bypasses validation. git diff --check passed. Generated installers are unchanged per CONTRIBUTING.md.

### DIFF
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -146,7 +146,7 @@ EXEC dbo.sp_BlitzCache
    procedure's comparisons case-sensitive even when master/tempdb are not. */
 IF DB_ID(N'FRKReservedNameTest') IS NOT NULL
     THROW 51000, 'Reserved-name fixture database already exists.', 1;
-CREATE DATABASE FRKReservedNameTest COLLATE Latin1_General_100_CS_AS;
+EXEC(N'CREATE DATABASE FRKReservedNameTest COLLATE Latin1_General_100_CS_AS;');
 BEGIN TRY
     DECLARE @Definition nvarchar(max) = OBJECT_DEFINITION(OBJECT_ID(N'dbo.sp_BlitzCache'));
     SET @Definition = REPLACE(@Definition, N'ALTER PROCEDURE dbo.sp_BlitzCache', N'CREATE PROCEDURE dbo.sp_BlitzCache');
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -152,12 +152,14 @@ BEGIN TRY
     SET @Definition = REPLACE(@Definition, N'ALTER PROCEDURE dbo.sp_BlitzCache', N'CREATE PROCEDURE dbo.sp_BlitzCache');
     EXEC FRKReservedNameTest.sys.sp_executesql @Definition;
 
-    DECLARE @Names TABLE (Name sysname, SortOrder varchar(50), Qualified bit);
+    DECLARE @Names TABLE (Name sysname COLLATE Latin1_General_100_BIN2, SortOrder varchar(50), Qualified bit);
     INSERT @Names VALUES
         (N'##BlitzCacheProcs', 'cpu', 0),
         (N'##BlitzCacheResults', 'cpu', 0),
         (N'##blitzcacheprocs', 'duplicate', 0),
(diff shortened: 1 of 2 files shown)


BrentOzar:
Addressed the casing, validation-order, and regression-coverage feedback in d4604056. The raw-name guard now uses an explicit case-insensitive collation and runs before query-hash/duplicate preprocessing. Added CI matrix cases that install the procedure in a case-sensitive database, reject both reserved names and casing variants (including duplicate/query-hash sorts), and verify ordinary global output contains rows.

The suggested exception for fully qualified permanent tables is not valid on SQL Server. On a disposable SQL Server 2022 instance, `USE FRKReservedProof; CREATE TABLE dbo.[##FRKReservedProof](n int);` created the object in tempdb: `OBJECT_ID(N'dbo.[##FRKReservedProof]')` was NULL and `OBJECT_ID(N'tempdb..##FRKReservedProof')` was non-NULL. Schema qualification does not turn a name starting with ## into a permanent table. The guard therefore intentionally rejects qualified reserved names too, and both qualified cases are covered by the regression.

Validation: the new regression fails against the previous PR head with Invalid column name 'ServerName' after the lowercase name bypasses validation, and passes against the fix. Ordinary global output also passed with a real cached workload. git diff --check passed. Generated installers remain unchanged per CONTRIBUTING.md.

BrentOzar:
Changed the fixture to execute CREATE DATABASE in a separate dynamic batch in f4b17126, as requested. The prior form did execute successfully in the SQL Server 2017 and 2025 CI runs and in local SQL Server 2022, but using a separate batch makes the intent explicit. The updated six rejection cases and nonempty normal-output control also passed locally. Please review the updated commit.

BrentOzar:
Addressed the additional accent/width-equivalence finding in 0fb546f9. The guard now reserves equivalent spellings using Latin1_General_100_CI_AI, which is insensitive to case, accents, kana, and width. This intentionally reserves those spellings even on a more sensitive tempdb, rather than allowing a potentially conflicting internal destination.

Added accented `##BlitzCachéProcs` and full-width-B `##ＢlitzCacheResults` cases. The test case list uses a binary collation so variant rows cannot collapse during iteration/deletion. All eight rejection cases and the nonempty ordinary global-output control passed on SQL Server 2022; git diff --check passed. Holding the merge for fresh review and CI of this commit.

BrentOzar:
Integrated current dev after #4095 merged. Both cache regression suites are preserved with unique step labels; all four new steps pass against the combined SQL Server 2022 procedure. The diff against dev remains only the reserved-name guard and its two regression steps. Requesting review/checks of the integration commit before merging.

PR #4097, opened 2026-09-12
sp_BlitzAnalysis: apply the schema default before building history names

### LINKED ISSUE #4067: sp_BlitzAnalysis: explicit NULL schema reports existing logging tables as missing
## Version and environment
Confirmed on dev `2e2ac07e`, SQL Server 2022 Linux, after the complete smoke suite populated the BlitzFirst logging tables.

## Problem
The procedure promises to default a NULL @OutputSchemaName to dbo, and includes code that does so. However, it constructs all six @FullOutputTableName... variables before applying that default. Concatenation with QUOTENAME(NULL) makes all six names NULL. Changing @OutputSchemaName later does not rebuild them, so the procedure reports that existing logging tables cannot be found and omits their data.

## Repro
With logged data in dbo.BlitzFirst and the other standard output tables:
```sql
EXEC dbo.sp_BlitzAnalysis
 @OutputDatabaseName=N'YourLoggingDatabase', @OutputSchemaName=N'dbo';
EXEC dbo.sp_BlitzAnalysis
 @OutputDatabaseName=N'YourLoggingDatabase', @OutputSchemaName=NULL;
```
Expected: equivalent results.
Actual: the dbo call returns logged findings; the NULL call prints `Getting BlitzFirst info from (null)` followed by `Table provided for BlitzFirst data: (null) does not exist`, and analogous messages for the other tables.

## Suggested fix
Normalize @OutputSchemaName before constructing the fully qualified table names. Add a regression comparing explicit dbo with explicit NULL against populated logging tables.

Fix PR will follow the complete review pass.

BrentOzar:
Passing @OutputSchemaName=NULL currently builds NULL fully qualified history-table names before applying the intended dbo fallback, then reports that existing history cannot be found. Apply the fallback before constructing those names.

Closes #4067.

Validation: complete-procedure SQL Server 2022 tests with separate dbo and custom history fixtures verified NULL and explicit dbo both return the default history, while an explicit custom schema returns only its own history. Fixture database removed; `git diff --check` passed.

### DIFF
--- a/.github/scripts/analysis-schema-regression.sql
+++ b/.github/scripts/analysis-schema-regression.sql
@@ -1,0 +1,29 @@
+USE FRKSmokeTest;
+SET NOCOUNT ON;
+IF SCHEMA_ID(N'FRKAnalysisSmoke') IS NULL EXEC(N'CREATE SCHEMA FRKAnalysisSmoke');
+DROP TABLE IF EXISTS dbo.FRKSchemaHistory;
+DROP TABLE IF EXISTS FRKAnalysisSmoke.FRKSchemaHistory;
+CREATE TABLE dbo.FRKSchemaHistory
+(ServerName nvarchar(128), CheckDate datetimeoffset, CheckID int, Priority int,
+ FindingsGroup nvarchar(128), Finding nvarchar(128), URL nvarchar(256),
+ Details nvarchar(max), HowToStopIt nvarchar(max), QueryPlan xml, QueryText nvarchar(max));
+INSERT dbo.FRKSchemaHistory VALUES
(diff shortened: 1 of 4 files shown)


BrentOzar:
Added automated populated-history coverage for NULL, explicit dbo, and explicit custom-schema output. The CI runner checks that each call returns its expected finding marker and excludes the other schema's marker. The SQL fixture cleans its dedicated history tables/schema. All three cases and the actual shell result-checking function passed on SQL Server 2022. The pre-fix procedure completes but omits the expected marker for NULL, so the new assertion catches the original silent failure.

The installer-regeneration suggestion conflicts with CONTRIBUTING.md: "Don't touch the files that start with Install, like Install-All-Scripts.sql. Those are dynamically generated." Those release artifacts therefore remain unchanged in this PR. git diff --check and shell syntax pass.

PR #4100, opened 2026-09-12
sp_ineachdb: honor include and exclude lists on Azure SQL Database

### LINKED ISSUE #4065: sp_ineachdb: Azure SQL DB ignores @database_list and @exclude_list
## Version and environment
Confirmed on dev `2e2ac07e` (sp_ineachdb 8.34, 2026-07-02) against Azure SQL Database, EngineEdition 5.

## Problem
The Azure branch inserts the current database into #ineachdb and bypasses the recursive list parser entirely. The later DELETE filters do not apply @database_list or @exclude_list, so a command executes even when the current database is explicitly excluded or a different database is explicitly selected.

This breaks the filter contract and can execute maintenance commands against a database the caller intended to protect. The Azure support design in #3943 explicitly says existing filters still apply.

## Repro
Run in any Azure SQL user database with current dev sp_ineachdb installed:
```sql
CREATE TABLE #Ran (DBName sysname);
DECLARE @CurrentDB sysname = DB_NAME();
EXEC dbo.sp_ineachdb
    @command = N'INSERT #Ran SELECT DB_NAME();',
    @exclude_list = @CurrentDB;
SELECT * FROM #Ran; -- Actual: current database; expected: no rows

TRUNCATE TABLE #Ran;
EXEC dbo.sp_ineachdb
    @command = N'INSERT #Ran SELECT DB_NAME();',
    @database_list = N'NoSuchDatabase_CodexReview';
SELECT * FROM #Ran; -- Actual: current database; expected: no rows
```
Both cases reproduced using an isolated copy of the unchanged procedure and temporary tables; no user data was modified.

## Suggested fix
Apply the existing include/exclude list semantics to the Azure current-database candidate too, including bracket escaping, quoted names, and exclusion taking precedence. Preserve boxed SQL Server and Managed Instance behavior.

Found during the requested full root-script review. Fix PR will follow after the complete review pass.

BrentOzar:
Azure's current-database shortcut bypasses the shared list parser, so @database_list and @exclude_list are ignored. Feed the current Azure database through the same parser and selection logic as SQL Server. Azure still offers only the current database as a candidate; SQL Server continues to use sys.databases.

Closes #4065.

Validation: eight cases passed on both SQL Server 2022 and Azure SQL Database: default selection, quoted inclusion, mixed inclusion, a different database only, current-database exclusion, simultaneous inclusion/exclusion, excluding a different database, and whitespace in an unquoted name. Assertions checked exact visit counts and database context. Temporary schemas were removed and existing dbo procedure hashes were unchanged. `git diff --check` passed.

### DIFF
--- a/.github/scripts/azure-ineachdb-regression.sql
+++ b/.github/scripts/azure-ineachdb-regression.sql
@@ -1,0 +1,32 @@
+/* Uses only local temporary state; the Azure runner already verifies edition 5. */
+SET NOCOUNT ON;
+DECLARE @Current sysname = DB_NAME(), @Quoted nvarchar(258) = QUOTENAME(DB_NAME()),
+        @Other sysname = N'FRKMissing_' + CONVERT(nvarchar(36), NEWID());
+CREATE TABLE #Visited (DatabaseName sysname);
+DECLARE @Cases TABLE (CaseID int IDENTITY, Included nvarchar(max), Excluded nvarchar(max), Expected int);
+INSERT @Cases VALUES
+    (NULL, NULL, 1),
+    (@Quoted, NULL, 1),
+    (NULL, @Quoted, 0),
+    (@Quoted, @Quoted, 0),
+    (QUOTENAME(@Other), NULL, 0),
+    (NULL, QUOTENAME(@Other), 1),
+    (QUOTENAME(@Other) + N',' + @Quoted, NULL, 1),
+    (N'  ' + @Current + N'  ', NULL, 1);
+DECLARE @ID int, @In nvarchar(max), @Out nvarchar(max), @Expected int;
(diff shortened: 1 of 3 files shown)


BrentOzar:
Added an Azure CI regression after installing sp_ineachdb, with eight assertions: default/current inclusion, current exclusion, exclusion precedence, different-database inclusion/exclusion, mixed lists, and whitespace. The fixture uses only a local temporary table and verifies both execution count and database identity. The Azure runner already requires EngineEdition 5, so this tests the actual Azure branch. Boxed SQL Server 2022 controls and shell syntax pass; the pushed Azure job will validate the real Azure path.

Generated installer changes remain intentionally excluded: CONTRIBUTING.md explicitly says files starting with Install are dynamically generated and should not be changed in individual contributions.

PR #4103, opened 2026-09-12
sp_DatabaseRestore: enumerate each log-backup directory separately

### LINKED ISSUE #4073: sp_DatabaseRestore: simple enumeration passes the entire comma-separated log path list to xp_dirtree
On dev 2e2ac07e, the log-directory loop calls `master.sys.xp_dirtree @BackupPathLog, 1, 1` instead of passing `@CurrentBackupPathLog`. The full/differential loops correctly use their current path variables.

Consequently, the multiple backup-directory support introduced for #2180 fails for log backups when `@SimpleFolderEnumeration = 1` (also selected automatically when xp_cmdshell is disabled). A comma-separated list is treated as one nonexistent directory.

Validated on a disposable SQL Server 2022 Linux instance with an actual log backup in `/var/opt/mssql/data/CodexReviewv2/log/`. A dry-run call with `@ContinueLogs=1, @Execute='N', @SimpleFolderEnumeration=1` and that single directory passed. The same call with the directory repeated as a comma-separated list failed with:

```
(LOG) No rows were returned for that database CodexReviewSource in path /var/opt/mssql/data/CodexReviewv2/log/
```

This occurs during directory enumeration before any restore. Two distinct valid directories are affected by the same incorrect argument.

Expected: enumerate each parsed directory, combine its files with the correct source directory, and restore striped log backups across multiple directories. Use the loop's current path in xp_dirtree and cover this with a real striped-log restore test.

BrentOzar:
When @BackupPathLog contains multiple directories and @SimpleFolderEnumeration=1, the loop passes the entire comma-separated list to xp_dirtree. Pass the current directory instead, matching the full and differential backup enumeration paths.

Closes #4073.

Validation: actual backup/restore tests passed on SQL Server 2022 Linux and Windows SQL Server 2025. Each test created a full backup containing one row, then two log backups in separate directories adding one row apiece. A single-directory restore recovered two rows; a two-directory restore recovered all three. Source and restored fixture databases were removed. `git diff --check` passed.

### DIFF
--- a/.github/scripts/run-sql-server-smoke-tests.sh
+++ b/.github/scripts/run-sql-server-smoke-tests.sh
@@ -279,6 +279,21 @@ echo "=== Seeding test data ==="
 run_file "$SEED_SQL"
 
 echo
+echo "=== Installing pinned restore-test dependencies ==="
+OLA_REVISION=41617578643e722b790ea4d57ef80f8de38ae747
+for dependency in CommandLog CommandExecute; do
+  curl -fsSL "https://raw.githubusercontent.com/olahallengren/sql-server-maintenance-solution/$OLA_REVISION/$dependency.sql" -o "$WORK_DIR/$dependency.sql"
+done
+(
+  cd "$WORK_DIR"
+  sha256sum --check <<'CHECKSUMS'
+190962fc9802ce8d8f5082a23119567d27f05d2bfc36e1bdf5ce79edea5d8e76  CommandExecute.sql
+7a508fa7ed562ec5ee09d4f587f7f2a87f7a4eb3c57450602ee99cdeb4e18b31  CommandLog.sql
+CHECKSUMS
+)
+run_file "$WORK_DIR/CommandLog.sql"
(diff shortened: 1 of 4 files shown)


BrentOzar:
Addressed both reviews in 9d0f9b92. Each simple-enumeration directory is now validated before its files join prior results. Added CI execution with pinned official Ola dependencies and real full/log backups: one directory, two sequential directories, a striped log across both directories, and populated-first plus empty/missing-second cases. The negative cases assert the named-path error and that the incomplete database remains RESTORING. All cases pass on disposable SQL Server 2022. The previous PR head fails the regression by recovering the incomplete database and silently accepting the empty path.

BrentOzar:
Added an explicit cutoff date to every xp_delete_file call in both cleanup paths. The three-argument form did not fail in the completed 2017/2025 CI or local 2022 run, but the explicit date removes ambiguity and ensures all files in the unique test directories are eligible for deletion. Full local restore regression passes again.

PR #4108, opened 2026-09-12
sp_BlitzLock: read deadlocks from the system_health ring buffer

### LINKED ISSUE #4074: sp_BlitzLock: system_health ring-buffer path silently returns no deadlocks
Confirmed on dev 2e2ac07e with SQL Server 2022 Linux and a real two-session deadlock in a disposable database.

```sql
EXEC dbo.sp_BlitzLock @TargetSessionType=N'ring_buffer';
```

returns no deadlock rows. Directly shredding `sys.dm_xe_session_targets.target_data` proves system_health contains the `xml_deadlock_report`. Reading the same event via `@TargetSessionType=N'event_file'` returns both participating sessions and the one-deadlock finding. A custom ring-buffer session also returns both sessions.

Both ring-buffer collection and shredding branches require `@EventSessionName NOT LIKE N'system_health%'`. There is a special system_health event-file branch, but no corresponding ring-buffer branch. Consequently no source rows are collected for this supported parameter combination. The procedure explicitly defaults Managed Instance's system_health calls to ring_buffer, so source inspection indicates that default is affected too (MI not live-tested).

Expected: system_health ring-buffer events should be parsed just like a custom ring-buffer session. Add regression coverage asserting the actual deadlock rows, not merely successful execution.

BrentOzar:
The ring-buffer collection and parsing branches both exclude system_health, but its special handling only supports event files. Consequently, an explicit system_health ring-buffer request returns no deadlocks even when the target contains them. Use the shared ring-buffer path for both system and custom sessions.

Closes #4074.

Validation: generated a deadlock between two sessions updating only a disposable fixture on SQL Server 2022. Verified the system_health ring buffer contained the event and that persisted sp_BlitzLock output was nonempty for system_health ring_buffer, a custom ring_buffer, and system_health event_file. Temporary output tables were removed. `git diff --check` passed.

### DIFF
--- a/.github/scripts/run-sql-server-smoke-tests.sh
+++ b/.github/scripts/run-sql-server-smoke-tests.sh
@@ -221,7 +221,10 @@ THROW 51000, 'The dedicated victim timed out without being killed.', 1;" \
 )
 
 run_step() {
-  if [[ "$2" == 'sp_kill kills a dedicated session' ]]; then
+  if [[ "$2" == 'sp_BlitzLock parses a real system_health ring-buffer deadlock' ]]; then
+    SQLCMDSERVER="$SQLCMDSERVER" SQLCMDUSER="$SQLCMDUSER" SQLCMD="$SQLCMD" \
+      python3 "$SCRIPT_DIR/test-lock-ring-buffer.py"
+  elif [[ "$2" == 'sp_kill kills a dedicated session' ]]; then
     run_kill_step "$1"
   else
     "$SQLCMD" "${SQLCMD_ARGS[@]}" -d master -i "$1"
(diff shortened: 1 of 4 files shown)


BrentOzar:
Added the requested real-deadlock CI regression. Two concurrent SQL sessions acquire opposite row locks and synchronize before taking the second lock. The test requires exactly one 1205 victim, waits for the unique fixture's event in system_health's ring buffer, calls the exact system_health/ring_buffer path, and requires two distinct parsed participant SPIDs in persisted output. Unique schema/table names prevent old deadlocks satisfying the assertion; owned fixtures and leftover synonyms are cleaned.

The regression passes with the fix on SQL Server 2022 and fails against the original dev procedure at the parsed-participant assertion, despite proving the event exists in the ring buffer. Shell syntax and git diff --check pass.

BrentOzar:
Integrated current dev and made SQLCMDPASSWORD handling explicit in 4b1098ee. sqlcmd supports that environment variable and subprocesses previously inherited it (the prior real-deadlock tests and all CI jobs passed). The test now checks it is present and supplies an explicit environment copy to every child, keeping the secret off process arguments. The real two-session deadlock and two-participant output assertion pass again on SQL Server 2022.

PR #4110, opened 2026-09-12
sp_BlitzCache: keep analysis and export updates within their session

### LINKED ISSUE #4084: sp_BlitzCache: ExportToExcel modifies query text belonging to other sessions
On dev commit 2e2ac07e3ea78f0d0156a4de457da888c87f78e2, the Excel-formatting UPDATE of ##BlitzCacheProcs has no SPID filter. Running @ExportToExcel = 1 rewrites and truncates QueryText in rows belonging to other sessions, changing their later output/reanalysis. The all/all avg export branches contain the same unfiltered UPDATE.

Confirmed with the unchanged procedure on a disposable SQL Server 2022 Linux instance. This deterministic fixture models another session's row without needing a timing race:

```sql
EXEC dbo.sp_BlitzCache @Top=1, @SkipAnalysis=1, @HideSummary=1;
INSERT ##BlitzCacheProcs(SPID, DatabaseName, QueryText)
VALUES(-9876, N'master', N'SELECT  123 /* other session */');
SELECT SPID, QueryText FROM ##BlitzCacheProcs WHERE SPID=-9876;
EXEC dbo.sp_BlitzCache @Top=1, @SkipAnalysis=1,
    @HideSummary=1, @ExportToExcel=1;
SELECT SPID, QueryText FROM ##BlitzCacheProcs WHERE SPID=-9876;
DELETE ##BlitzCacheProcs WHERE SPID=-9876;
```

Before: `SELECT  123 /* other session */` (two spaces). After: `SELECT 123 /* other session */` (one space), even though the sentinel SPID is unrelated to @@SPID. The same expression also caps text at 32,000 characters.

Expected: only this invocation's rows are modified. Add session scoping to these updates and audit the other shared-table analysis updates for missing SPID restrictions.

Related historical issue #653 was closed as fixed; this is a confirmed remaining path in current dev. No production objects were changed by the reproduction.

BrentOzar:
sp_BlitzCache analysis and Excel formatting can update another invocation's rows in the shared global cache table. Scope analysis and direct-export updates to @@SPID, and format the local #bou_allsort result table in multi-sort exports.

Closes #4084.

Added an automated regression that shares real plan handles with sentinel rows for another SPID, verifies analysis covers those handles without changing the other rows, and runs complete direct, all, and all avg Excel exports while asserting the other rows' count, text, and cost remain unchanged. The complete regression passed with the Linux SQL client on SQL Server 2022, including after integrating current dev. The old procedure fails the cross-session analysis assertion.

An earlier macOS sqlcmd attempt stalled and was terminated; its exit status is not counted as validation. The subsequent complete Linux-client runs finished naturally with exit 0 and the final assertion marker. git diff --check passed.

### DIFF
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -353,3 +353,44 @@ IF @QueryPlanHash IS NULL
     RAISERROR('Seeded marker query was not found in the plan cache; sp_BlitzPlanCompare was not exercised.', 16, 1);
 
 EXEC dbo.sp_BlitzPlanCompare @QueryPlanHash = @QueryPlanHash, @DatabaseName = 'FRKSmokeTest';
+
+--#STEP: sp_BlitzCache isolates analysis and all Excel export paths
+EXEC FRKSmokeTest.sys.sp_executesql
+     N'SELECT COUNT_BIG(*) FROM dbo.Posts WHERE Id > 10 /* FRK isolation workload */';
+EXEC dbo.sp_BlitzCache @Top=100, @IgnoreSystemDBs=0, @SkipAnalysis=1, @HideSummary=1;
+IF NOT EXISTS(SELECT 1 FROM ##BlitzCacheProcs WHERE SPID=@@SPID AND QueryHash IS NOT NULL)
+    THROW 51000,'Isolation fixture has no cached statement.',1;
+DELETE ##BlitzCacheProcs WHERE SPID=-9876;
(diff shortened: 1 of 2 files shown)


BrentOzar:
The previously interrupted full multi-sort regression is now completed and committed. It executes real analysis and direct/all/all-avg Excel exports against shared real plan handles, asserting another SPID's sentinel rows retain their count, text, and cost. Both this branch and its integration with current dev passed using the Linux SQL client, with natural exit 0 and the final assertion marker. Original dev fails at 'Analysis changed another session.' The stale description has been replaced with the completed validation details.

PR #4111, opened 2026-09-12
sp_BlitzFirst: isolate history deltas by server and upgrade existing views

### LINKED ISSUE #4081: sp_BlitzFirst: delta views duplicate samples when multiple servers share check dates
Confirmed on unchanged dev `2e2ac07e3ea78f0d0156a4de457da888c87f78e2`, SQL Server 2022 Linux. Related closed #1826 addressed ordering by server, but the generated delta views still omit ServerName from the RowDates/CheckDates output and date joins.

### Reproduction
Use sp_BlitzFirst to generate an isolated wait-statistics table and view:
```sql
EXEC master.dbo.sp_BlitzFirst @Seconds=0,
    @OutputDatabaseName=N'YourTestDatabase', @OutputSchemaName=N'dbo',
    @OutputTableNameWaitStats=N'WaitReview';
```
In that test database only, replace the test table's contents with:
```sql
DELETE dbo.WaitReview;
INSERT dbo.WaitReview
 (ServerName,CheckDate,wait_type,wait_time_ms,signal_wait_time_ms,waiting_tasks_count)
VALUES
 ('ServerA','2026-09-12T00:00:00+00:00','LCK_M_X',1000,0,10),
 ('ServerA','2026-09-12T00:05:00+00:00','LCK_M_X',2000,0,20),
 ('ServerB','2026-09-12T00:00:00+00:00','LCK_M_X',3000,0,30),
 ('ServerB','2026-09-12T00:05:00+00:00','LCK_M_X',4000,0,40);
SELECT ServerName,wait_time_ms_delta,ElapsedSeconds
FROM dbo.WaitReview_Deltas;
```
Expected two rows: one 1000ms/300-second delta for each server. Actual output has four rows: each server's delta appears twice. Aggregating the view therefore doubles the reported waits.

Carry ServerName through both date CTEs, calculate previous samples within each server, and include ServerName in the joins to those CTEs. The generated FileStats and PerfmonStats delta views use the same defective date-join pattern and should be corrected and tested alongside WaitStats. Existing views will also need a reliable upgrade path; checking only for the already-existing JoinKey column will leave this bug in place.

BrentOzar:
When multiple servers write to the same history tables, the file, perfmon, and wait delta views can duplicate rows or pair samples using another server's timeline. Partition sample numbering by ServerName and include ServerName in both date joins.

Existing views are upgraded with CREATE OR ALTER VIEW, preserving their object IDs and grants. A definition marker prevents repeating the alteration on subsequent collections.

Validation: reproduced incorrect row counts in all three original views on SQL Server 2022 and Windows SQL Server 2025. With the fix, two servers with shared and staggered sample times produced exactly four expected deltas per view, with correct values and elapsed times. Upgrade assertions confirmed unchanged object IDs and SELECT grants; repeat calls preserved modify_date. Fresh creation and a single-server control passed on SQL Server 2022. git diff --check passed.

Closes #4081.

### DIFF
--- a/.github/scripts/run-sql-server-smoke-tests.sh
+++ b/.github/scripts/run-sql-server-smoke-tests.sh
@@ -243,10 +243,25 @@ run_analysis_schema_step() {
   done
 }
 
+run_delta_step() {
+  local existing status=0
+  existing="$(run_scalar "SET NOCOUNT ON; SELECT COUNT(*) FROM sys.databases WHERE name=N'FRKDeltaSmoke';")" || return 1
+  if [[ "$existing" != "0" ]]; then
+    echo "Delta fixture already exists; refusing to overwrite it."
+    return 1
+  fi
+  run_file "$1" || status=$?
+  # The matrix continues after failures, so clean our database on both paths.
+  run_query "IF DB_ID(N'FRKDeltaSmoke') IS NOT NULL DROP DATABASE FRKDeltaSmoke;" || status=1
+  return "$status"
+}
+
 run_step() {
   if [[ "$2" == 'sp_BlitzLock parses a real system_health ring-buffer deadlock' ]]; then
     SQLCMDSERVER="$SQLCMDSERVER" SQLCMDUSER="$SQLCMDUSER" SQLCMD="$SQLCMD" \
       python3 "$SCRIPT_DIR/test-lock-ring-buffer.py"
+  elif [[ "$2" == 'sp_BlitzFirst multi-server deltas and in-place upgrades' ]]; then
+    run_delta_step "$1"
   elif [[ "$2" == 'sp_BlitzAnalysis defaults and isolates output schemas' ]]; then
     run_analysis_schema_step
   elif [[ "$2" == 'sp_kill kills a dedicated session' ]]; then
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -505,6 +505,24 @@ DELETE dbo.Files WHERE ServerName NOT IN(N'ServerA',N'ServerB');
 DELETE dbo.Perfmon WHERE ServerName NOT IN(N'ServerA',N'ServerB');
 DELETE dbo.Waits WHERE ServerName NOT IN(N'ServerA',N'ServerB');
 IF (SELECT COUNT(*) FROM dbo.Files_Deltas)<>4 OR (SELECT COUNT(*) FROM dbo.Perfmon_Deltas)<>4 OR (SELECT COUNT(*) FROM dbo.Waits_Deltas)<>4 THROW 51000,'Incorrect view row counts',1;
+IF EXISTS(
+    SELECT 1 FROM (SELECT ServerName,CheckDate FROM dbo.Expected WHERE ElapsedSeconds IS NOT NULL) e
+    FULL OUTER JOIN (SELECT ServerName,CheckDate,COUNT(*) AS Copies FROM dbo.Files_Deltas GROUP BY ServerName,CheckDate) d
+    ON e.ServerName=d.ServerName AND e.CheckDate=d.CheckDate
+    WHERE e.ServerName IS NULL OR d.ServerName IS NULL OR d.Copies<>1)
+    THROW 51000,'Missing, extra, or duplicate Files delta key.',1;
+IF EXISTS(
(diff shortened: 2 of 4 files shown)


BrentOzar:
Added an automated regression covering two servers with matching and staggered sample times across file, Perfmon, and wait delta views. It asserts exact counts/values, upgrades existing markerless views in place, preserves object IDs and SELECT grants, verifies marker visibility for a user without ALTER permission, and confirms repeat collection does not change modify_date. The regression passes on SQL Server 2022 and fails on original dev with incorrect view row counts.

Documented the privileged migration path in Documentation/sp_BlitzFirst_History_View_Upgrade.md and linked it from @Help: run one collection as the history database owner with the job's actual output names, then ensure the regular collector has VIEW DEFINITION on the three delta views so it recognizes the migration marker without needing permanent ALTER rights. Existing procedure/DMV/output-table permissions remain necessary. This explicitly covers the scheduled-login concern; object grants are preserved rather than automatically broadened.

git diff --check and the Help-output check pass.

BrentOzar:
Integrated current dev to clear the test overlap. The combined procedure passed the full multi-server delta/upgrade/metadata/repeat regression on SQL Server 2022, and both the delta and Analysis schema smoke steps remain present with unique labels. Requesting review of the integration commit.

BrentOzar:
Addressed the earlier suppressed detailed feedback in 08cfca64. All three nested CREATE OR ALTER VIEW literals now use N prefixes. Both fresh-create and in-place-upgrade regression blocks compare expected and actual keys bidirectionally and require exactly one row per server/date, preventing duplicate-plus-missing rows from satisfying the old count/value check. The complete local regression passes: exact file/Perfmon/wait deltas, stable object IDs/grants, collector metadata visibility without ALTER, and unchanged view modification dates on repeat collection.

BrentOzar:
Both regression blocks now explicitly reject NULL delta values and NULL elapsed times. The focused history-view regression passes on SQL Server 2022, and latest dev (including restore, uninstall, ring-buffer, and cache changes) is integrated in 36321770.

BrentOzar:
Addressed the latest fixture feedback in 5d5c00fe. Server names now include generated GUIDs, and cleanup retains only names recorded in the Expected fixture table. The CI runner wraps this step with a preexisting-database guard and unconditional post-run cleanup, preserving the original failure status. Tested the actual wrapper both with the complete passing regression and an intentional SQL assertion failure; the fixture database is removed in both cases.

BrentOzar:
Fixed the Codex long-identifier finding: each of the three migrated view statements now starts with an nvarchar(max) operand, preventing truncation before assignment. Added a regression using a 128-character schema and 120-character file-history table, verifying the stored definition exceeds 4,000 characters. The full fixture passes with the fix; the previous head fails with truncated SQL / an unclosed literal. The prior version was restored only for that negative test, then the fixed version reinstalled.

BrentOzar:
4186215d addresses the latest findings. All three migration gates match the complete marker comment. The legacy fixtures now retain JoinKey and include the bare marker token as an identifier, so the previous gate fails this regression. Repeat collection explicitly verifies object IDs and SELECT grants as well as modification dates. Upgrade instructions now grant metadata visibility only for views enabled by the job. The full migration/long-name regression passes locally.

BrentOzar:
The latest fixture finding is a false positive: the stubs contain only the bare identifier FRK_ServerScopedDeltas_v1. The production gate now searches for the complete comment /* FRK_ServerScopedDeltas_v1 */ including delimiters and spaces (all three gates changed in 4186215d). Those strings do not match. Keeping the bare identifier is intentional: it reproduces the collision found in the preceding review. Removing it would weaken the regression.

The full existing-view migration test passes locally and on all three current-head CI checks, including SQL Server 2017 and 2025. The assertions after the stub replacement verify four rows per view, correct per-server deltas, unchanged object IDs and retained grants; a skipped upgrade cannot pass those checks. Resolving this finding with this code/test evidence and requesting reassessment of the actual complete-marker predicate.

BrentOzar:
51d18604 corrects the help text to say each configured _Deltas view. It also adds an explicit pre-migration assertion that none of the three legacy fixture definitions contains the complete comment marker. The bare identifier is intentionally retained to test the old collision; it does not satisfy CHARINDEX of the complete comment. The full combined local smoke run has also passed this migration step.

PR #4113, opened 2026-09-12
Escape output names correctly in seven procedures

### LINKED ISSUE #4070: Output names containing apostrophes fail in seven procedures
## Version / environment
Reproduced on dev `2e2ac07e`, SQL Server 2022 Linux, using the disposable smoke-test database and unchanged installed root procedures.

## Problem
Several output-table paths put QUOTENAME'd identifiers inside single-quoted dynamic SQL metadata predicates without escaping apostrophes. QUOTENAME protects identifier delimiters, not a surrounding string literal. A valid output table name such as `Codex'Who` breaks the generated SQL.

Confirmed affected procedures:
- sp_Blitz
- sp_BlitzCache
- sp_BlitzFirst
- sp_BlitzIndex
- sp_BlitzWho
- sp_kill

## Examples
```sql
EXEC dbo.sp_BlitzWho
 @OutputDatabaseName=N'YourLoggingDB', @OutputSchemaName=N'dbo',
 @OutputTableName=N'Codex''Who';

EXEC dbo.sp_BlitzIndex
 @DatabaseName=N'YourLoggingDB', @Mode=2,
 @OutputDatabaseName=N'YourLoggingDB', @OutputSchemaName=N'dbo',
 @OutputTableName=N'Codex''Index';
```
Expected: create/use the legal bracketed output tables and log results.
Actual: syntax errors near the portion after the apostrophe. The equivalent normal-name output calls passed in the full smoke matrix.

The same test was run for the other four listed procedures with their supported output parameters; sp_BlitzFirst used @Seconds=0 to keep the reproduction short. sp_BlitzLock returned before its table-output path due to no deadlocks, so it is not claimed as confirmed affected here.

## Suggested fix
Keep QUOTENAME for identifiers, and parameterize metadata lookup values or separately escape apostrophes where values are embedded in dynamic SQL string literals. Review database/schema names in the same paths too. Add regressions using apostrophes, closing brackets, and Unicode names.

Fix PRs will follow the full review pass.

Additional full-pass validation: sp_BlitzLock is also affected when reusing an existing output table with an apostrophe in its name. With a real captured deadlock as table input, `@OutputDatabaseName=N'FRKSmokeTest', @OutputTableName=N'Codex''Lock'` fails with error 102 near `]` in the output-table metadata checks. Its OBJECT_ID literal construction also needs to escape apostrophes independently from identifier quoting.

BrentOzar:
Depends on #4111; merge that history-view fix first. The output-name changes are combined with its server-scoped views here because Git's automatic merge of the two independent changes duplicated part of the view definitions.

Valid output names such as `Codex'Who` fail because bracket-quoted identifiers are embedded in SQL string literals without escaping their apostrophes. Fix metadata lookups, global temporary output paths, and nested commands in sp_Blitz, sp_BlitzCache, sp_BlitzFirst, sp_BlitzIndex, sp_BlitzWho, sp_BlitzLock, and sp_kill. Keep identifier quoting, escape each surrounding string layer separately, and use Unicode literals. Remote output escapes the whole nested command.

Validation: all seven persistent output paths passed creation/reuse on SQL Server 2022 and Windows SQL Server 2025. SQL Server 2022 also passed names combining apostrophes, Unicode, and closing brackets. First's generated history views were exercised. Blitz, Cache, and Index passed a temporary Windows loopback linked-server test. Global temporary output passed for Blitz, Cache, Index, and First main/file/wait outputs. Six procedures passed Azure persistent output independently; Index's separate Azure failure is fixed by #4114.

Combined validation completed: history-view upgrades, retained grants, per-server deltas, and quoted Unicode view names passed after combining with #4111. The local integration branch containing all review fixes passed 41 Linux smoke checks, 37 applicable Windows checks, and 17 Azure calls (including this change with #4114). Azure's 21 existing dbo procedure hashes stayed unchanged. Temporary databases, schemas, backup files, linked server, and container were cleaned up. This PR's Azure/SQL Server 2017/SQL Server 2025 CI checks passed.

Closes #4070.

### DIFF
--- a/.github/scripts/analysis-schema-regression.sql
+++ b/.github/scripts/analysis-schema-regression.sql
@@ -1,0 +1,29 @@
+USE FRKSmokeTest;
+SET NOCOUNT ON;
+IF SCHEMA_ID(N'FRKAnalysisSmoke') IS NULL EXEC(N'CREATE SCHEMA FRKAnalysisSmoke');
+DROP TABLE IF EXISTS dbo.FRKSchemaHistory;
+DROP TABLE IF EXISTS FRKAnalysisSmoke.FRKSchemaHistory;
+CREATE TABLE dbo.FRKSchemaHistory
+(ServerName nvarchar(128), CheckDate datetimeoffset, CheckID int, Priority int,
+ FindingsGroup nvarchar(128), Finding nvarchar(128), URL nvarchar(256),
+ Details nvarchar(max), HowToStopIt nvarchar(max), QueryPlan xml, QueryText nvarchar(max));
+INSERT dbo.FRKSchemaHistory VALUES
+(@@SERVERNAME, SYSDATETIMEOFFSET(), 1, 10, N'Test', N'FRK_DBO_SCHEMA_SENTINEL', NULL, NULL, NULL, NULL, NULL);
+SELECT * INTO FRKAnalysisSmoke.FRKSchemaHistory FROM dbo.FRKSchemaHistory;
+UPDATE FRKAnalysisSmoke.FRKSchemaHistory SET Finding = N'FRK_CUSTOM_SCHEMA_SENTINEL';
+BEGIN TRY
+    EXEC master.dbo.sp_BlitzAnalysis @OutputDatabaseName = N'FRKSmokeTest',
+         @OutputSchemaName = $(SchemaArgument), @OutputTableNameBlitzFirst = N'FRKSchemaHistory',
+         @OutputTableNameFileStats = NULL, @OutputTableNamePerfmonStats = NULL,
+         @OutputTableNameWaitStats = NULL, @OutputTableNameBlitzCache = NULL,
+         @OutputTableNameBlitzWho = NULL;
+    DROP TABLE dbo.FRKSchemaHistory;
+    DROP TABLE FRKAnalysisSmoke.FRKSchemaHistory;
+    DROP SCHEMA FRKAnalysisSmoke;
+END TRY
+BEGIN CATCH
+    DROP TABLE IF EXISTS dbo.FRKSchemaHistory;
+    DROP TABLE IF EXISTS FRKAnalysisSmoke.FRKSchemaHistory;
+    DROP SCHEMA FRKAnalysisSmoke;
+    THROW;
+END CATCH;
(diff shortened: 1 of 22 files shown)


BrentOzar:
Integrated the latest #4111 parent and current cache/Azure changes, then reconciled the nested First view SQL explicitly (Git's automatic merge duplicated fragments despite reporting no conflict). Added CI coverage across all seven procedures using a database, schema, and tables containing an apostrophe, Unicode, and a closing bracket. Checks cover creation/reuse, populated append where applicable, First history tables/views plus plain-text history retrieval, and two parsed deadlock participants from a sanitized fixture. All seven procedures compile and the full quoted-output regression passes on SQL Server 2022; the combined exact multi-server delta/upgrade regression also passes. This remains stacked on #4111.

BrentOzar:
The bracketed Index inputs are intentional: the initial normalization at sp_BlitzIndex.sql lines 336–339 runs PARSENAME on all three names before the later QUOTENAME calls. Raw closing-bracket names fail that parser; the delimited names correctly identify the fixture. Both create and append assertions pass on local SQL Server 2022 and the prior head's 2017/2025 CI. Added an explanatory comment and reran all seven cases after integrating the latest parent, with an explicit assertion for both deadlock participants; all pass.

BrentOzar:
The Lock first-create branch already receives quoted identifiers: the shared SELECT immediately before `IF (@r IS NOT NULL)` assigns QUOTENAME to all three output names, so it applies equally to the existing-table and ELSE create branches. Added an explicit comment at that shared normalization point. The fresh-create test uses a new database, creates the table, verifies nonempty output with both participants, then verifies reuse/append; local 2022 and 2017/2025 CI pass. Integrated the latest parent fixture cleanup changes as well.

BrentOzar:
Pushed a352e3ce: merged the parent MAX-length view fix, added passing quoted Unicode global-temp coverage for Blitz/Cache/Index and all four BlitzFirst outputs, and verified output-fixture cleanup on both success and intentional failure. Also fixed the unrelated deadlock CI timing failure: system_health was configured for 120-second dispatch with truncated ring XML, while the fixture waited 20 seconds. The disposable-instance test now resets the ring and temporarily uses one-second dispatch, restoring its original setting afterward. The real two-participant deadlock regression passes locally.

BrentOzar:
6c501729 incorporates the parent's complete migration marker and narrows both cleanup predicates to dbo.DeadLockTbl and dbo.DeadlockFindings with the fixture database match. Tested deliberate failure through the actual shell wrapper: its own synonym and database were removed, an unrelated synonym pointing at that same database survived, and failure status was preserved. The combined history migration regression also passes.

PR #4114, opened 2026-09-12
sp_BlitzIndex: allow local output tables on Azure SQL Database

### LINKED ISSUE #4112: sp_BlitzIndex: output tables fail on Azure SQL DB because local names include a server prefix
Reproduced twice with unchanged dev 2e2ac07e on Azure SQL Database (EngineEdition 5), using ordinary output names. This is independent of the apostrophe bug in #4070.

```sql
DECLARE @db sysname = DB_NAME();
EXEC dbo.sp_BlitzIndex
    @Mode = 2,
    @OutputDatabaseName = @db,
    @OutputSchemaName = N'dbo',
    @OutputTableName = N'BlitzIndexReview';
```

Expected: save index inventory in the current database.

Actual: the procedure reports "Failure analyzing temp tables" and raises error 50000 with the underlying message that the server/database-qualified INFORMATION_SCHEMA.SCHEMATA reference is not supported in this version of SQL Server. No output table is created.

The local output path assigns SERVERPROPERTY('ServerName') to @OutputServerName, then builds four-part metadata references. Azure SQL DB cannot use those server-qualified references even for the current database. The same output path builds additional server-qualified metadata, upgrade, and INSERT statements, so all of them need a consistent Azure-local prefix. Explicit remote output should remain unsupported.

Validation used an isolated procedure schema; test objects were removed and all 21 existing dbo procedure definition hashes were unchanged. SQL Server 2022 and Windows SQL Server 2025 local output passed separately.

BrentOzar:
Azure SQL DB output fails because the local output path adds the server name to metadata lookups and INSERT targets. Omit that server prefix on Azure while retaining database qualification, including the command that upgrades older output tables. Reject an explicit @OutputServerName on Azure with an actionable message.

Validation on Azure SQL DB: Modes 0–4 each created and reused an output table; Mode 2 also restored a deliberately removed total_forwarded_fetch_count column and appended rows with the output database omitted. Explicit remote output raised the expected message. Temporary objects were removed and all 21 existing dbo procedure definition hashes were unchanged. SQL Server 2022 local create/append control and git diff --check passed.

Closes #4112.

### DIFF
--- a/.github/scripts/azure-index-output-regression.sql
+++ b/.github/scripts/azure-index-output-regression.sql
@@ -1,0 +1,58 @@
+/* Dedicated test schema belongs to this serialized Azure CI workflow. */
+SET NOCOUNT ON;
+IF SCHEMA_ID(N'FRKIndexSmoke') IS NULL
+    EXEC(N'CREATE SCHEMA FRKIndexSmoke AUTHORIZATION dbo');
+DROP TABLE IF EXISTS FRKIndexSmoke.Inventory;
+DROP TABLE IF EXISTS FRKIndexSmoke.SourceData;
+CREATE TABLE FRKIndexSmoke.SourceData (ID int NOT NULL PRIMARY KEY, Payload char(100));
+INSERT FRKIndexSmoke.SourceData VALUES (1, 'a'), (2, 'b');
+DECLARE @Database sysname = DB_NAME();
+BEGIN TRY
+    EXEC dbo.sp_BlitzIndex @Mode = 2, @DatabaseName = @Database,
+         @OutputDatabaseName = @Database, @OutputSchemaName = N'FRKIndexSmoke',
+         @OutputTableName = N'Inventory';
+    IF OBJECT_ID(N'FRKIndexSmoke.Inventory') IS NULL
+        THROW 51000, 'Azure index output table was not created.', 1;
+    IF NOT EXISTS (SELECT 1 FROM FRKIndexSmoke.Inventory)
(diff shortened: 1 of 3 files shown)


BrentOzar:
Added real Azure CI coverage: the runner now clears and installs this checkout's sp_BlitzIndex, then executes a dedicated-schema fixture that verifies nonempty local inventory creation, append, missing-column upgrade with append, and the explicit remote-output rejection. The runner requires EngineEdition 5. The fixture cleans its owned tables/schema on success and SQL errors, and clears leftovers from interrupted runs.

Create/append/upgrade controls passed on disposable SQL Server 2022, and shell syntax/git diff --check pass. The new Azure CI job will exercise the Azure-specific path and rejection. No generated installers were edited.

BrentOzar:
Expanded the regression to create and reuse output tables for modes 0, 1, 3, and 4, alongside the existing nonempty Mode 2 create/append checks. The legacy-column upgrade now omits @OutputDatabaseName on Azure and uses the qualified output-table name; boxed controls retain explicit database output because that default is Azure-specific. All boxed mode controls pass.

Integrated current dev and resolved the Azure runner overlap by preserving both the merged include/exclude regression and the index-output regression. Shell syntax and git diff --check pass. The new CI run will validate all Azure-specific paths before merge.

BrentOzar:
Scoped the Azure procedure cleanup verification to dbo, matching the procedures actually dropped. Same-named procedures in another schema no longer cause a false stale-install failure. Shell syntax and git diff --check pass; all-mode and omitted-database Azure coverage remain intact.

PR #4116, opened 2026-09-12
sp_BlitzBackups: count logs overlapping the full or differential endpoint

### LINKED ISSUE #4115: sp_BlitzBackups: restore estimates omit log backups overlapping the full or differential endpoint
Confirmed on unchanged dev 2e2ac07e using unmodified backup history from an actual full/differential/log restore test on SQL Server 2022.

The recovery point had:
- Full last_lsn: 42000000039200001
- Differential last_lsn: 42000000053600001
- Log first_lsn: 42000000036800001
- Log last_lsn / recovery endpoint: 42000000054400001
- Required log backup size: 147456 bytes

The log extends beyond the differential and is required to reach that recovery endpoint. However, LogTotals requires `bLog.first_lsn > COALESCE(rp.diff_last_lsn, rp.full_last_lsn)`, so it excludes the overlapping log. The full procedure's recovery point reports NULL log_backups even though one log backup is required. Log size and duration are consequently omitted from the restore estimate. This is independent of the time-unit correction in #4069/#4090.

Reproduction copied the actual backupset and backupmediafamily rows into an isolated database, used @MSDBName to read that copy, and added only a final SELECT of #RTORecoveryPoints for inspection. No LSNs, sizes, dates, or calculation logic were changed.

Log totals should include logs whose ending LSN extends beyond the restored full/differential endpoint and is at or before the target log endpoint. Please cover overlapping, exactly adjacent, and already-covered logs, with and without a differential.

BrentOzar:
RTO estimates can omit logs that overlap a full or differential backup’s ending LSN. Select those logs correctly, retain the actual endpoint backup ID, and remove redundant contained intervals with an ordered running maximum. Equal intervals prefer a known regular backup.

Include the preceding full when computing worst-case recovery points for the reporting window, preserve datetime precision at that anchor, and support differential endpoints without logs. Skip unusable optional backups. Required discard/damaged logs, missing required media metadata, incompatible known forks, or an unusable full base produce an explained NULL RTO. Media validation requires every expected stripe family. A later full does not conceal an earlier restore gap within the requested interval.

Centralized history remains readable when metadata is incomplete. Missing fork identifiers produce a warning; pushes now preserve fork, differential-base, copy-only, and media-family fields. CheckIds 15 and 16 are documented.

Validation: the native SQL Server fixture creates full, differential, regular, copy-only, later-full, and two-stripe backups. It asserts endpoint IDs, counts, sizes, durations, visible RTO, legacy/partial metadata, optional and required failures, differential-only recovery, preceding anchors, and actual history-push field mappings. The expanded fixture passes on SQL Server 2022. Reverting the timestamp comparison or mixed-fork deduplication makes their regressions fail. The combined kit passed all 53 native smoke steps plus uninstall; CI runs the final change on SQL Server 2017, SQL Server 2025, and Azure. git diff --check passes.

Closes #4115.

### DIFF
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -1039,6 +1039,28 @@ BEGIN TRY
     IF EXISTS(SELECT 1 FROM #FRKRecoveryProof) OR EXISTS(SELECT 1 FROM #FRKBackupProof WHERE RTOWorstCaseMinutes IS NOT NULL)
        OR NOT EXISTS(SELECT 1 FROM #FRKWarningProof WHERE Finding=N'RTO estimate unavailable')
         THROW 51000,'Damaged regular log received a restore estimate.',1;
+    /* A real striped full requires metadata for both media families. */
+    DECLARE @SecondFile nvarchar(512)=@Root+N'striped2.bak';
+    SET @File=@Root+N'striped1.bak';
+    BACKUP DATABASE FRKLogSource TO DISK=@File,DISK=@SecondFile WITH COPY_ONLY,INIT;
+    DECLARE @StripedFull int=(SELECT MAX(backup_set_id) FROM msdb.dbo.backupset WHERE database_name=N'FRKLogSource');
+    DROP TABLE FRKLogHistory.dbo.backupset;
+    SELECT * INTO FRKLogHistory.dbo.backupset FROM msdb.dbo.backupset WHERE backup_set_id=@StripedFull;
+    DROP TABLE FRKLogHistory.dbo.backupmediafamily;
+    SELECT * INTO FRKLogHistory.dbo.backupmediafamily FROM msdb.dbo.backupmediafamily
+      WHERE media_set_id IN(SELECT media_set_id FROM FRKLogHistory.dbo.backupset);
+    IF (SELECT COUNT(*) FROM FRKLogHistory.dbo.backupmediafamily)<>2
+        THROW 51000,'Striped fixture did not produce two families.',1;
+    DELETE #FRKRecoveryProof; DELETE #FRKBackupProof; DELETE #FRKWarningProof;
+    EXEC FRKLogHistory.dbo.sp_BlitzBackups @MSDBName=N'FRKLogHistory';
+    IF NOT EXISTS(SELECT 1 FROM #FRKBackupProof WHERE RTOWorstCaseMinutes IS NOT NULL)
+        THROW 51000,'Complete striped metadata did not permit an estimate.',1;
+    DELETE #FRKRecoveryProof; DELETE #FRKBackupProof; DELETE #FRKWarningProof;
+    DELETE FRKLogHistory.dbo.backupmediafamily WHERE family_sequence_number=2;
+    EXEC FRKLogHistory.dbo.sp_BlitzBackups @MSDBName=N'FRKLogHistory';
+    IF EXISTS(SELECT 1 FROM #FRKBackupProof WHERE RTOWorstCaseMinutes IS NOT NULL)
+       OR NOT EXISTS(SELECT 1 FROM #FRKWarningProof WHERE Finding=N'RTO estimate unavailable')
+        THROW 51000,'Missing stripe family received a restore estimate.',1;
     /* Exercise the unmodified push path with four-part local-server names.
        Precreate the destination so this tests insertion/mapping without remote DDL. */
     DROP TABLE FRKLogHistory.dbo.backupset;
@@ -1049,11 +1071,11 @@ BEGIN TRY
     IF NOT EXISTS(SELECT 1 FROM FRKLogHistory.dbo.backupset WHERE database_name=N'FRKLogSource')
         THROW 51000,'History push did not insert fixture backups.',1;
     IF EXISTS(
-      SELECT backup_set_uuid,first_recovery_fork_guid,last_recovery_fork_guid,fork_point_lsn,
+      SELECT backup_set_uuid,first_family_number,last_family_number,first_recovery_fork_guid,last_recovery_fork_guid,fork_point_lsn,
         differential_base_lsn,differential_base_guid,is_copy_only FROM msdb.dbo.backupset
       WHERE database_name=N'FRKLogSource' AND database_guid=(SELECT database_guid FROM sys.database_recovery_status WHERE database_id=DB_ID(N'FRKLogSource'))
       EXCEPT
-      SELECT backup_set_uuid,first_recovery_fork_guid,last_recovery_fork_guid,fork_point_lsn,
+      SELECT backup_set_uuid,first_family_number,last_family_number,first_recovery_fork_guid,last_recovery_fork_guid,fork_point_lsn,
         differential_base_lsn,differential_base_guid,is_copy_only FROM FRKLogHistory.dbo.backupset)
         THROW 51000,'History push lost or mis-mapped recovery metadata.',1;
     DROP DATABASE FRKLogHistory;
--- a/.github/scripts/smoke-test-matrix.sql
+++ b/.github/scripts/smoke-test-matrix.sql
@@ -577,16 +577,23 @@ PRINT 'Analysis, direct export, all, and all avg preserved the other session.';
 --#STEP: sp_BlitzBackups excludes redundant copy-only logs
 IF DB_ID(N'FRKLogSource') IS NOT NULL OR DB_ID(N'FRKLogHistory') IS NOT NULL
     THROW 51000, 'Backup overlap fixture already exists.', 1;
-EXEC(N'CREATE DATABASE FRKLogSource;');
-EXEC(N'CREATE DATABASE FRKLogHistory;');
+BEGIN TRY
+    EXEC(N'CREATE DATABASE FRKLogSource;');
+    EXEC(N'CREATE DATABASE FRKLogHistory;');
+END TRY
+BEGIN CATCH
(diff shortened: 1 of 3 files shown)


BrentOzar:
Fixed the overlapping alternatives in 3ee8ed33: an interval covered by another eligible log backup on the same recovery fork is excluded, with an ID tie-break for identical intervals. A copy-only endpoint is retained when no covering backup reaches that endpoint. Added a CI regression using actual full/diff/copy-only/regular/copy-only-endpoint backups and a private copy of their native LSNs and sizes. It runs the installed procedure's unchanged calculations with test-only result capture, checking counts, sizes, durations, and user-visible RTO for four histories (including full-only). Fixed SQL Server 2022 run passes; prior head fails the log totals assertion. Current dev is integrated.

BrentOzar:
Addressed the follow-up findings in ab363b5b. Older-full recovery points now match the same database, use ending-LSN overlap with the next-full upper bound, and retain the selected log ID. Future centralized-history pushes include fork, differential-base, and copy-only metadata; legacy NULL forks compare consistently and produce a metadata warning. Histories containing multiple known recovery forks or recent backups to NUL /dev/null now leave RTO NULL with an explicit explanation rather than inventing an unusable chain. This deliberately does not attempt a full cross-fork restore planner. Expanded native-history regression passes on SQL Server 2022: correct endpoint IDs, regular/copy-only totals, legacy NULL metadata, explained unavailable estimates for forks/discard devices, and the earlier full's endpoint after adding a later full. Cleanup uses explicit cutoff dates.

BrentOzar:
Addressed the remaining detailed review in 03179167. All RTO calculations now share indexed candidate history bounded by the reporting window and its preceding usable full backup. Discarded COPY_ONLY rows are omitted before endpoint/coverage selection, so they cannot suppress a valid regular chain. Native-history tests cover that case and an ancient fork outside a one-hour window; repeated runs pass. Added deterministic endpoint tie-breakers, prevented duplicate exclusion warnings, documented CheckIDs 15/16, and integrated current dev. Recent regular discard backups and multiple known candidate forks remain explicitly unavailable; missing legacy fork metadata is warned.

BrentOzar:
Fixed the next-full crossing case in e580b6c9. The older full now chooses the earliest log spanning the next full's boundary (usable with STOPAT), falling back to the latest preceding log when no crossing log exists. Extended the native test to take another log after the later full and assert that both recovery points retain their correct endpoint/counts. The complete expanded regression passes on SQL Server 2022.

BrentOzar:
faeefcd2 addresses the outstanding reviews. Media reads now tolerate centralized history without backupmediafamily and explicitly leave RTO unknown with Check 15. Unusable full/diff alternatives are omitted; regular discard logs are checked back to the selected full, including before the reporting window. Databases with an older full and current logs remain visible in the output. Endpoint selection is restricted to the reporting window; equal intervals prefer regular backups. A running MAX replaces the log range self-join. Listener push now includes differential_base_lsn. Fixture setup cleanup and Check 16 assertions are included.

The expanded real-backup regression passes all cases, including missing media, unusable preceding full, usable chains alongside discarded full/diff alternatives, pre-window discarded logs, equal intervals, and old forks. Testing also found a datetime/datetime2 promotion bug that intermittently excluded the anchor full at its own timestamp: the comparison now preserves datetime precision. A deterministic .007 timestamp passes with the fix and fails when only that comparison is reverted.

BrentOzar:
40f9dc56 fixes the mixed legacy/new fork-metadata finding. Histories with conflicting known forks are already excluded, so the remaining estimate deduplicates intervals per recovery point even when one backup's fork metadata is NULL. The explicit missing-metadata warning remains. The expanded real-backup regression fails with the previous partition and passes with this correction; it was run in separately named disposable databases so the ongoing full integration smoke pass remained isolated.

BrentOzar:
0d50c4fc handles NULL is_copy_only conservatively for discard logs and includes /dev/null in both Check 14 branches. Regression coverage verifies NULL copy-only discard history cannot receive RTO and Linux copy-only discard warnings are emitted.

Two latest findings are false positives, now demonstrated by explicit assertions. Case 2 deletes @Endpoint before case 3; a new guard verifies its absence before case 3 removes @Regular. Also, a differential is optional for restoration: when its base GUID is unknown, the code selects the full-plus-log path and counts logs from the full rather than from that differential. A new native-backup case clears the differential base GUID and verifies the complete full-plus-log estimate (10-second full plus 40-second regular log, displayed as 0.8 minutes). Inferring an unknown differential base would weaken restore-chain validation. All expanded cases pass.

BrentOzar:
94497680 validates media-row coverage for each candidate backup, not merely the existence of backupmediafamily. Missing candidate media rows now produce an explained NULL RTO; the partial-history regression passes. e8615f5 merges current dev after #4111 and #4113, preserving all 53 smoke steps.

The complete combined suite has now passed all 53 steps plus uninstall using native Linux sqlcmd inside the disposable SQL Server container, including the cache isolation step, real ring-buffer deadlock, upgraded history views, all quoted-output regressions, and expanded backups. The final per-candidate media-coverage regression also passes independently. Azure/SQL Server 2017/2025 CI is running on this exact integrated head.

BrentOzar:
933d1c0d uses explicit CASE ordering for known regular, known copy-only, then unknown at all three tie-breakers. The native equal-interval regression now includes NULL copy-only metadata and verifies the regular endpoint and duration win. The fixture also invokes the unmodified @PushBackupHistoryToListener path with a four-part local-server destination and asserts every added fork/base/copy-only field after insertion; this passes.

The media-row coverage finding was already fixed in 94497680/e8615f5, with an explicit partial-media regression. Regarding discards before a later in-window full: this reports the worst case over the requested interval, not just the latest restore point. A later full does not restore the earlier portion across that discard-log gap. Restricting the estimate to only the later healthy portion would silently change that meaning; the conservative NULL result is now explicitly documented.

BrentOzar:
b09db39c adds the preceding anchor full to recovery points even when a newer full exists, and filters unusable full/differential/copy-only alternatives while reporting damaged required logs as unavailable RTO. The native fixture verifies both the older anchor-to-next-full point and damaged full/diff versus damaged regular-log behavior. The complete expanded backup fixture passes, including the real push-path assertion.

The timestamp regression concerns comparison against the selected full's own .007 datetime value, not equality to @StartTime. Reverting only the previous COALESCE(datetime, datetime2) expression makes that fixture fail; the documented failing control is backup-precision-before-regression.log. Changing the reporting-window <= predicate is a different scenario. The other NULL-order/push-coverage comments were against e8615f5 and already fixed in 933d1c0d.

BrentOzar:
8e1b4c63 addresses the latest Codex findings. A valid differential can now be the endpoint when there is no log endpoint; the SIMPLE-style full/differential regression asserts its 20-second differential and 0.5-minute full-plus-diff RTO. Unverifiable optional full/diff/copy-only alternatives are skipped, while unknown required log media remains an explicit unavailable estimate. The healthy regular-chain test now deletes the optional copy-only media row and verifies its RTO remains available. Full-only history with no usable base receives an explanatory warning. The full expanded backup fixture passes.

BrentOzar:
bb232298 validates every expected media family using the first/last family metadata and distinct recorded family sequence numbers, rather than accepting any single non-NULL device row. A real two-stripe full backup is now in the automated fixture: complete metadata permits RTO, removing its second family produces an explained NULL. The history push now preserves first_family_number and last_family_number, and the actual push-path assertion compares those fields too. The entire expanded native backup regression passes.

BrentOzar:
The requirement for one complete mirror is a false positive. SQL Server permits replacing a family's media with the corresponding family from another mirror: https://learn.microsoft.com/en-us/sql/relational-databases/backup-restore/mirrored-backup-media-sets-sql-server . The fixture now creates a real two-family/two-mirror backup, retains complementary family rows from different mirrors, and executes RESTORE VERIFYONLY using those two files. SQL Server reports the backup set valid, and the RTO regression passes.

The other mirror finding is fixed: known media now requires a usable non-discard copy of each family, and discard classification no longer contaminates another usable set of copies. Tests cover the cross-mirror restore, a discarded mirror alongside a usable mirror, and a missing family across all mirrors. The entire expanded native backup regression passes.

BrentOzar:
Final bounded compatibility fix: 3f085e32 persists per-backup source media facts through the built-in history push, so pushed history can produce RTO without a backupmediafamily table. The expanded focused regression passed on the disposable SQL Server 2022 instance, including actual push/readback; git diff --check passed. The broader 53-step native smoke suite passed earlier, before this last compatibility change.

Remaining gaps are tracked in #4117 (continuous log coverage) and #4118 (remote push validation and metadata refresh). At the maintainer's request, the automatic review/fix loop is paused and no further review requests are being queued. This PR remains open; this comment does not claim current-head Copilot approval or remote-server validation.

BrentOzar:
Bounded follow-up in 7f0b17e0 addresses the three remaining findings, with no broader review requested:

- A pre-window full is retained only when there is an eligible recent differential/log endpoint; it cannot become a standalone out-of-window recovery point.
- Missing source media facts are backfilled for retained full backups even outside the push window.
- Existing listeners with RPC OUT disabled keep their distributed-DML push path. If media columns already exist, facts are updated without RPC. If absent, the push continues with an explicit message explaining the two nullable bit columns needed for centralized RTO metadata, rather than attempting remote DDL and failing.

Focused SQL Server 2022 backup regression PASS, including added unusable-endpoint and out-of-window-full backfill assertions. git diff --check PASS. Actual remote linked-server validation remains tracked in #4118; this comment does not claim that test was executed. #4117 and the remaining scope of #4118 are deferred. Automatic monitoring remains paused.

PR #4120, opened 2026-09-12
sp_BlitzBackups: refresh pushed media facts

### LINKED ISSUE #4118: sp_BlitzBackups: cover remote history-push media facts and refresh behavior
Follow-up to #4116, deferred at the maintainer's request to bound further work.

The push path now persists source-derived frk_media_is_usable and frk_media_has_discard facts by backup UUID, avoiding ambiguous instance-local media IDs in centralized history. The focused regression passes using the local server's four-part destination name, including reading pushed history without a backupmediafamily table.

Remaining coverage/behavior to define:

- Exercise an actual remote linked server, including adding the columns to an existing destination, permissions/RPC requirements, and the source-to-destination metadata update.
- Facts are backfilled only while either flag is NULL and only within @WriteBackupsLastHours. Rows outside that window remain unknown. Previously populated facts do not refresh if source metadata is corrected later. Define whether an explicit historical backfill/refresh option is needed, without turning source retention into a claim that physical backup files disappeared.

This records untested remote behavior and metadata lifecycle limitations; the local push regression does not establish remote deployment compatibility.

BrentOzar:
Recompute existing media facts in the push window and retained full backups, so stale values are corrected. Missing source rows remain unchanged; widen the existing lookback to refresh older logs.

Fixes #4118.

Validation: committed manual integration test passed across two disposable SQL Server 2022 instances: remote schema upgrade, populated-fact refresh outside the window, and RPC-disabled pushes with/without media columns. No new options or unrelated changes.

### DIFF
--- a/.github/scripts/backup-remote-push-regression.sql
+++ b/.github/scripts/backup-remote-push-regression.sql
@@ -11,7 +11,7 @@ IF EXISTS(SELECT 1 FROM FRKPushRemote.master.sys.databases WHERE name=N'FRKPushH
 CREATE DATABASE FRKPushSource;
 GO
 DECLARE @File nvarchar(512)=CONVERT(nvarchar(400),SERVERPROPERTY('InstanceDefaultDataPath'))+N'FRKPushSource.bak',
-        @Definition nvarchar(max),@UUID uniqueidentifier;
+        @Definition nvarchar(max),@UUID uniqueidentifier,@RemoteCreated bit=0;
 BEGIN TRY
     BACKUP DATABASE FRKPushSource TO DISK=@File WITH INIT;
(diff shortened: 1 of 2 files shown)


BrentOzar:
Addressed the two in-scope findings in c19ef13f:

- Compare recomputed media facts and all copied metadata using NULL-safe EXCEPT; update only changed rows. The remote regression now audits affected rows and confirms an unchanged repeat push writes zero rows.
- Track remote fixture creation before cleanup. Injecting a failure before backup/remote creation preserves the original error and leaves neither fixture database behind.

The suppressed RPC-disabled finding is not applicable: the cited table/index EXEC AT batches are inside IF @@ROWCOUNT = 0 for the missing backupset table. Existing-table pushes skip them. Actual two-server SQL Server 2022 regression passed again with RPC OUT disabled, both with and without media columns. diff check passed. No unrelated changes or expanded review scope.

